PR #4, opened 2020-10-23
Implement type constructors and basic interactions

ltacker:
Implement types constructor and basic interactions.
Tests on types


fadeev:
@ltacker So, right now a user is effectively an sdk.AccAddress. Shall we add a username as well? Later on a “address” will have to be an array to keep all the addresses a user has.

ltacker:
> @ltacker So, right now a user is effectively an sdk.AccAddress. Shall we add a username as well? Later on a “address” will have to be an array to keep all the addresses a user has.

The `User` structure holds data that can be converted into a `AddressableUser` an interface that implements `Address() sdk.AccAddress`. So that for future implementations of User we will always have an address to verify messages signatures

The current structure used for users for now is `AccountAddressUser` which contains only a sdk.AccAddress. I agree that we could add a required username as well so that the UI of the chat and SN could look better. I just wonder if it's not subject to security risks: A coordinator agree for a specific user to add it in the genesis then another address with the same username send the genTx, the coordinator add this genTx without verifying the address.

ltacker:
If we agree that in the future a user will have an array of address, we can change directly the implementation of `AddressableUser` with the interface `Addresses() []sdk.AccAddress`

fadeev:
User names will not replace addresses. Any important action will still require you to confirm the AccAddress. I think we can consider username as just a helpful label for a set of addresses.

x/chat/types/messages.go
```diff
@@ -0,0 +1,233 @@
...
+	bz := ModuleCdc.MustMarshalJSON(&msg)
+	return sdk.MustSortJSON(bz)
+}
+
+// ValidateBasic implements the sdk.Msg interface.
+func (msg MsgVotePoll) ValidateBasic() error {
+	// The user must possess the sign address
+	chatUser, err := msg.Author.DecodeChatUser()
+	if err != nil {
+		panic(err)
```
ilgooz, line 218:
```suggestion
		return err
```

x/chat/types/messages.go
```diff
@@ -0,0 +1,233 @@
...
+	bz := ModuleCdc.MustMarshalJSON(&msg)
+	return sdk.MustSortJSON(bz)
+}
+
+// ValidateBasic implements the sdk.Msg interface.
+func (msg MsgCreateChannel) ValidateBasic() error {
+	// The user must possess the sign address
+	chatUser, err := msg.Creator.DecodeChatUser()
+	if err != nil {
+		panic(err)
```
ilgooz, line 76: return err

x/chat/types/messages.go
```diff
@@ -0,0 +1,233 @@
...
+	bz := ModuleCdc.MustMarshalJSON(&msg)
+	return sdk.MustSortJSON(bz)
+}
+
+// ValidateBasic implements the sdk.Msg interface.
+func (msg MsgSendMessage) ValidateBasic() error {
+	// The user must possess the sign address
+	chatUser, err := msg.Author.DecodeChatUser()
+	if err != nil {
+		panic(err)
```
ilgooz, line 149: return err

x/chat/types/channel.go
```diff
@@ -0,0 +1,52 @@
...
+
+// NewChannel creates a new channel
+func NewChannel(
+	id int32,
+	creator User,
+	name string,
+	subject string,
+	metadata *types.Any,
+) (Channel, error) {
+	channel := new(Channel)
```
ilgooz, line 21: to not dereference on returns, maybe we can initialize these like this

```suggestion
	channel := Channel{}
```

x/chat/types/messages_test.go
```diff
@@ -0,0 +1,172 @@
...
+	msg, _ = types.NewMsgVotePoll(
+		"0xaaa",
+		user,
+		0,
+		nil,
+		otherAddress,
+	)
+	err = msg.ValidateBasic()
+	if err == nil {
+		t.Errorf("NewMsgVotePoll with invalid address should create an invalid MsgVotePoll")
```
ilgooz, line 170: not mandatory but we can also use _testify/require_ for assertions since it is also used at _starport_.

x/chat/types/poll.go
```diff
@@ -0,0 +1,100 @@
...
+	vote, ok := poll.Votes[chatUser.Identifier()]
+	if !ok {
+		return nil, errors.New("No vote found")
+	}
+
+	return vote, nil
+}
+
+// AppendVote appends a vote into the poll
+func (poll *Poll) AppendVote(vote *Vote) error {
```
ilgooz, line 52: Some uses _*Poll_ and some _Poll_ while declaring the methods, maybe we can stick with one way for all.

x/chat/types/messages.go
```diff
@@ -0,0 +1,233 @@
...
+// Chat message types
+const (
+	TypeMsgCreateChannel = "create_channel"
+	TypeMsgSendMessage   = "send_message"
+	TypeMsgVotePoll      = "vote_poll"
+)
+
+// Verify interface at compile time
+var (
+	_ sdk.Msg = &MsgCreateChannel{}
```
ilgooz, line 19: not mandatory but these checks can be moved to tests

ilgooz (approved):
Great work! :confetti_ball:

fadeev (approved):
Replacing `author` with `creator` as discussed, otherwise 👍

x/chat/types/channel.go
ltacker: Thanks for the comment, I used `new` because the function was returning pointer at first but it actually complicates the code for nothing.
I'm not sure you can have dereference on return in Go though

x/chat/types/poll.go
ltacker: `*Poll` allows explicitly telling that the method is mutable for the object.

We could have `func (poll Poll) AppendVote(vote *Vote) (Poll, error)` instead but I think it's a bit less efficient

x/chat/types/messages.go
ltacker: I followed the Cosmos SDK implementation, I think the advantage is the the IDE can tell you the unimplemented method of your object in real time

PR #13, opened 2020-10-27
Initialize CI environment

ltacker:
- Create Github actions for lint, integration tests and unit tests.
- Add an integration test


integration/env_test.go
```diff
@@ -0,0 +1,183 @@
+package integration_test
```
ilgooz, line 1: What do you think about actually making this utility public in Starport repo so we can use it here by importing? -Tho this can be any time, I think for now it is OK to duplicate the code here.

ilgooz (approved):
:beer:

fadeev (approved):
Magic ✨

integration/env_test.go
ltacker: That's a good idea. Let's see in the future if we will need this again :)

PR #17, opened 2020-10-28
Create a simple Identity module and connect it to chat

ltacker:
- The `chat` module is now dependent of an `identity` module that exposes the following interface:
```
type IdentityKeeper interface {
	GetIdentifier(ctx sdk.Context, address sdk.AccAddress) (string, error)
}
```
This function returns an `identifier String` that uniquely identifies a user and should never change over time. This identifier is stored as channel/message/vote creator and ensures a user doesn't vote twice.

- Create a simple `identity` module:
This first `identity` module simply maps addresses with a username.
The identifier is the address itself.

The module exposes these command to the user:
```
q get-username [identifier] - returns the username associated with a user identifier
q get-username-from-address [address] - returns the username associated with a user address
tx set-username [username] - set the username of the user that signs the message
```

`get-username` and `get-username-from-address` have the same behavior for now but we must separate the two commands because, in the future, `spn` could switch to a more developed `identity` module where addresses can change for the same identifier.


ltacker:
More generally,  the interface of any `identity` module could be as follow:
```
GetIdentifier(ctx sdk.Context, address sdk.AccAddress) (string, error) // Get the user identifier from its address
GetAddresses(ctx sdk.Context, identifier string) ([]sdk.AccAddress, error) // Get all the addresses associated with a user identifier
```

x/identity/client/rest/rest.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/tendermint/spn/x/identity/types"
+)
+
+const (
+	MethodGet  = "GET"
+	MethodPost = "POST"
+)
+
+var (
+	UsernameRoute            = fmt.Sprintf("custom/%s/%s", types.QuerierRoute, types.QueryUsername)
```
fadeev, line 22: @ltacker Is there a reason to use `custom`? Just curious.

fadeev (approved):
Looks like a solid start for an identity module 🙂

Here is some vintage light reading on the topic:

https://github.com/tendermint/identity 📇

ilgooz (approved):
Nice! :chains:

x/identity/client/rest/rest.go
ltacker: I actually needed to remove them for the REST route but you must put this prefix for the querier route when it's a custom module.

PR #18, opened 2020-10-30
feat(chat): create channel transaction and describe channel query

ltacker:
- Add the transaction to create a channel
- Add the query to describe a channel
- Use `[]byte` for payload data instead of `proto.Any` because of a bug
- Add `identityExist()` method to the identity module so that if other module uses `chat`, the identity provided is checked in the `chat` keeper functions


ilgooz (approved):
:beers: Cheers.

x/chat/client/cli/tx.go
```diff
@@ -20,7 +21,43 @@ func GetTxCmd() *cobra.Command {
...
-	// this line is used by starport scaffolding # 1
+	cmd.AddCommand(CmdCreateChannel())
+
+	return cmd
+}
+
+// CmdCreateChannel returns the transaction command to create a new channel
+func CmdCreateChannel() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create-channel [name] [subject]",
```
fadeev, line 32: Not sure the `subject` should be required.

x/chat/client/cli/query.go
```diff
@@ -24,8 +25,48 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
...
+	)
 
-	return cmd 
+	return cmd
 }
 
+// CmdDescribeChannel returns the command to describe a channel
+func CmdDescribeChannel() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe-channel [channel-id]",
```
fadeev, line 38: "Describe" sounds like action with side-effects. Let's stick to CRUD-like verbs: show, create, update, etc.

x/chat/client/rest/rest_tx.go
```diff
@@ -0,0 +1,55 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/rest"
+	"github.com/tendermint/spn/x/chat/types"
+)
+
+type createChannelRequest struct {
+	BaseReq rest.BaseReq `json:"base_req"`
+	Creator string       `json:"creator"`
+	Name    string       `json:"name"`
+	Subject string       `json:"subject"`
```
fadeev, line 17: I propose using "Title" and "Description" instead.

fadeev (approved):
🔥

Issue #23, opened 2020-11-02
Chat: add pagination options to queries


lubtd:
Currently the queries `list-messages` and `search-messages` send all the messages requested in the response body.
We should be able to provide options to have paginated responses

PR #28, opened 2020-11-04
feat(chat): query to list channels

ltacker:
Query to list all channels
This query is also concerned by: #23


x/chat/types/keys.go
```diff
@@ -22,7 +22,7 @@ const (
 	ChannelCountKey = "channel_count"
 
 	// ChannelKey is the key to store the channels
-	ChannelKey = "channel"
+	ChannelKey = "channel_store"
```
fadeev, line 25: Is there a reason, why we can't use `channel-` as a prefix?

fadeev (approved):
🚀

x/chat/types/keys.go
ltacker: That would work, you just need to have a name that is not a prefix of another key name

ilgooz (approved):
:wine_glass:

PR #30, opened 2020-11-08
feat(genesis): keeper

ltacker:
Keeper for chains and proposals


ilgooz (approved):
utACK :confetti_ball:

fadeev (approved):
Awesome! 👏

Are `proposalID` globally unique or unique to a particular chain (like GH PRs)? And also, will `k.GetApprovedProposals` return approved proposals only for a particular chain or all approved proposals?

PR #31, opened 2020-11-10
feat(genesis): append transaction and queries for chains

ltacker:
Append the transaction:
```
spnd tx genesis create-chain [chain-id] [source-URL] [source-hash] [genesis-file]
```
Create and validate the new chain to launch. `genesis-file` is the path to a file `genesis.json`

Append the queries:
```
spnd q genesis list-chains // List all the chains id (with optional pagination)
spnd q genesis show-chain [chain-id] // Show info of a chain
```
`show-chain` has the optional flag `--save-genesis` where you provide a file path to save the genesis json of the chain into

---

The format of the genesis inside the chain is still `[]byte` but we use `github.com/tendermint/tendermint/libs/json.Marshal` to encode the genesis file


ilgooz (approved):
:+1: :confetti_ball:

fadeev (approved):

PR #40, opened 2020-11-15
feat(genesis): approve and reject proposals

ltacker:
- Implement `MsgApprove` and `MsgReject` transaction handlers
- Add a store in the keeper to register current accounts and validators for a genesis
- Implement `ApprovedProposals` and `RejectedProposals` queries


proto/genesis/v1beta/querier.proto
```diff
@@ -39,6 +41,22 @@ message QueryPendingProposalsResponse {
     repeated Proposal proposals = 2;
 }
 
+message QueryApprovedProposalsRequest {
```
ilgooz, line 44: What do you think about combining all query proposals into one request with a `type=proposal-type` filter in the request to make this API simpler to use?

`proposal-type` can be one of  _pending_, _approved_, _rejected_.

ilgooz (approved):
Brilliant! :t-rex:

fadeev (approved):
Great work, Lucas! 🤩 We're getting so close to a working PoC!

proto/genesis/v1beta/querier.proto
ltacker: I did this because I made the assumption that you rarely want to query all the proposals whatever the state is. It's easier to implement also since we group proposals in the store by their state.
If you think querying all proposals for a chain should be a feature, don't hesitate to raise an issue 👍

PR #41, opened 2020-11-17
feat(genesis): implement current-genesis query

ltacker:
Implement the query to get the current genesis from the initial genesis provided in the chain and the approved genesis.
Refactor the management of the genesis data.

Implement a test for `CurrentGenesis` with 20 accepted accounts and 10 accepted validators


fadeev (approved):
Looks good to me! 🦾

ilgooz (approved):
:fire: :fire_engine:

PR #44, opened 2020-11-17
fix(genesis): encoding error when broadcasting messages with CLI

ltacker:
Encoding errors prevented us using CLI to broadcast messages.
Fix:
- Use the protobuf files from Cosmos SDK to represent Coin and GenTx instead of the Any type
- `unpack` the message contained in the gentx before performing the message validation in `messages.go`


ltacker:
Now `spnd tx genesis proposal-add-account` and `spnd tx genesis proposal-add-validator` work!

ilgooz (approved):
:+1:

fadeev (approved):
🚀

PR #50, opened 2020-11-19
fix(genesis): MsgProposalAddAccount verify address

ltacker:
Verify that the address provided to `MsgProposalAddAccount` contains a valid Bech32 format


ilgooz (approved):
:ok_hand:

fadeev (approved):
👍

PR #55, opened 2020-11-19
feat(genesis): verify account funds for validator self-delegation

ltacker:
For approving a validator: verify that the associated account has enough funds for the self-delegation. If not, deny the approval for the add validator proposal


fadeev (approved):
👍

PR #56, opened 2020-11-19
feat(genesis): prevent creating a chain with a genesis containing existing accounts or gentxs

ltacker:
Ensure that when a chain is created, the genesis is empty


fadeev (approved):
👍

PR #65, opened 2020-11-23
ci: add action to auto deploy using ansible

helder-moreira:


fadeev (approved):
Looks great! Thanks again @helder-moreira 👏

.ansible/roles/cosmos-faucet/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+faucet_version: master
+log_level: info
+key_name: faucet
+cli_name: gaiacli
```
ltacker, line 4: What gaia is used for here?

ltacker (approved):
💪

.ansible/roles/cosmos-faucet/defaults/main.yml
helder-moreira: great catch! thank you! It should be `spnd`

helder-moreira: oh, actually those are the default values, as that role was initially designed for `gaia`. The value is overwritten [here](https://github.com/tendermint/spn/blob/helder/add-deploy-action/.ansible/playbook.yml#L58)

ltacker: Ok great 👍

PR #66, opened 2020-11-24
refactor: use new Cosmos SDK version

ltacker:
Update the code to support Cosmos-SDK v0.40.0-rc3 and Tendermint v0.34.0-rc6


fadeev (approved):
Great! 👍 I think we can merge this after CI to check that everything gets deployed properly.

ilgooz (approved):
:fire_engine:

PR #67, opened 2020-11-24
docs(genesis): first specification

ltacker:
Create the first specification for the `genesis` module based on the implementation of the POC


fadeev (approved):
That's very cool, thanks Lucas! You even added Vuepress tags to the frontmatter 😌

PR #79, opened 2020-11-30
refactor(genesis): remove genesis data from store

ltacker:
Genesis can be deterministically generated from the source URL + hash.
Since we don't perform stateful, on-chain verification, there is no need to keep this value on-chain anymore.


ilgooz (approved):

fadeev (approved):
👍

PR #86, opened 2020-12-03
refactor(query): ListChains sends all the whole chain information

ltacker:
`ListChains()` sends the whole information of the chains instead of just the chain ID


ilgooz (approved):
Small step for man giant leap for mankind. :man_astronaut:

PR #87, opened 2020-12-03
feat(genesis): ProposalCount query

ltacker:
Query to get the count of proposals of a chain


ilgooz (commented):
Do you think we can optionally filter by proposal `Type` & `Status` in the request or group by proposal `Type` & `Status` in the response in order to be able to implement Validators/Proposals sections in [chain list](https://github.com/tendermint/starport/issues/479#issuecomment-737815586)?

If this doesn't make sense to implement on server-side, I think we can pull back `ProposalCount()` and calculate it with new `Proposals()` function in the client-side.

ltacker:
We can implement it, the implementation will be a bit different since we need to go through the proposals (or store directly the count for all proposalType*Status but I think it would become lot of on-chain data)

ilgooz:
Yep, I agree on not to keep it on-chain.

Still, I started think about that backend api for count will become complicated and will complicate the client side usage as well. For ex, client needs to call this api individually get the count for total proposals, approved proposals, total validator proposals and approved validator proposals. Instead, client can just call `Proposals()` to get all proposals for the chain at once and do the calculation locally.

How do you feel about this?

ilgooz:
Do you think is there any other approach can be done in server-side to make this not complicated for server and client-side usage?

ltacker:
We need overall proposal count on-chain for the proposal ID so this query is not expensive to keep.

For specific status and type we could use one call to `ListProposals()` but with #479 it would be pretty expensive to call it for every chain (though I think we should implement #479 directly with filtering capability). If most chain that we list, we're not interested in proposal details, it will be a lot of useless operations.

In fact, it would not be that expensive to keep Type*Status counts on-chain, we will always have 3 status and I don't think we will have more than 5 proposal types. So it would be only 15 integers to keep on-chain (nothing compared to a gentx) but it's not a very clean design for a blockchain. You normally only keep on-chain what is minimally necessary and useful for on-chain logic. For data that can be interpreted from the current on-chain data and that are not useful for on-chain logic (like the number of approved add-validator proposals) you normally design a caching capability on the server.

I think we could now implement #479 with `ListProposals` query, and see how we could implement caching on `spn` for queries (that could be used to get the proposals data as well)

ilgooz:
Great! Let's experiment with `ListProposals` first. Do we hold this PR for now?

ltacker:
If getting the overall counts of proposals is something useful we can merge it. The implementation of the other cases would need a different implementation so we would create another PR

ilgooz (approved):

fadeev (approved):
utACK

PR #91, opened 2020-12-08
fix(genesis): sort pending proposals from list-proposals query

ltacker:
Sort the pending proposals so that the user can them in the order they have been created.

Improve the tests of proposals a bit


ilgooz (approved):

fadeev (approved):
utACK 👍

PR #96, opened 2020-12-15
feat: option in ListChains query to search list chains with a specific prefix

ltacker:
The `prefix ` of `QueryListChainsRequest ` can be used to only list chains with a specific prefix in the chain ID


fadeev (approved):
utACK 👍

ilgooz (approved):

PR #97, opened 2020-12-17
fix: update home directory

ltacker:
Update home directory to adapt to the new standard of Starport


fadeev (approved):

fadeev:
We would have to change this in the Starport repo as well, right? It's using `~/spnd`.

ilgooz:
@fadeev I guess no change needed in Starport's side because we use it for accounts only by managing keyring programatically.

PR #100, opened 2021-01-19
Chain Name Service Draft

fadeev:
Ideas on a Chain Name Registry module. [Rendered doc](https://github.com/tendermint/spn/blob/cns/docs/cns.md).

### DIFF
--- a/docs/cns.md
+++ b/docs/cns.md
@@ -6,7 +6,7 @@ The goal of this module is to be a simple registry with the ability to transfer 
 
 ## Chain name
 
-Chain name contains information about the chain, most importantly, chain ID used to uniquly identify the chain and a list of peer nodes.
+Chain name contains information about the chain, most importantly, chain ID used to uniquely identify the chain and a list of peer nodes.
 
 Key: `chain-[chainID] -> Chain Name` 
 
--- a/docs/cns.md
+++ b/docs/cns.md
@@ -4,7 +4,7 @@ The goal of this module is to be a simple registry with the ability to transfer 
 
 # State
 
-## Chain name
+## Chain Info
 
 Chain name contains information about the chain, most importantly, chain ID used to uniquely identify the chain and a list of peer nodes.
 
--- a/docs/cns.md
+++ b/docs/cns.md
@@ -1,110 +1,113 @@
 # Chain Name Service Draft
 
-The goal of this module is to be a simple registry with the ability to transfer ownership. Auctions and other mechanisms are out of scope and can be implemented alongside or on top of this system (perhaps, in a custodial way).
+The goal of this module is to be a simple registry of Cosmos Blockchains with the ability to transfer ownership.
+
+The development of CNS is planned to be divided into 3 separate phases in order to ensure there's enough time for existing chains and prevent front running
+
+**Phase 0** - Here, we expect owners of existing chains to claim their chain names which are approved by a previously agreed upon multi-sig. Estimated: 6 months.
+
+Ideas:
+
+1) Have a fixed price for reserved names, which would be transfered to community pool.
+
+**Phase 1** - After an agreed-upon time is over, we would open CNS to general users to register chain names of their choice.
+
+From this phase, the name will have an expiry time.
+
+**Phase 2** - Finally, we add a way to transfer ownership of chain names. It's still to be determined if auction should be implemented as a module, or make CNS NFT compatible and let NFT handle the transfer or a CosmWasm smart contract which handles the auction utilizing the transfer ownership feature of CNS.
 
 # State
 
-## Chain Info
+Goal 1 (end-users): let users determine whether the tokens they got through IBC came from a chain they claim to come from. This will be possible through the mapping of chain names to IBC clients.
 
-Chain name contains information about the chain, most importantly, chain ID used to uniquely identify the chain and a list of peer nodes.
-
-Key: `chain-[chainID] -> Chain Name` 
+Goal 2 (validators): let nodes (validators, full-nodes, etc.) join networks.
 
 ```go
-type ChainName struct {
-	ChainID string
-	Owner sdk.AccAddress
-	Peers []peer
-	SourceURL string
-	SourceCommitHash string
+type ChainInfo struct {
+	ChainName     string
+	Owner         OwnerProps
+	Expiration    int64
+	Metadata      [][2]string
+
+	// Goal 1
+	CanonicalIBCClientId string
+
+	// Goal 2
+	Seed          []seed
+	SourceCodeURL string //how likely is the possibility of code url changing
+	Version       VersionInfo
 }
 ```
 
-## Offer
-
-Offers are used to transfer ownership of a chain name.
+```go
+type VersionInfo struct {
+	 Version          int64
+	 SourceCommitHash string
+	 GenesisHash      string
+}
+```
 
 ```go
-type Offer struct {
-	ID int32
-	From sdk.AccAddress
-	Amount sdk.Coins
-	Status status // pending | rejected | approved
+type OwnerProps interface {
+	String()  string
+}
+```
+
+```go
+type Claim struct {
+  ID        int64
+  ChainName string
+  Owner     OwnerProps
+  Proof     string
 }
 ```
 
 # Messages
 
-## MsgEditChainName
+### Phase - 0
 
-Chain owner can make edits to the chain name.
+Ideally, we would want the chains to approve via a gov proposal the owner (likely address) of the chain name. This makes it easy for approvers and also prevent collision for the same namespace.
+
+### MsgClaimChainName
+
+Claim is about mapping chain name to owner. Name, owner, IBC client, meta.
 
 ```go
-type MsgEditChainName struct {
-	
+type MsgClaimChainName struct {
+	ChainName     string
+	CanonicalIBCClientId string
+	Proof         string
+	Owner         OwnerProps
 }
 ```
 
-## MsgTransferChainName
-
-To transfer a name from one owner from another without creating an offer with a zero amount.
+### MsgApproveClaim
 
 ```go
-type MsgTransferChainName struct {
-	From string
-	To string
-	ChainID string
+type MsgApproveClaim struct {
+	ClaimID  int32
+	Approver sdk.AccAddress
 }
 ```
 
-## MsgCreateOffer
+### MsgRejectClaim
 
 ```go
-type MsgCreateOffer struct {
-	ChainID string
-	Amount sdk.Coins
+type MsgRejectClaim struct{
+	ClaimID int32
+	Approver sdk.AccAddress
 }
 ```
 
-## MsgRejectOffer
+### MsgUpdateInfo
 
 ```go
-type MsgRejectOffer struct {
-	ChainID string
+type MsgUpdateInfo struct{
+	ChainName     string
+	Owner         OwnerProps
+	Seed          []seed
+	CanonicalIBCClientId string
+	Version       VersionInfo
+	Metadata      [][2]string
 }
 ```
-
-## MsgApproveOffer
-
-```go
-type MsgApproveOffer struct {
-	ChainID string
-}
-```
-
-## MsgRemoveOffer
-
-```go
-type MsgRejectOffer struct {
-	ChainID string
-}
-```
-
-## MsgRegisterName
-
-To accuire a name that doesn't have an owner. By default each name has a price. When registering a name, amount equivalent to the price goes to the community pool.
-
-```go
-type MsgRejectOffer struct {
-	ChainID string
-}
-```
-
-# Ideas
-
-- Reserve names for existing chains
-- Names with a chain IDs that have `tesnet-` prefix are registered for free
-- Off-chain tool to check that nodes in peer list are part of a network with a given chain ID. Can loop through all the chain names and create a table.
-- If you're launching a chain and don't want a single entity to be responsible for managing a chain name, use a multisig
-- This should probably be an IBC module
-- Open question: how does relate to IBC, relayers. How can this module be used with IBC.
--- a/docs/cns.md
+++ b/docs/cns.md
@@ -1,0 +1,110 @@
+# Chain Name Service Draft
+
+The goal of this module is to be a simple registry with the ability to transfer ownership. Auctions and other mechanisms are out of scope and can be implemented alongside or on top of this system (perhaps, in a custodial way).
+
+# State
+
+## Chain name
+
+Chain name contains information about the chain, most importantly, chain ID used to uniquly identify the chain and a list of peer nodes.
+
+Key: `chain-[chainID] -> Chain Name` 
+
+```go
+type ChainName struct {
+	ChainID string
+	Owner sdk.AccAddress
+	Peers []peer
+	SourceURL string
+	SourceCommitHash string
+}
+```
+
+## Offer
+
+Offers are used to transfer ownership of a chain name.
+
+```go
+type Offer struct {
+	ID int32
+	From sdk.AccAddress
+	Amount sdk.Coins
+	Status status // pending | rejected | approved
+}
+```
+
+# Messages
+
+## MsgEditChainName
+
+Chain owner can make edits to the chain name.
+
+```go
+type MsgEditChainName struct {
+	
+}
+```
+
+## MsgTransferChainName
+
+To transfer a name from one owner from another without creating an offer with a zero amount.
+
+```go
+type MsgTransferChainName struct {
+	From string
+	To string
+	ChainID string
+}
+```
+
+## MsgCreateOffer
+
+```go
+type MsgCreateOffer struct {
+	ChainID string
+	Amount sdk.Coins
+}
+```
+
+## MsgRejectOffer
+
+```go
+type MsgRejectOffer struct {
+	ChainID string
+}
+```
+
+## MsgApproveOffer
+
+```go
+type MsgApproveOffer struct {
+	ChainID string
+}
+```
+
+## MsgRemoveOffer
+
+```go
+type MsgRejectOffer struct {
+	ChainID string
+}
+```
+
+## MsgRegisterName
+
+To accuire a name that doesn't have an owner. By default each name has a price. When registering a name, amount equivalent to the price goes to the community pool.
+
+```go
+type MsgRejectOffer struct {
+	ChainID string
+}
+```
+
+# Ideas
+
+- Reserve names for existing chains
+- Names with a chain IDs that have `tesnet-` prefix are registered for free
+- Off-chain tool to check that nodes in peer list are part of a network with a given chain ID. Can loop through all the chain names and create a table.
+- If you're launching a chain and don't want a single entity to be responsible for managing a chain name, use a multisig
+- This should probably be an IBC module
+- Open question: how does relate to IBC, relayers. How can this module be used with IBC.


docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+
+Chain name contains information about the chain, most importantly, chain ID used to uniquly identify the chain and a list of peer nodes.
+
+Key: `chain-[chainID] -> Chain Name` 
+
+```go
+type ChainName struct {
+	ChainID string
+	Owner sdk.AccAddress
+	Peers []peer
```
okwme, line 17: are these all peers or just seed nodes?

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+	SourceURL string
```
okwme, line 18: what is this for?

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+	SourceCommitHash string
```
okwme, line 19: what is this?

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+## Offer
+
+Offers are used to transfer ownership of a chain name.
+
+```go
+type Offer struct {
+	ID int32
+	From sdk.AccAddress
+	Amount sdk.Coins
+	Status status // pending | rejected | approved
```
okwme, line 32: offer doesn't point to any specific chain-id. There doesn't seem to be any sort of join data structure which would link them either.

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+
+```go
+type MsgEditChainName struct {
+	
+}
+```
+
+## MsgTransferChainName
+
+To transfer a name from one owner from another without creating an offer with a zero amount.
```
okwme, line 50:
```suggestion
To transfer a name from one owner to another without creating an offer with a zero amount.
```

okwme: why would this require not making an offer with a zero amount? does it require making an offer with non-zero amount or not making an offer at all?

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+
+```go
+type MsgTransferChainName struct {
+	From string
```
okwme, line 54:
```suggestion
	From sdk.AccAddress
```
right?

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+	To string
```
okwme, line 55:
```suggestion
	To sdk.AccAddress
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+	ChainID string
+	Amount sdk.Coins
+}
+```
+
+## MsgRejectOffer
+
+```go
+type MsgRejectOffer struct {
+	ChainID string
```
okwme, line 73:
```suggestion
	ID int32
	ChainID string
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+## MsgApproveOffer
+
+```go
+type MsgApproveOffer struct {
+	ChainID string
```
okwme, line 81:
```suggestion
	ID int32
	ChainID string
```
same as last one, actually chainID becomes unnecessary but probably good to ensure there's not incorrect target of offer ID

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+## MsgRemoveOffer
+
+```go
+type MsgRejectOffer struct {
+	ChainID string
```
okwme, line 89:
```suggestion
	ID int32
	ChainID string
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+## MsgRegisterName
+
+To accuire a name that doesn't have an owner. By default each name has a price. When registering a name, amount equivalent to the price goes to the community pool.
```
okwme, line 95: ENS uses an oracle now to ensure price stability. Until that feature was available they just locked ETH and gave it back once the name was revoked. Maybe that's a way that would be simple for this to begin? It would be unfortunate if early users were punished by paying in ATOM instead of the fiat equivalent at the time. On the other hand going to the community pool is pretty easy to justify.

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+
+```go
+type MsgRejectOffer struct {
+	ChainID string
+}
+```
```
okwme, line 101:
```suggestion
```
redundant

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+
+# Ideas
+
+- Reserve names for existing chains
+- Names with a chain IDs that have `tesnet-` prefix are registered for free
```
okwme, line 106:
```suggestion
- Names with chain IDs that have `testnet-` prefix are registered for free
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+- Off-chain tool to check that nodes in peer list are part of a network with a given chain ID. Can loop through all the chain names and create a table.
```
okwme, line 107: This is an oracle's job for constant updates. Maybe it can be combined with the Gravity Bridge (peggy) since the validator set will essentially be a functional oracle by operating the bridge.

As a short term solution I'd suggest maintaining a list of seed nodes would be sufficient since they can be used to expose all other nodes. Otherwise there's a potentially huge cost in keeping constantly updated a list of active nodes beyond the seed list.

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+- If you're launching a chain and don't want a single entity to be responsible for managing a chain name, use a multisig
+- This should probably be an IBC module
```
okwme, line 109: the IBC keeper could be imported so that state could be read. Maybe eventually an endblocker to check if any IBC connection got updated with that chain id?

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+	ChainID string
+}
+```
+
+# Ideas
+
+- Reserve names for existing chains
+- Names with a chain IDs that have `tesnet-` prefix are registered for free
+- Off-chain tool to check that nodes in peer list are part of a network with a given chain ID. Can loop through all the chain names and create a table.
+- If you're launching a chain and don't want a single entity to be responsible for managing a chain name, use a multisig
```
okwme, line 108: could be good enough. Security is really only part of the consideration. IBC is providing the app hash of all potential chain-ids already. I think the goal should be to leverage the actual connections as the source of updated information

docs/cns.md
jackzampolin: We also have a generalized oracle module I'm working on...

docs/cns.md
jackzampolin: Ideally this is an `[]net.URL` of open RPC endpoints.

docs/cns.md
jackzampolin: Add `[]seed`

docs/cns.md
jackzampolin: I think hes saying genesis file location?

docs/cns.md
jackzampolin: What data does the oracle bring in? Also I'm in favor of community pool.

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+- This should probably be an IBC module
+- Open question: how does relate to IBC, relayers. How can this module be used with IBC.
```
okwme, line 110: ah, began addressing above. Will copy past previous comment here:
> IBC is providing the app hash of all potential chain-ids already. I think the goal should be to leverage the actual connections as the source of updated information.

I think first step should be to replicate what the github is trying to do right now:
 * Chain-ID
 * Seed Nodes
 * Whitelist for administration of those values

I think second step should be some way to associate a chain ID with a list of connection IDs (since there can be multiple connections between the same chain). This is a tricky oracle problem. If we're trusting the "admins" of the chain id, we could extend that to their judgement of which connections are legitimate. There should be a way to confirm a second connection once a first one is made as well but I'm not sure exactly what steps are needed.

For simplicity maybe admin can approve manually is a good first step though, and it's just weak subjectivity as a guarantee?

ApeUnit is working on the flow for a wallet or block explorer to confirm an asset displayed with an account query by utilizing the github + tmcrawl/atlas at the moment. I think this should be the initial goal for the chain-registry as well.

docs/cns.md
jackzampolin: I think this would be a cool first usecase for interchain accounts potentially?

docs/cns.md
fadeev: It's the URL of the source code from which the chain is built.

docs/cns.md
fadeev: Commit hash of the source code.

docs/cns.md
fadeev: Yes, I think seed nodes.

docs/cns.md
```diff
@@ -0,0 +1,110 @@
+# Chain Name Service Draft
+
+The goal of this module is to be a simple registry with the ability to transfer ownership. Auctions and other mechanisms are out of scope and can be implemented alongside or on top of this system (perhaps, in a custodial way).
+
+# State
+
+## Chain name
+
+Chain name contains information about the chain, most importantly, chain ID used to uniquly identify the chain and a list of peer nodes.
```
AdityaSripal, line 9: Chain Name is too confusing with Chain-ID imo, I think it can be renamed to `ChainInfo`?

AdityaSripal, line 9:
```suggestion
Chain name contains information about the chain, most importantly, chain ID used to uniquely identify the chain and a list of peer nodes.
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+
+Key: `chain-[chainID] -> Chain Name`
```
AdityaSripal, line 11: Why does this key need to be prefixed if this is the only chains are being stored by the registry?

docs/cns.md
AdityaSripal, line 19:
```suggestion
	SourceCommitHash string
	CanonicalIBCClientId string
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+# Messages
+
+## MsgEditChainName
+
+Chain owner can make edits to the chain name.
+
+```go
+type MsgEditChainName struct {
```
AdityaSripal, line 43: It would be nice to have all of these msgs implemented as IBCPackets so these can be sent from the domain-controlling chain to the registry chain

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+## Chain name
+
+Chain name contains information about the chain, most importantly, chain ID used to uniquly identify the chain and a list of peer nodes.
+
+Key: `chain-[chainID] -> Chain Name` 
+
+```go
+type ChainName struct {
+	ChainID string
+	Owner sdk.AccAddress
```
AdityaSripal, line 16: Given these might eventually be controlled over IBC, this should be a string. The counterparty chains may not have a notion of sdk.AccAddress. The owner may be a channel/port pair.

okwme:
[Notes](https://docs.google.com/document/d/1lkDa1JsTw_LUfTkUGi4lGV7FCrlX8jWgl7k9xtlvntQ/edit?usp=sharing) from the call on Jan 26

cmwaters:
I have a few [notes](https://docs.google.com/document/d/1Bd0ZvDUBRJ4ZGuBReT6RntSk5IAuK1cP5-ljl9dvaEw/edit) of my own proceeding the call on Jan 26th that I would just like to add here

docs/cns.md
cmwaters: I think it would also be beneficial for light clients and state sync that trusted RPC endpoints are added here

docs/cns.md
cmwaters: This could also be the hash of the binary (but this is much the same as commit hash)

cmwaters: I would also think it is necessary to add a genesis hash

docs/cns.md
```diff
@@ -0,0 +1,110 @@
+# Chain Name Service Draft
+
+The goal of this module is to be a simple registry with the ability to transfer ownership. Auctions and other mechanisms are out of scope and can be implemented alongside or on top of this system (perhaps, in a custodial way).
+
+# State
+
+## Chain name
```
okwme, line 7:
```suggestion
## Chain Info
```

docs/cns.md
okwme, line 9:
```suggestion
Chain Info contains information about the chain, most importantly, chain ID used to uniquly identify the chain and a list of peer nodes.
```

docs/cns.md
okwme, line 11:
```suggestion
Key: `[chainID] -> Chain Info` 
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+
+Chain name contains information about the chain, most importantly, chain ID used to uniquly identify the chain and a list of peer nodes.
+
+Key: `chain-[chainID] -> Chain Name` 
+
+```go
+type ChainName struct {
```
okwme, line 14:
```suggestion
type ChainInfo struct {
```

docs/cns.md
okwme: If it were controlled over IBC it would still use an Interchain Account which is represented by `sdk.AccAddress`, right? Is there another format where an IBC channel/port pair would be preferable?

docs/cns.md
okwme, line 17:
```suggestion
	Seeds []seed
```

docs/cns.md
okwme:
```suggestion
	SourceCodeURL string
```

okwme: @cmwaters brings up an important question about the function of the registry:

Should it be a place with infrequent manual updates (seednodes, sourceCodeURL, genesis file) or with frequent manual updates (trusted RPC endpoints). I think it would be good getting some estimates about the expected costs for maintaining a list of RPC endpoints. @alexanderbez is almost done with the node crawler feature of https://github.com/cosmos/atlas. I'm sure we'll be able to get a lot of info of the rate that changes take place from here and can multiply by tx gas estimates to get ar ough idea of how feasible it is to update RPC endpoints. I'd argue this opens up a much higher risk of security exploits occurring as well, as verifying RPC endpoints to be valid is difficult if not impossible as far as I understand.

docs/cns.md
okwme: I could imagine a genesis hash is more valuable than a binary hash. There would ideally be multiple binaries for operating the node anyway, or some customization at a non-state-breaking level. The genesis hash needs to be consistent across them though.

okwme, line 19:
```suggestion
	GenesisHash string
```

docs/cns.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+## Offer
+
+Offers are used to transfer ownership of a chain name.
+
+```go
+type Offer struct {
+	ID int32
```
okwme, line 29:
```suggestion
	ID int32
	ChainId string
```

docs/cns.md
okwme: isn't this possible already with interchain accounts?

docs/cns.md
okwme: I think an oracle might be valuable in other areas more than here (confirming seednodes, potentially RPC endpoints) but in this case it was a USD/Crypto oracle so that prices of new names weren't volatile with the underlying crypto used to claim them.

okwme:
pasting @cmwaters notes here with comments:
> * It might have been done before, but there should be a clear write up of the problems that the chain registry sets out to solve. There seems to me that there are a couple i.e. connocalizing chains, signing security to validators (by ensuring chain uniqueness), promote awareness of chains and ease of use to validators, wallets and nodes by signalling trusted endpoints. This might help narrow down the scope and identify what should be part of the primary phase and what can be added later in secondary and tertiary phases.

https://github.com/tendermint/spn/issues/101
> * I think before any development there should also be a sort of worst case analysis to understand what scenarios that the registry might cause and what could be done to mitigate and avoid it.
>     * What happens if there are fundamental disagreements about the chain.
>     * What happens if the chains security is compromised
>     * etc..

https://github.com/tendermint/spn/issues/102
> * I'm a little bit cautious about relying purely on some pricing mechanism to "clear the market". At least to begin with, bids for a chain name (id) should be vetted with some quorum agreeing that the bid is valid. This is to avoid front-running, name squatting or unnecessarily pumping up the price.

We could take a lesson out of the name service books that already exist:
 * https://docs.ens.domains/
 * https://hsd-dev.org/guides/auctions.html
> * This brings me on to my next point. We should make a clear divide between the static component and dynamic component of a chain name.  A bid is purely for the chain-id with a bidder attaching their respective context to the chain-id whether that be source url, seed nodes, rpc endpoints, ibc channels, genesis hash, or even a home page. But all these (with perhaps the exception of the genesis hash), should be allowed to be updated almost at will by the owner of the chain name. Competing bids would have the same static component (chain id) but different dynamic component.

I agree here. In fact the content may want to borrow on the work of the NFT WG and use DIDs (or IIDs). This would allow the data attached to a name to be generalizable in the same way that a zone record is. It contains any types of record (CNAME, ANAME, TXT) but could contain potentially any key value pairing. It's up to the owner to register the key value pairs that matter. Although probably worthwhile name reserving the ones listed above in some way since they may interface with the IBC module.
> * I like the idea of chains being able to form representative accounts in other chains, whose transactions are controlled by governance. Is this already a thing?

https://github.com/chainapsis/cosmos-sdk-interchain-account
> * Also I agree that trying to imitate existing name service architectures like ENS is a good idea

 * https://docs.ens.domains/
> * Alongside the source url it might be helpful that there be a list of hashes of the accepted binaries that are running on the network. This means that new users can check whether the code/binary they've downloaded is valid (not malicious) before running it

I like this!
> * Note: the notion that hard forks require a new chain id might be challenged in the future with some ongoing work in Tendermint looking to improve upgrade tooling. It's feasible for different application binaries to run with the same chain id and be separated by height only. This may mean that a single network doesn’t need to own a set of chain-id’s but just one

I think this deserves thinking through thoroughly. What does it mean for a chain to upgrade and what are the responsibilities to the previous chain-id.

https://github.com/tendermint/spn/issues/103

docs/cns.md
sahith-narahari: Agree having genesis hash is more useful, but I also feel it's better to include a commit hash as going forward chains using upgrade module will not have genesis files being changed but just the binaries

docs/cns.md
cmwaters: True. Trusted is probably not the right word and this is also more a thing of convenience, a nice to have, then something that is core to the chain registry. I have just heard in a few separate instances that finding a good set of RPC endpoints is a bit of a pain point for new users. This might be redundant if crawling is effective enough or if light client and state sync can also work via p2p

docs/cns.md
ltacker: > We also have a generalized oracle module I'm working on...

It would be a great solution.
I think we should distinguish in the chainInfo, the info that can be arbitrarily decided by the chain owning entity (source code URL, canonical IBC client ID) and what can be trustlessly verified (seed list, genesis hash?) and having this managed by an oracle system

docs/cns.md
ltacker: This is the prefix from the root of the module key. I think it's good to always have a prefix for a specific entry in the store even if it's the only data for the module.

ltacker:
Updated the draft from our latest discussions.

The development of CNS would be organized into 3 phases. The draft gives specs and goals for phase 0.

fadeev:
Moved to https://github.com/tendermint/cns

PR #118, opened 2021-03-17
refactor: rename genesis module to launch

lubtd:
Merging this change will break `starport network` functionalities. We must implement https://github.com/tendermint/starport/issues/885 test it with this branch and merge both.


lubtd:
I tested it with https://github.com/tendermint/starport/pull/889 but running it locally and commands like `starport network chain list --spn-api-address "http://0.0.0.0:1317" --spn-faucet-address "http://0.0.0.0:4500" --spn-node-address "http://0.0.0.0:26657"`
Once you think it's ready we can merge this PR and then update https://github.com/tendermint/starport/pull/889 dependency and merge it

fadeev (approved):
Tested locally with Starport, worked for me ✅

(this was actually the first time I connected them locally, so easy with Starport 😃)

ilgooz (approved):

PR #127, opened 2021-03-22
feat: chain with custom initial genesis

lubtd:
`chain` has a new field for the initial genesis.
For now, this field can either determine that we use the default genesis generated by the binary or that we use a custom genesis URL

```
message Chain {
  string chainID = 1;
  string creator = 2;
  repeated string peers = 3; 
  string sourceURL = 4; 
  string sourceHash  = 5; 
  int64 createdAt = 6;
  bool final = 7;
  InitialGenesis initialGenesis = 8;
}

enum InitialGenesisType {
  DEFAULT = 0;
  URL = 1;
}

message InitialGenesis {
  oneof source {
    DefaultInitialGenesis defaultGenesis = 1;
    GenesisURL genesisURL = 2;
  }
}

message DefaultInitialGenesis {}

message GenesisURL {
  string url = 1;
  string hash = 2;
}
```

The type of the initial genesis can be retrieve with:

```
chain.InitialGenesis.GetType() // InitialGenesisType_DEFAULT or InitialGenesisType_URL
```

If the type of initial genesis is a genesis URL:

```
genesisURL = chain.InitialGenesis.GenesisURL()
genesisURL.Url
genesisURL.Hash
```

The hash is the SHA 256 hash of the genesis content

```
genesisURL.Hash == types.GenesisURLHash(<fetched content from genesisURL.Url>)
```

The `MsgChainCreate` has a `genesisURL` and `genesisHash` fields. If `genesisURL` is empty a chain with default genesis is created otherwise a GenesisURL is created with the URL and the hash.
The handler verifies the hash has the correct length but doesn't check if the hash represents the content since this operation is not deterministic. The users will have to verify the chain integrity

## Testing

`spnd tx launch create-chain hihi haha.com hoho --from user1 --keyring-backend test --chain-id spn`

```
chain:
  chainID: hihi
  createdAt: "1616428718"
  creator: cosmos12q3s45p646hedkcnargym2cfknx7r8csfyz6su
  final: false
  initialGenesis: {}
  peers: []
  sourceHash: hoho
  sourceURL: haha.com
```

`spnd tx launch create-chain hihihi haha.com hoho --from user1 --keyring-backend test --chain-id spn --genesis-url https://raw.githubusercontent.com/cosmos/mainnet/master/genesis.cosmoshub-1.json`

```
chain:
  chainID: hihihi
  createdAt: "1616428931"
  creator: cosmos12q3s45p646hedkcnargym2cfknx7r8csfyz6su
  final: false
  initialGenesis:
    genesisURL:
      hash: "s�f�\x17#��(��YQ��:��D6P�m���U�^���"
      url: https://raw.githubusercontent.com/cosmos/mainnet/master/genesis.cosmoshub-1.json
  peers: []
  sourceHash: hoho
  sourceURL: haha.com
```


x/launch/types/genesis_url.go
```diff
@@ -0,0 +1,48 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+)
+
+const HashLength = 64
+
+// NewGenesisURL returns a genesis URL from the URL
+func NewGenesisURL(url string, hash string) (gu GenesisURL, err error) {
+	if len(url) == 0 {
```
ilgooz, line 13:
```suggestion
	if url == ""
```

x/launch/types/genesis_url.go
```diff
@@ -0,0 +1,48 @@
...
+		return gu, errors.New("url cannot be empty")
+	}
+	if len(hash) != HashLength {
+		return gu, errors.New("hash must be 32 bytes")
```
ilgooz, line 17:
```suggestion
		return gu, fmt.Errorf("hash must be %d bytes", HashLength)
```

x/launch/types/genesis_url.go
```diff
@@ -0,0 +1,48 @@
...
+	return GenesisURL{
+		Url: url,
+		Hash: hash,
+	}, nil
+}
+
+
+// NewGenesisURLFromContent returns a genesis URL from the URL, hash is automatically computed from the content
+func NewGenesisURLFromContent(url string, content string) (gu GenesisURL, err error) {
+	if len(url) == 0 {
```
ilgooz, line 29:
```suggestion
	if if url == ""
```

x/launch/types/genesis_url.go
```diff
@@ -0,0 +1,48 @@
...
+		return gu, errors.New("url cannot be empty")
+	}
+	if len(content) == 0 {
```
ilgooz, line 32:
```suggestion
	if content == "" {
```

x/launch/types/initial_genesis.go
```diff
@@ -0,0 +1,49 @@
...
+
+	ig.Source = &InitialGenesis_GenesisURL{
+		GenesisURL: &genesisURL,
+	}
+
+	return &ig
+}
+
+// GetType returns the type of the initial genesis
+func (ig InitialGenesis) GetType() (InitialGenesisType, error) {
```
ilgooz, line 28: These enums seems to be defined in proto but never used in the source code or by grpc apis.

x/launch/types/initial_genesis.go
```diff
@@ -0,0 +1,49 @@
...
+		return InitialGenesisType_DEFAULT, nil
+	case *InitialGenesis_GenesisURL:
+		return InitialGenesisType_URL, nil
+	default:
+		return InitialGenesisType_DEFAULT, errors.New("unrecognized initial genesis type")
+	}
+}
+
+// GenesisURL returns the genesis URL of the initial genesis if it's its type
+func (ig InitialGenesis) GenesisURL() (gURL GenesisURL, err error) {
```
ilgooz, line 40: Is this getter used anywhere?

x/launch/client/cli/tx.go
```diff
@@ -34,6 +35,8 @@ func GetTxCmd() *cobra.Command {
 	return cmd
 }
 
+const genesisURLFlag = "genesis-url"
```
ilgooz, line 38:
```suggestion
const genesisURLFlag = "genesis"
```

fo consistency with starport.

x/launch/types/genesis_url.go
lubtd: sha256 is 32 bytes the hexa encoding is 64

x/launch/types/initial_genesis.go
lubtd: We can remove those functions if you don't use them in `starport` :)

fadeev (approved):
Tested locally, looks good!

x/launch/types/genesis_url.go
ilgooz: yes, but the error is misleading. we can do something like `hex encoded sha256 hash must to be 64 bytes`.

ilgooz (dismissed):
👍

fadeev (approved):

PR #132, opened 2021-04-06
refactor: encode all addresses as string

lubtd:
Encode address of message creators and values for proposals (account and validator) as string instead of `AccAddress` or `ValAddress`

Change to do on `starport`:
- Use string for messages and accounts and validators in proposal payloads
- In proposal verification: when statically checking the gentx: compare the value `delegator_address` with the validator address instead of the `validator_address` (because this value is no longer converted into the validator address format)


x/identity/keeper/keeper.go
```diff
@@ -75,9 +75,8 @@ func (k Keeper) GetUsernameFromAddress(ctx sdk.Context, address sdk.AccAddress)
 }
 
 // GetIdentifier returns a string that uniquely identities the user of the corresponding address
-func (k Keeper) GetIdentifier(ctx sdk.Context, address sdk.AccAddress) (string, error) {
-	// We return the address since its the only data taht identifies the user
-	return address.String(), nil
+func (k Keeper) GetIdentifier(ctx sdk.Context, address string) (string, error) {
```
ilgooz, line 78: Do we need this func at all?

ilgooz (approved):

fadeev (approved):
utACK 👍

x/identity/keeper/keeper.go
lubtd: We can keep it to not change the `identity` module interface. We'll certainly remove this module in the future and use a better solution for identity like DID

PR #134, opened 2021-04-18
refactor: no on-chain checks for proposal approvals

lubtd:


ilgooz (approved):

fadeev (approved):
💯 before merging let's make sure that there is no `juno-testnet-5` at the time.

PR #144, opened 2021-07-14
chore: initialize application

lubtd:
Initialize the app with `starport s chain`

- No default module is scaffolded
- `spn` is used as an address prefix


fadeev (approved):
And so it begins! 🚀

Pantani (approved):

PR #147, opened 2021-07-15
chore: initialize `account` and `launch` modules

lubtd:
- Rename `AccountKeeper` of `auth` into `AuthKeeper` in `app.go` to avoid collision with `account` module
- Create the `account` module
- Create the `launch` module that has `account` as a dependency


Pantani (approved):

fadeev (approved):
Love how we're using Starport CLI features like `--dep` ✨

PR #168, opened 2021-07-16
feat(launch): initialize `chain`

lubtd:
Initialize the `chain` structure for the `launch` module.

Define it as a map that is indexed by the `ChainID`


fadeev (approved):
👍

Pantani (approved):

PR #169, opened 2021-07-16
feat(account): initialize the coordinator struct

### LINKED ISSUE #145: account: initialize the coordinator struct
Create an internal map for the coordinator with the necessary index

We must provide `—-no-message` option since we want to implement our own message to create and edit the coordinator

```go
starport list coordinator ... --no-message --module account
```

An entry must also be manually created in the store to allow indexing the coordinator ID from the address of the coordinator

Pantani:
Closes #145

### What does this PR does?

- Create `Coordinator` structure;
- Create `CoordinatorByAddress` structure;
- Add tests for mapping structures with custom index (`CoordinatorByAddress`)


proto/account/coordinator_by_address.proto
```diff
@@ -0,0 +1,5 @@
+syntax = "proto3";
```
lubtd, line 1: Let's remove this file

proto/account/query.proto
```diff
@@ -4,12 +4,69 @@ package tendermint.spn.account;
...
 service Query {
     // this line is used by starport scaffolding # 2
+
+	// Queries a coordinatorByAddress by index.
+	rpc CoordinatorByAddress(QueryGetCoordinatorByAddressRequest) returns (QueryGetCoordinatorByAddressResponse) {
+		option (google.api.http).get = "/tendermint/spn/account/coordinatorByAddress/{address}";
+	}
+
+	// Queries a list of coordinatorByAddress items.
+	rpc CoordinatorByAddressAll(QueryAllCoordinatorByAddressRequest) returns (QueryAllCoordinatorByAddressResponse) {
```
lubtd, line 22: I think we can manually remove this query with everything related to it.

The benefit of this `coordinatorByAddress` structure is to get the coordinator from the address but if we want to get all of them, we can use the query generated for `coordinator`

x/account/types/genesis.go
```diff
@@ -22,6 +24,25 @@ func (gs GenesisState) Validate() error {
 	// this line is used by starport scaffolding # ibc/genesistype/validate
 
 	// this line is used by starport scaffolding # genesis/types/validate
+	// Check for duplicated index in coordinatorByAddress
+	coordinatorByAddressIndexMap := make(map[string]struct{})
+
+	for _, elem := range gs.CoordinatorByAddressList {
```
lubtd, line 30: This function is used to verify if a genesis file is valid and can be used for a chain. The check is static and depends only on the file itself.
By default, we only add a check that verifies that no duplicated index keys exist in the provided lists.

Additional steps we can do for genesis validation based on our structure is:
- all IDs retrieved from the `CoordinatorByAddress` must exist in `Coordinator`
- All addresses retrieved from the `Coordinator` must exist in `CoordinatorByAddress`

We we can simply reuse the maps in this file and add those verifications

lubtd (commented):
🚀

lubtd:
I renamed the PR titles. Any PR where we create a new component or a new message can be considered as a `feat`

fadeev (approved):

x/account/types/genesis.go
```diff
@@ -23,5 +25,32 @@ func (gs GenesisState) Validate() error {
...
+	// Check for duplicated index in coordinatorByAddress
+	coordinatorByAddressIndexMap := make(map[string]uint64)
+	for _, elem := range gs.CoordinatorByAddressList {
+		index := string(CoordinatorByAddressKey(elem.Address))
+		if _, ok := coordinatorByAddressIndexMap[index]; ok {
+			return fmt.Errorf("duplicated index for coordinatorByAddress")
+		}
+		coordinatorByAddressIndexMap[index] = elem.CoordinatorId
+	}
+
```
lubtd, line 37: Can we create `genesis_test.go` to test it correctly fails if there is inconsistencies with `Coordinator` and `CoordinatorByAddress`?

We don't generate one with the command since duplication check is a bit trivial but uit may be good for the future to always create tests when we have some more advance logic with genesis

lubtd: I created one for the `launch` module if you want some inspiration: https://github.com/tendermint/spn/blob/feat/init-genesis-account/x/launch/types/genesis_test.go

x/account/types/genesis_test.go
```diff
@@ -0,0 +1,98 @@
...
+		err      error
+	}{
+		{
+			name:     "default is valid",
+			genState: DefaultGenesis(),
+		}, {
+			name: "valid custom genesis",
+			genState: &GenesisState{
+				CoordinatorByAddressList: []*CoordinatorByAddress{
+					{CoordinatorId: 10, Address: "cosmos1c7gh3kejxm3pzl8fwe65665xncs24x5rl7a8sm"},
```
lubtd, line 24: The prefix for the addresses on Starport Network are actually `spn` instead of `cosmos`

lubtd: Also we may have many tests in the future that necessitate to use addresses. Maybe we can add helpers in `testutil` package to provide the address.

We used to use this method in precious spn version to get a randomized address: https://github.com/tendermint/spn/blob/cd4de2f915a75e8fd67b2dfeb1e4c023fbb1349f/internal/testing/mock.go#L37

Pantani: I like the idea of having some fixtures for future tests, not only addresses but coordinators, chain, validators...

lubtd (dismissed):
👍

 We can address lint issues in a separate PR [EDIT] I haven't seen your other PR :)

x/account/types/genesis_test.go
lubtd: What do you mean by fixtures?

Pantani: sorry! fixtures are the same of the create some data for the tests, but the name is more used in the Python language

lubtd: Ok I didn't know👍 FYI I created such a package in another PR https://github.com/tendermint/spn/blob/feat/init-genesis-account/testutil/sample/sample.go

fadeev (dismissed):

fadeev (approved):

x/account/types/genesis_test.go
```diff
@@ -0,0 +1,98 @@
...
+			err: nil,
+		}, {
+			name: "duplicated coordinator",
+			genState: &GenesisState{
+				CoordinatorByAddressList: []*CoordinatorByAddress{
+					{CoordinatorId: 10, Address: "spn1c7gh3kejxm3pzl8fwe65665xncs24x5rl7a8sm"},
+					{CoordinatorId: 10, Address: "spn1c7gh3kejxm3pzl8fwe65665xncs24x5rl7a8sm"},
+				},
+			},
+			err: fmt.Errorf("duplicated index for coordinatorByAddress: spn1c7gh3kejxm3pzl8fwe65665xncs24x5rl7a8sm"),
```
lubtd, line 45: I'm not strongly against providing the exact error in the assertion, it still make the verification more detailed. Although, I think providing just a boolean `isValid` is sufficient. The main purpose of `Validate` tests is to ensure we don't have false positive that could be really problematic in production

lubtd (approved):
👍

x/account/types/genesis_test.go
Pantani: I can remove it later, I like to assert the error because we can assert all possible options, and it's clearer for someone who is reading the code. But we can avoid it in the next ones

PR #172, opened 2021-07-17
feat(profile): allows to create a new coordinator 

### LINKED ISSUE #146: account: allows to create a new coordinator account
Implement a message to create a new coordinator in the store

```go
starport message create-coordinator ... -r coordinatorID:uint --module account
```

This message must create the coordinator entry in the store, map its address with the `CoordinatorByAddress` structure and returns the generated ID for the coordinator

Pantani:
Closes #146

### What does this PR does?

- Create `create-coordinator` structure;
- Implemente `CreateCoordinator` method;
- Map the address with the `CoordinatorByAddress` structure;
- Add `CreateCoordinator` test (`TestMsgCreateCoordinator`);


proto/account/tx.proto
```diff
@@ -8,6 +8,18 @@ option go_package = "github.com/tendermint/spn/x/account/types";
...
+  rpc CreateCoordinator(MsgCreateCoordinator) returns (MsgCreateCoordinatorResponse);
 }
 
-// this line is used by starport scaffolding # proto/tx/message
\ No newline at end of file
+// this line is used by starport scaffolding # proto/tx/message
+message MsgCreateCoordinator {
+  string creator = 1;
+  string address = 2;
+  string identity = 3;
```
lubtd, line 18: We can use a single `CoordinatorDescription` field. Having nested type is not a problem for the message.

Although we can't provide this information in the CLI so we must modify the type manually afterward

x/account/client/cli/tx_create_coordinator.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/tendermint/spn/x/account/types"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdCreateCoordinator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create-coordinator [address] [identity] [website] [details]",
```
lubtd, line 17: `identity`, `website` and `details` are optional information to provide information about the coordinator.

We should provide these values through flags instead of having positional parameters

lubtd: `address` will generally be the signer of the message, so I think we can have `clientCtx.GetFromAddress().String()` by default and provide a flag in case the sender  want to provide a custom address that is the coordinator

x/account/keeper/coordinator.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"encoding/binary"
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/tendermint/spn/x/account/types"
+	"strconv"
+)
+
+// GetCoordinatorCount get the total number of TypeName.LowerCamel
```
lubtd, line 11: `TypeName.LowerCamel` seems like we have a templating issue here 😄 We must create an issue for it

Pantani: fixed, and issues created
https://github.com/tendermint/starport/issues/1390

x/account/keeper/msg_server_create_coordinator.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func (k msgServer) CreateCoordinator(
+	goCtx context.Context,
+	msg *types.MsgCreateCoordinator,
+) (*types.MsgCreateCoordinatorResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Validate the coordinator address
+	_, err := sdk.AccAddressFromBech32(msg.Address)
```
lubtd, line 20: We don't need this check since it's already performed here: https://github.com/tendermint/spn/blob/b87cc435681cfe48f2d8ed8f5936ddcd54c3c07c/x/account/types/message_create_coordinator.go#L42

More generally we should keep all stateless checks under `types/` and checks that depend on the state of the app under `keeper/`

x/account/keeper/msg_server_create_coordinator.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return &types.MsgCreateCoordinatorResponse{},
+			status.Error(codes.InvalidArgument,
```
lubtd, line 23: For errors, we should use the standard package of Cosmos SDK for this: https://github.com/cosmos/cosmos-sdk/tree/master/types/errors

When we have custom error that can be recurrent, we should register them ourself: https://github.com/tendermint/spn/blob/master-legacy/x/launch/types/errors.go

x/account/keeper/msg_server_create_coordinator_test.go
```diff
@@ -0,0 +1,79 @@
...
+					Identity: "identity_invalid_address",
+					Website:  "website_invalid_address",
+					Details:  "details_invalid_address",
+				},
+			},
+			err: status.Error(codes.InvalidArgument,
+				"invalid coordinator address (invalid_address): decoding bech32 failed: invalid index of 1"),
+		}, {
+			name: "valid coordinator 1",
+			msg: types.MsgCreateCoordinator{
```
lubtd, line 34: We can create a new file `sample/profile` to return samples of structure from the `profile` module (where here you would provide the address as a parameter and it would randomize the description) so it simplifies tests writing

lubtd (commented):
Cool👍 
We need to use the standard package of Cosmos SDK for errors.

Also we will need https://github.com/tendermint/spn/pull/185 merged to correctly manually test the feature

x/account/keeper/msg_server_create_coordinator_test.go
lubtd: https://github.com/tendermint/spn/blob/feat/init-vested-account/testutil/sample/launch.go

Pantani: I agree, but for this tests, we can create an import cycle using the sample and returning a `MsgCreateCoordinator`

lubtd (approved):

x/account/keeper/msg_server_create_coordinator_test.go
lubtd: This is not a blocker for this PR but we should try to use `package keeper_test` as much as possible for this problem

https://github.com/tendermint/spn/issues/191

lubtd (approved):

x/profile/client/cli/tx_create_coordinator.go
```diff
@@ -0,0 +1,66 @@
...
+const (
+	flagIdentity = "identity"
+	flagWebsite  = "website"
+	flagDetails  = "details"
+)
+
+func CmdCreateCoordinator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create-coordinator",
+		Short: "Broadcast message create-coordinator",
```
lubtd, line 24:
```suggestion
		Short: "Create a new coordinator profile",
```

`message` command actually has a `--desc` flag to precise a description of the message other than the default generated one

lubtd (approved):

fadeev (approved):
LGTM

PR #173, opened 2021-07-17
feat(Makefile): Makefile commands and build version

Pantani:
### What does this PR does?

- Create `Makefile` with useful commands;
- Add version package to print current build version;

#### How to test

```shell
$ make help
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,62 @@
+#! /usr/bin/make -f
+
+# Project variables.
+PROJECT_NAME = spn
+DATE := $(shell date '+%Y-%m-%dT%H:%M:%S')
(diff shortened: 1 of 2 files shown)


fadeev:
`Makefile`, that's **bold** 😁 I'm ok with having it temporarily until we figure out a way to include this functionality into Starport. Ideally, `starport serve` should do most of the work with some help from additional commands.

lubtd (approved):
I don't have a strong opinion against having it. I think is practical to have `lint`, `test-unit` and `format` executor inside a single file and those will not be handled by Starport

fadeev (approved):

PR #174, opened 2021-07-20
feat(profile): allows to update a coordinator description

### LINKED ISSUE #148: account: allow to update coordinator description
Implement a message to update a coordinator information description

```go
starport message update-coordinator-description ... --module account
```

Pantani:
Closes #148

### What does this PR does?

- Create `update-coordinator-description` structure;
- Implemente `UpdateCoordinatorDescription` method;
- Add `UpdateCoordinatorDescription` test (`TestMsgUpdateCoordinatorDescription`);


proto/profile/tx.proto
```diff
@@ -21,3 +23,12 @@ message MsgCreateCoordinator {
...
   uint64 coordinatorId = 1;
 }
+
+message MsgUpdateCoordinatorDescription {
+  string address = 1;
+  CoordinatorDescription description = 2;
+}
+
+message MsgUpdateCoordinatorDescriptionResponse {
+  uint64 coordinatorId = 1;
```
lubtd, line 33:
```suggestion
```

We can have an empty response since we already know the coordinator ID at this point or it can be queried with `coordinator-from-address`

The message response should contain data that are generated from the message logic

x/profile/keeper/msg_server_update_coordinator_description.go
```diff
@@ -0,0 +1,39 @@
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Check if the coordinator address is already in the store
+	coordByAddress, found := k.GetCoordinatorByAddress(ctx, msg.Address)
+	if !found {
+		return &types.MsgUpdateCoordinatorDescriptionResponse{},
+			sdkerrors.Wrap(types.ErrCoordAddressNotFound, msg.Address)
+	}
+
+	coord := k.GetCoordinator(ctx, coordByAddress.CoordinatorId)
```
lubtd, line 24: If we can't find the coordinator while the address mapping exist in the store, this is an inconsistency issue

Inconsistency issue, that theorically cannot occur from the design of the app, must be handled with a `panic`

x/profile/types/codec.go
```diff
@@ -11,12 +11,15 @@ import (
...
 func RegisterCodec(cdc *codec.LegacyAmino) {
 	// this line is used by starport scaffolding # 2
+	cdc.RegisterConcrete(&MsgUpdateCoordinatorDescription{}, "profile/UpdateCoordinatorDescription", nil)
 	cdc.RegisterConcrete(&MsgCreateCoordinator{}, "profile/CreateCoordinator", nil)
-
 }
 
 func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 	// this line is used by starport scaffolding # 3
+	registry.RegisterImplementations((*sdk.Msg)(nil),
```
lubtd, line 20: All registration are separated to make the file modification easier from Starport but we can have:
```
registry.RegisterImplementations((*sdk.Msg)(nil),
    &MsgCreateCoordinator{},
    &MsgUpdateCoordinatorDescription{},
)
```

x/profile/keeper/msg_server_update_coordinator_description.go
```diff
@@ -0,0 +1,40 @@
...
+	if (coord == types.Coordinator{}) {
+		panic("Inconsistency error: coordinator id not exist into the keeper")
```
lubtd, line 26:
```suggestion
		panic("a coordinator address is associated to a non-existent coordinator ID")
```
We don't need `Inconsistency error:` since the system panicking implies a serious issue

lubtd (dismissed):

ilgooz (dismissed):

x/profile/keeper/msg_server_update_coordinator_description.go
```diff
@@ -0,0 +1,40 @@
...
+	// Check if the coordinator address is already in the store
+	coordByAddress, found := k.GetCoordinatorByAddress(ctx, msg.Address)
+	if !found {
+		return &types.MsgUpdateCoordinatorDescriptionResponse{},
+			sdkerrors.Wrap(types.ErrCoordAddressNotFound, msg.Address)
+	}
+
+	coord := k.GetCoordinator(ctx, coordByAddress.CoordinatorId)
+	if (coord == types.Coordinator{}) {
+		panic("a coordinator address is associated to a non-existent coordinator ID")
```
ilgooz, line 26: if there is a chance that coordinator will not be exists in a handler, we should return with an error instead to let client know what actually did went wrong.

lubtd: The coordinator ID returned in `GetCoordinatorByAddress` must always exist in the store. If it doesn't, this means that the blockchain logic is broken

ilgooz: Hm, what does it mean blockchain logic is broken, can you demonstrate me a case how this might happen?

lubtd: The state of the blockchain can only be modified through the defined messages. If we have an impossible state from the defined messages it means the blockhain logic is broken.
In the case, wit our messages `MsgCreateCoordinator`, `MsgUpdateCoordinatorAddress`, etc... a `Coordinator` in the store is ALWAYS associated with a `CoordinatorByAddress`, if one exists without the other, it means one of the message has a vulnerability

x/profile/keeper/msg_server_update_coordinator_description.go
lubtd:
```suggestion
	coord, found := k.GetCoordinator(ctx, coordByAddress.CoordinatorId)
```

I didn't notice.
The `Get<type>` method returns a boolean if the value is found or not. We must check this value to determine if the coordinator exist

x/profile/keeper/msg_server_update_coordinator_description_test.go
```diff
@@ -0,0 +1,83 @@
...
+			require.NoError(t, err)
+			coordByAddr, found := k.GetCoordinatorByAddress(ctx, tt.msg.Address)
+			assert.True(t, found, "coordinator by address not found")
+
+			coord := k.GetCoordinator(ctx, coordByAddr.CoordinatorId)
+			assert.True(t, found, "coordinator id not found")
+			assert.EqualValues(t, tt.msg.Address, coord.Address)
+			assert.EqualValues(t, coordByAddr.CoordinatorId, coord.CoordinatorId)
+
+			if len(tt.msg.Description.Identity) > 0 {
```
lubtd, line 71: We should add a condition that the value remain unchanged if the field in the message is empty.
It would imply fetching the value and parsing the fields before sending the message with `srv.UpdateCoordinatorDescription(wCtx, &tt.msg)`

```
if tt.err != nil {
  coordByAddr, found := k.GetCoordinatorByAddress(ctx, tt.msg.Address)
  coord := k.GetCoordinator(ctx, coordByAddr.CoordinatorId)
  oldIdentity, oldWebsite,... := coord.Identity, coord.Website, ...
}
```

lubtd (changes requested):
Great👍 We must use the return `found` boolean for get methods

x/profile/keeper/msg_server_update_coordinator_description_test.go
Pantani: great, nice catch. I made an implementation based on your suggestion

x/profile/keeper/msg_server_update_coordinator_description_test.go
```diff
@@ -0,0 +1,95 @@
...
+			require.EqualValues(t, tt.msg.Address, coord.Address)
+			require.EqualValues(t, coordByAddr.CoordinatorId, coord.CoordinatorId)
+
+			if len(tt.msg.Description.Identity) > 0 {
+				require.EqualValues(t, tt.msg.Description.Identity, coord.Description.Identity)
+			} else {
+				require.EqualValues(t, oldCoord.Description.Identity, coord.Description.Identity)
+			}
+
+			if len(tt.msg.Description.Website) > 0 {
```
ilgooz, line 82:
```suggestion
			if tt.msg.Description.Website != "" {
```

ilgooz (dismissed):

x/profile/keeper/msg_server_update_coordinator_description_test.go
Pantani: Both are idiomatic and are clear enough. It is more a matter of personal taste and about clarity. Go compiler does generate identical code in both cases

ilgooz: I'm no worried about what compiler generates actually, just found it easier to read in the second way, but both ok.

lubtd: No preference on my side

x/profile/keeper/msg_server_update_coordinator_description_test.go
```diff
@@ -0,0 +1,95 @@
...
+			_, err := srv.UpdateCoordinatorDescription(wCtx, &tt.msg)
+			if tt.err != nil {
+				require.ErrorIs(t, err, tt.err)
+				return
+			}
+			require.NoError(t, err)
+
+			coordByAddr, found := k.GetCoordinatorByAddress(ctx, tt.msg.Address)
+			coord := k.GetCoordinator(ctx, coordByAddr.CoordinatorId)
+			require.True(t, found, "coordinator id not found")
```
lubtd, line 72: It would be cleaner to have it above `GetCoordinator` otherwise we call `GetCoordinator` with `CoordinatorId` 0 but this is not blocking because the test will fail anyway

lubtd (dismissed):
👍

ilgooz (approved):

lubtd (approved):

PR #176, opened 2021-07-20
feat(profile): allows to delete a coordinator

### LINKED ISSUE #175: account: allow to delete a coordinator
This message is to delete a coordinator account

```go
starport s message delete-coordinator address --module account
```

Pantani:
Closes #175

### What does this PR does?

- Create `delete-coordinator` structure;
- Implemente `DeleteCoordinator` method;
- Add `UpdateCoordinatorDescription` test (`TestMsgDeleteCoordinator`);


proto/profile/tx.proto
```diff
@@ -21,3 +23,10 @@ message MsgCreateCoordinator {
 message MsgCreateCoordinatorResponse {
   uint64 coordinatorId = 1;
 }
+
+message MsgDeleteCoordinator {
+  string address = 1;
+}
+
+message MsgDeleteCoordinatorResponse {
```
lubtd, line 31: Here is the opposite of `MsgUpdateCoordinatorDescription`

It makes sense to add the `CoordinatorID` in the response because in the store this value is removed, we then communicate an information from a change in the store and that cannot be get afterward (`coordinator-from-addr` will not return the id)

x/profile/client/cli/tx_delete_coordinator.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/profile/types"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdDeleteCoordinator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete-coordinator",
+		Short: "Delete a coordinator associate to the sender address",
```
lubtd, line 18:
```suggestion
		Short: "Delete the coordinator profile associated to the sender address",
```

x/profile/keeper/msg_server_delete_coordinator.go
```diff
@@ -0,0 +1,28 @@
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Check if the coordinator address is already in the store
+	coordByAddress, found := k.GetCoordinatorByAddress(ctx, msg.Address)
+	if !found {
+		return &types.MsgDeleteCoordinatorResponse{},
+			sdkerrors.Wrap(types.ErrCoordAddressNotFound, msg.Address)
+	}
+
+	coord := k.GetCoordinator(ctx, coordByAddress.CoordinatorId)
```
lubtd, line 24: We already have the coordinator ID from `GetCoordinatorByAddress`

lubtd: We can actually let this line though, it wasn't mentioned in the specs but wecan do the same as `UpdateCoordinatorDescription`: if the coordinator doesn't exist while the address is registered, we panic

lubtd (dismissed):
⚽️

x/profile/keeper/msg_server_delete_coordinator.go
```diff
@@ -0,0 +1,33 @@
...
+	if (coord == types.Coordinator{}) {
+		panic("a coordinator address is associated to a non-existent coordinator ID")
```
ilgooz, line 26: I think we should return with an error instead. Otherwise, clients wouldn't know what actually went wrong because the content of panic will be kept secret.

lubtd: We use `panic` because this is part of code that should never be reached under any condition. If we reach this, this mean there is a serious inconsistency issue in the chain and it must then be stopped.

This is what is done with Cosmos SDK for this kind of case. You have lot of examples in this file for ex: https://github.com/cosmos/cosmos-sdk/blob/760bde3fbd0b3811a3b7a9ee7d4431544662d5dd/x/bank/keeper/keeper.go#L283

ilgooz: If this case is never reached we don't need to handle it. But if it somehow reaches, then we should let the client know so it can know what actually did went wrong.

lubtd: We can't return an error otherwise the blockchain continue to run.

The whole network must be stopped if we reach that point

ilgooz: What do you mean by whole network needs to be stopped? You mean nodes' process should be stopped or returning with an error is not determenistic and cannot be reproduced among other nodes?

lubtd: Any node that receives the message will be stopped therefore the blockchain will be halted

ilgooz: No, I don't think that's the case. Chains won't get halted when you panic inside the message handler because it's executed in seperate goroutine and later recovered before returning the tx result to the client. Halting it would make the chain vulnerable to attacks anyway as this opens the opportunity of taking the whole network down.

So, I think it's better to just return an error to make the client experience better, panic'ing won't help with anything as it obscures the underlying the error message.

lubtd: > Halting it would make the chain vulnerable to attacks anyway as this opens the opportunity of taking the whole network down.

This is the point, the network must go down if we reach this point

This is the methodology used by Cosmos SDK so far. Another example is the `crisis` module, specifically created to halt the network if an invariant is broken. The `VerifyInvariant` will `panic` if an invariant in the chain is broken: https://github.com/cosmos/cosmos-sdk/blob/760bde3fbd0b3811a3b7a9ee7d4431544662d5dd/x/crisis/keeper/msg_server.go#L50
We must keep synced with Cosmos SDK methodology for this and not implement our own method. If they eventually update this, we must then follow them.

But we must never let the network alive if a severe inconsistency is detected

x/profile/keeper/msg_server_delete_coordinator_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"testing"
+
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/profile/types"
+)
+
+func TestMsgDeleteCoordinator(t *testing.T) {
```
lubtd, line 12: I just realized implementing `MsgCreateChain`

For the tests, we must also assert the state changes in the keeper.
If a coordinator is created, we assert that the coordinator and coordinatorByAddress exist in the keeper
If a coordinator is deleted, it doesn't exist anymore in the store, etc....

We can create a new issue since `MsgCreateCoordinator` has been merged, but we can implement those for the current open PRs for messages

x/profile/keeper/msg_server_delete_coordinator.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// Check if the coordinator address is already in the store
+	coordByAddress, found := k.GetCoordinatorByAddress(ctx, msg.Address)
+	if !found {
+		return &types.MsgDeleteCoordinatorResponse{},
+			sdkerrors.Wrap(types.ErrCoordAddressNotFound, msg.Address)
+	}
+
+	coord := k.GetCoordinator(ctx, coordByAddress.CoordinatorId)
+	if (coord == types.Coordinator{}) {
```
lubtd, line 25: Same comment as `MsgUpdateCoordinatorDescription` we must use `found` boolean

x/profile/keeper/msg_server_delete_coordinator_test.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := srv.DeleteCoordinator(wCtx, &tt.msg)
+			if tt.err != nil {
+				require.ErrorIs(t, err, tt.err)
+				return
+			}
+			require.NoError(t, err)
+			_, found := k.GetCoordinatorByAddress(ctx, tt.msg.Address)
+			assert.False(t, found, "coordinator by address was not removed")
```
lubtd, line 48: We must also check that `Coordinator` doesn't exist from the ID as well by using the ID returned from the message response

lubtd (changes requested):

x/profile/keeper/msg_server_delete_coordinator.go
ilgooz: I'm no sure how crisis module halts the chain actually or if halting is relatated to panic'ing inside a message handler. As I checked, chain does not stop and continues to generate blocks even if you panic inside a handler. 

Also, error message inside this panic is never sent to the client so, this is why I suggest to use errors instead. If you try this yourself and panic inside a handler, you'll see that panic's message is obscured. Maybe it's only printed in serverside if you enabled debug mode.

lubtd: Yes, the blockchain actually continue to run. This is not the behavior I was expected to. Thanks for having checked that.

I need to raise the question with engineering, maybe halting chain can only occur through invariants.

lubtd: Let's try to settle this for now.

What about implementing a method:
```
func CritialError(message) error {
  return sdkerrors.Wrap(ErrCritical, message)
}
```

That actually returns a message `CRITICAL: <message>` from the handler

So that whatever the solution we gonna use to handle crisis situation, we can later modify this function

x/profile/keeper/msg_server_delete_coordinator.go
lubtd: Sorry for the ambiguous response. I actually realize that unlike `map`, for `list` we don't have a `found` boolean returned. We must use the method `k.HasCoordinator` instead to check for coordinator existence (same for other PRs)

x/profile/keeper/msg_server_delete_coordinator.go
ilgooz: Looks good to me.

x/profile/keeper/msg_server_delete_coordinator_test.go
```diff
@@ -0,0 +1,54 @@
...
+			name: "delete coordinator",
+			msg:  types.MsgDeleteCoordinator{Address: msgCoord.Address},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := srv.DeleteCoordinator(wCtx, &tt.msg)
+			if tt.err != nil {
+				require.ErrorIs(t, err, tt.err)
+				return
```
ilgooz, line 44:
```suggestion
```

require should already exit the routine

ilgooz (approved):

x/profile/keeper/msg_server_delete_coordinator_test.go
Pantani: I'm not sure. I tried here and it break the test

lubtd (approved):

PR #179, opened 2021-07-20
feat(launch): initialize genesis account

lubtd:
Genesis accounts are the accounts that will be included in the genesis of a specific chain. This is a map indexed by the chain id and the address of the account

- Initialize the genesis account structure for `launch` with the command:
```
starport s map genesisAccount --index chainID,address --no-message --module launch
```
- Add `sdk.Coin` proto definition and include it inside the genesis account type
- Add test for genesis state validation for `launch`
- Utility package for tests


fadeev:
What's the advantage of having `proto/cosmos/coin.proto` as opposed to importing it from standard proto files supplied by Starport?

lubtd:
> What's the advantage of having `proto/cosmos/coin.proto` as opposed to importing it from standard proto files supplied by Starport?

Are they?🤔

I can't find the definitions in the `protoc` package https://github.com/tendermint/starport/tree/develop/starport/pkg/protoc/data/include

lubtd:
Ok I may see we use `protoc-gen-gocosmos`, let me check this

lubtd:
> What's the advantage of having `proto/cosmos/coin.proto` as opposed to importing it from standard proto files supplied by Starport?

Fixed

fadeev (dismissed):
🙌

x/launch/client/cli/query_genesis_account.go
```diff
@@ -0,0 +1,82 @@
...
+	cmd := &cobra.Command{
+		Use:   "show-genesis-account [chainID] [address]",
+		Short: "shows a genesisAccount",
+		Args:  cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+
+			queryClient := types.NewQueryClient(clientCtx)
+
+			argsChainID, err := cast.ToStringE(args[0])
```
Pantani, line 56: Why cast a string to string?

x/launch/client/cli/query_genesis_account_test.go
```diff
@@ -0,0 +1,157 @@
+package cli_test
+
+import (
+	"fmt"
+	"github.com/tendermint/spn/testutil/sample"
```
Pantani, line 5: wrong lint import sort

x/launch/genesis.go
```diff
@@ -23,6 +28,13 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 	genesis := types.DefaultGenesis()
 
 	// this line is used by starport scaffolding # genesis/module/export
+	// Get all genesisAccount
+	genesisAccountList := k.GetAllGenesisAccount(ctx)
+	for _, elem := range genesisAccountList {
+		elem := elem
```
Pantani, line 34:
```suggestion
```

x/launch/keeper/genesis_account.go
```diff
@@ -0,0 +1,68 @@
...
+		genesisAccount.Address,
+	), b)
+}
+
+// GetGenesisAccount returns a genesisAccount from its index
+func (k Keeper) GetGenesisAccount(
+	ctx sdk.Context,
+	chainID string,
+	address string,
+
```
Pantani, line 24:
```suggestion
```

x/launch/client/cli/query_genesis_account.go
lubtd: Those are some cases where generated code is made a bit more complicated in order to simplify the template.

In this example, it simplifies the template because we can add a patter `cast.To<Type>E(...)` instead of having a if statement to handle different type.

But it's actually a good catch because we need this refactor this since it will not be a reliable solution when we will want to handle more advanced type for the field of a map

Pantani (approved):

x/launch/genesis.go
ilgooz: Why do we even use pointers in the GenesisState?

lubtd: `GenesisState` structs are protobuf messages and therefore all non-native type fields are pointers

ilgooz: So what's the benefit of using a pointer actually and what do you mean by non-native type fields?

lubtd: There is no point it just comes from proto. I actually saw that this is the default generation and we can prevent this by appending the option `[(gogoproto.nullable) = false]`

I will create issues for this good catch

PR #183, opened 2021-07-21
feat(launch): initialize genesis validator

lubtd:
Genesis validators are the validators that will be part of the validator set (if sufficient self delegation) when a chain is start. This is a map indexed by the chain id and the address of the validator.

Waiting for #179 to be merged
TODO:

- [x]  Add genesis validation tests


lubtd:
I will wait on https://github.com/tendermint/spn/pull/195 to finish to get the validate genesis refactoring

x/launch/client/cli/query_genesis_validator.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			argsAddress, err := cast.ToStringE(args[1])
+			if err != nil {
+				return err
+			}
+
+			params := &types.QueryGetGenesisValidatorRequest{
+				ChainID: argsChainID,
+				Address: argsAddress,
+			}
```
Pantani, line 68:
```suggestion
			params := &types.QueryGetGenesisValidatorRequest{
				ChainID: args[0],
				Address: args[1],
			}
```

Pantani (dismissed):
🖖🏽

x/launch/client/cli/query_genesis_validator.go
lubtd: That's pretty ugly, we need to address that haha

Pantani (approved):

ilgooz (approved):

PR #190, opened 2021-07-22
feat(profile): init the validator structure

### LINKED ISSUE #187: profile: initialize validator profile
Initialize the validator profile. This is a map indexed by the address of the validator.
This map contains the description of the validator:
```
type ValidatorDescription struct {
	Moniker string
	Identity string
	Website string
	SecurityContact string
	Details string
}
```

We can use Starport from this branch to generate the map with tests: https://github.com/tendermint/starport/pull/1392

Pantani:
Closes #187

### What does this PR does?

- Create `Validators` structure: `Validators: [ValidatorAddress] -> Validator`; 
- Add genesis validation for validators;
- Add genesis validation tests;


proto/profile/coordinator.proto
```diff
@@ -22,3 +22,8 @@ message CoordinatorByAddress {
   uint64 coordinatorId = 2;
 }
 
+message ConsensusKeyNonce {
```
lubtd, line 25: Should be in `validator.proto`

x/profile/client/cli/query_consensus_key_nonce.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cast"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/profile/types"
+)
+
+func CmdShowConsensusKeyNonce() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-consensus-key-nonce [consAddress]",
+		Short: "shows a ConsensusKeyNonce",
```
lubtd, line 16:
```suggestion
		Short: "shows the currently required nonce to sign a consensus key proof of ownership",
```

x/profile/client/cli/query_validator_by_address.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/profile/types"
+)
+
+func CmdListValidatorByAddress() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list-validator-by-address",
+		Short: "list all validatorByAddress",
```
lubtd, line 15:
```suggestion
		Short: "list all validator",
```

proto/profile/query.proto
```diff
@@ -4,62 +4,119 @@ package tendermint.spn.profile;
...
+message QueryGetValidatorByAddressRequest {
+  string address = 1;
+
+}
+
+message QueryGetValidatorByAddressResponse {
+  ValidatorByAddress validatorByAddress = 1;
+}
+
+message QueryAllValidatorByAddressRequest {
```
lubtd, line 82: Since this is the only query to list validators. We can rename it into `AllValidator` and `list-validator` for the CLI

x/profile/client/cli/query_validator_by_address.go
```diff
@@ -0,0 +1,68 @@
...
+	flags.AddPaginationFlagsToCmd(cmd, cmd.Use)
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func CmdShowValidatorByAddress() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-validator-by-address [address]",
+		Short: "shows a validatorByAddress",
```
lubtd, line 49:
```suggestion
		Short: "shows a validator",
```

x/profile/client/cli/query_validator_by_cons_address.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cast"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/profile/types"
+)
+
+func CmdShowValidatorByConsAddress() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-validator-by-cons-address [consAddress]",
+		Short: "shows a validatorByConsAddress",
```
lubtd, line 16:
```suggestion
		Short: "shows a validator from the consensus address used in the chain to launch",
```

That's something I will check as well in the `launch` module as well. We should see if we can make auto-generated description more descriptive for commands

x/profile/types/genesis.go
```diff
@@ -25,6 +28,62 @@ func (gs GenesisState) Validate() error {
...
 	// this line is used by starport scaffolding # genesis/types/validate
 
+	if err := gs.validateValidator(); err != nil {
+		return err
+	}
+
+	return gs.validateCoordinator()
+}
+
+func (gs GenesisState) validateValidator() error {
```
lubtd, line 38:
```suggestion
func (gs GenesisState) validateValidators() error {
```

x/profile/types/genesis.go
```diff
@@ -25,6 +28,62 @@ func (gs GenesisState) Validate() error {
...
+		delete(validatorByConsAddressIndexMap, indexCons)
+	}
+	// Check if all coordinator by address exist
+	for _, address := range validatorByConsAddressIndexMap {
+		return fmt.Errorf("validator not found for consensus address: %s", address)
+	}
+	return nil
+}
+
+func (gs GenesisState) validateCoordinator() error {
```
lubtd, line 86:
```suggestion
func (gs GenesisState) validateCoordinators() error {
```

x/profile/types/genesis.go
```diff
@@ -25,6 +28,62 @@ func (gs GenesisState) Validate() error {
...
+			return fmt.Errorf("duplicated index for validatorByConsAddress: %s", elem.ConsAddress)
+		}
+		validatorByConsAddressIndexMap[index] = elem.ConsAddress
+
+		// Remove to check if all consensus key exist
+		delete(consensusKeyNonceIndexMap, index)
+	}
+	// Check if all consensus key exist
+	for _, address := range consensusKeyNonceIndexMap {
+		return fmt.Errorf("consesus key not found for address: %s", address)
```
lubtd, line 63: The error message mentions "address" and the variable is `consensusKeyNonceIndexMap`. Which one do you want to check?

We actually don't need to check if `consensusKeyNonce` is associated with a `validatorByConsAddress` because theoretically `validatorByConsAddress` could be set in the genesis without signing a proof and so without having to set `consensusKeyNonce` and we can have a nonce value for the signature without an existing consensus pub key in the store.

So `consensusKeyNonce`. can be set in the genesis totally independently from the other values

lubtd: Concerning validator address:
A consensus key must be associated with validator address but the opposite is actually not true.
Validator profiles are used for:
- Giving a description of the validator
- Matching a consensus address on a launched blockchain to a validator address on SPN

Since a validator profile may be used to just provide a description of the validator, we can have a validator that is not associated with a consensus address

Although if the consensus address value is determined for this validator, then it must exist in the store

x/profile/types/genesis_test.go
```diff
@@ -6,9 +6,140 @@ import (
...
+				ConsensusKeyNonceList: []*ConsensusKeyNonce{
+					{ConsAddress: consAddr1, Nonce: 1},
+					{ConsAddress: consAddr2, Nonce: 2},
+				},
+				ValidatorByAddressList: []*ValidatorByAddress{
+					{Address: addr1, ConsensusAddress: consAddr1},
+					{Address: addr1, ConsensusAddress: consAddr2},
+				},
+			},
+			err: fmt.Errorf("consesus key not found for address: %s", consAddr1),
```
lubtd, line 83:
```suggestion
			err: fmt.Errorf("consensus key not found for address: %s", consAddr1),
```

x/profile/types/genesis_test.go
```diff
@@ -6,9 +6,140 @@ import (
...
+					{Address: addr2, ConsensusAddress: consAddr2},
+				},
+				ValidatorByConsAddressList: []*ValidatorByConsAddress{
+					{ConsAddress: consAddr1, Address: addr1},
+					{ConsAddress: consAddr1, Address: addr1},
+				},
+			},
+			err: fmt.Errorf("duplicated index for validatorByConsAddress: %s", consAddr1),
+		}, {
+			name: "validator by consensus address not found",
```
lubtd, line 102: Can we add a case where providing a `ValidatorByAddress` with no `ConsensusAddress` is valid

lubtd (commented):
Nice

Just to mention that we don't need the `ConsensusKeyNonce` and `ValidatorByConsAddr` for now.

Although there is of course no reason to remove those structures now that those have been generated, keep in mind we will not implement the `MsgSetValidatorConsAddress` for now

lubtd:
@Pantani based on some feedbacks from the specs, `ValidatorByConsAddr` is highly likely to be changed.

I finally think it is proper to remove everything related to this (ConsensusKey nonce as well) now so that we can have a atomic PR when we will actually implement this functionality (with Starport generation included in the PR)

ilgooz (commented):
Can we revert the changes in `generated/` and `docs/static/openapi.yml` so it could be easier to review this PR?

proto/profile/query.proto
```diff
@@ -30,6 +40,25 @@ service Query {
 }
 
 // this line is used by starport scaffolding # 3
+
+message QueryGetValidatorByAddressRequest {
+	string address = 1;
+
```
lubtd, line 46:
```suggestion
```

proto/profile/validator.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+package tendermint.spn.profile;
+
+option go_package = "github.com/tendermint/spn/x/profile/types";
+
+message ValidatorByAddress {
+  string address = 1;
+  string consensusAddress = 2;
```
lubtd, line 8:
```suggestion
```

Let's just keep the description

lubtd (commented):
We should rename `ValidatorByAddress` into simply `Validator` since it's the main structure (and only one for now) for `Validator`

x/profile/types/key_validator_by_address.go
```diff
@@ -0,0 +1,23 @@
...
+var _ binary.ByteOrder
+
+const (
+	// ValidatorKeyPrefix is the prefix to retrieve all Validator
+	ValidatorKeyPrefix = "Validator/value/"
+)
+
+// ValidatorKey returns the store key to retrieve a Validator from the index fields
+func ValidatorKey(
+	address string,
```
ilgooz, line 14: Just a general observation, I think we can use multiline arguments when there are more than 3 or 4 args otherwise we can use regular single liner for better code readability.

ilgooz (approved):

ilgooz:
👏

x/profile/types/key_validator_by_address.go
Pantani: Sure. Totally agree!!

Pantani: This is scaffolded by starport. Maybe it's better we also handle this behavior into the starport too?

x/profile/keeper/validator_by_address.go
```diff
@@ -0,0 +1,63 @@
...
+	store.Set(types.ValidatorKey(
+		validator.Address,
+	), b)
+}
+
+// GetValidator returns a validator from its index
+func (k Keeper) GetValidator(
+	ctx sdk.Context,
+	address string,
+
```
Pantani, line 22:
```suggestion
```

ilgooz (approved):

x/profile/types/key_validator_by_address.go
lubtd: Agree.
I put it this way in the template because I find it cleaner to have a plush for statement on two lines rather than one

I think we can a general review of those details at the end of the milestone, removing blank lines, formating proto files, etc... (that of course doesn't prevent us to do it on the fly when we spot those details in the other PRs)

x/profile/client/cli/query_validator.go
```diff
@@ -0,0 +1,68 @@
...
+
+	flags.AddPaginationFlagsToCmd(cmd, cmd.Use)
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func CmdShowValidator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "show-validator-by-address [address]",
```
lubtd, line 48:
```suggestion
		Use:   "show-validator[address]",
```

x/profile/types/key_validator.go
```diff
@@ -0,0 +1,21 @@
...
+
+// ValidatorKey returns the store key to retrieve a Validator from the index fields
+func ValidatorKey(address string) []byte {
+	var key []byte
+
+	addressBytes := []byte(address)
+	key = append(key, addressBytes...)
+	key = append(key, []byte("/")...)
+
+	return key
```
lubtd, line 20: This is another example on where the template produce more complex code than needed because it need to be generate (the template must be able to create the key for an index with multiple different types)

Here we could have simply: `return []byte(address + "/")`

This is not blocking for the PR because I will note that all `<Type>Key(...)` methods will have to be reviewed

ilgooz (approved):

lubtd (approved):

PR #193, opened 2021-07-23
fix(launch): use interface for `InitialGenesis` of `Chain`

lubtd:
`InitialGenesis` of `Chain` is the information that allows to determine what is the initial genesis to use when starting the launch process of a chain

We use `Any` of proto to define this type.

We must associate an interface (empty for now) to define the type that can be used as initial genesis

This PR also improve tests for chains


Pantani (dismissed):

fadeev (dismissed):

ilgooz (approved):

lubtd:
I just solved a small conflict in `go.mod`, I'm merging the PR

PR #195, opened 2021-07-23
feat(launch): create chain message

lubtd:
Implement the message to create chains `MsgCreateChain`

TODO:
- [x] Test `types/chain.go`
- [x] Test the message
- [x] Add InitialGenesis option


x/launch/keeper/chain_name_count.go
```diff
@@ -0,0 +1,63 @@
...
+	store.Set(types.ChainNameCountKey(
+		chainNameCount.ChainName,
+	), b)
+}
+
+// GetChainNameCount returns a chainNameCount from its index
+func (k Keeper) GetChainNameCount(
+	ctx sdk.Context,
+	chainName string,
+
```
Pantani, line 22:
```suggestion
```

x/launch/keeper/chain_name_count.go
```diff
@@ -0,0 +1,63 @@
...
+	k.cdc.MustUnmarshalBinaryBare(b, &val)
+	return val, true
+}
+
+// RemoveChainNameCount removes a chainNameCount from the store
+func (k Keeper) RemoveChainNameCount(
+	ctx sdk.Context,
+	chainName string,
+
```
Pantani, line 41:
```suggestion
```

x/launch/keeper/keeper_test.go
```diff
@@ -1,38 +1,50 @@
 package keeper
 
 import (
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/tendermint/spn/testutil/sample"
+	profilekeeper "github.com/tendermint/spn/x/profile/keeper"
```
Pantani, line 6: The import sort lint is only enabled into the Makefile because the lint can show the starport placeholders as an issue. But maybe we can put back this rule in the CI fixing these cases. We need to add a line after and before all placeholder comment's

x/launch/keeper/msg_server_create_chain_test.go
```diff
@@ -0,0 +1,91 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch/types"
+	"testing"
```
Pantani, line 9:
```suggestion
	"testing"

	sdk "github.com/cosmos/cosmos-sdk/types"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	"github.com/tendermint/spn/testutil/sample"
	"github.com/tendermint/spn/x/launch/types"
```

x/launch/keeper/msg_server_create_chain.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}
+	chainID := types.ChainIDFromChainName(msg.ChainName, chainNameCount.Count)
+	chainNameCount.Count++
+
+	// chainID must always be unique by design
+	// if it already exists then something is wrong in the protocol
+	_, found = k.GetChain(ctx, chainID)
+	if found {
+		panic(fmt.Sprintf("chain id %s already exists", chainID))
```
Pantani, line 37: I think it makes sense to use the critical error here as well

x/launch/keeper/msg_server_create_chain.go
```diff
@@ -0,0 +1,65 @@
...
+		SourceHash:      msg.SourceHash,
+		LaunchTriggered: false,
+		LaunchTimestamp: 0,
+	}
+
+	// Initialize initial genesis
+	if msg.GenesisURL == "" {
+		chain.InitialGenesis = types.AnyFromDefaultInitialGenesis()
+	} else {
+		chain.InitialGenesis = types.AnyFromGenesisURL(msg.GenesisURL, msg.GenesisHash)
```
Pantani, line 55: and we can also use the critical error here to handle invalid genesis

x/launch/keeper/msg_server_create_chain_test.go
```diff
@@ -0,0 +1,91 @@
...
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := srv.CreateChain(ctx, &tc.msg)
+			if !tc.valid {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.EqualValues(t, tc.wantedChainID, got.ChainID)
```
Pantani, line 63:
```suggestion
			require.EqualValues(t, tc.wantedChainID, got.ChainID)
```

x/launch/keeper/msg_server_test.go
```diff
@@ -1,14 +1,29 @@
...
 
 import (
-	"context"
+	"github.com/cosmos/cosmos-sdk/codec"
+	profilekeeper "github.com/tendermint/spn/x/profile/keeper"
+	profiletypes "github.com/tendermint/spn/x/profile/types"
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/tendermint/spn/x/launch/types"
```
Pantani, line 10:
```suggestion
	"testing"

	"github.com/cosmos/cosmos-sdk/codec"
	sdk "github.com/cosmos/cosmos-sdk/types"
	"github.com/tendermint/spn/x/launch/types"
	profilekeeper "github.com/tendermint/spn/x/profile/keeper"
	profiletypes "github.com/tendermint/spn/x/profile/types"
```

x/launch/types/chain_test.go
```diff
@@ -0,0 +1,85 @@
+package types_test
+
+import (
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch/types"
+	"testing"
```
Pantani, line 7:
```suggestion
	"testing"

	"github.com/stretchr/testify/require"
	"github.com/tendermint/spn/testutil/sample"
	"github.com/tendermint/spn/x/launch/types"
```

x/launch/types/genesis.go
```diff
@@ -27,6 +28,16 @@ func (gs GenesisState) Validate() error {
...
+	// Check for duplicated index in chainNameCount
+	chainNameCountIndexMap := make(map[string]struct{})
+
+	for _, elem := range gs.ChainNameCountList {
+		index := string(ChainNameCountKey(elem.ChainName))
+		if _, ok := chainNameCountIndexMap[index]; ok {
+			return fmt.Errorf("duplicated index for chainNameCount")
+		}
+		chainNameCountIndexMap[index] = struct{}{}
+	}
```
Pantani, line 40: maybe we should check if the `ChainNameCountList` object exists in the `ChainlList`

x/launch/types/key_chain_name_count.go
```diff
@@ -0,0 +1,23 @@
...
+
+const (
+	// ChainNameCountKeyPrefix is the prefix to retrieve all ChainNameCount
+	ChainNameCountKeyPrefix = "ChainNameCount/value/"
+)
+
+// ChainNameCountKey returns the store key to retrieve a ChainNameCount from the index fields
+func ChainNameCountKey(
+	chainName string,
+) []byte {
```
Pantani, line 15:
```suggestion
func ChainNameCountKey(chainName string) []byte {
```

x/launch/types/message_create_chain.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (msg *MsgCreateChain) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgCreateChain) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Coordinator)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
```
Pantani, line 45:
```suggestion
		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid coordinator address (%s)", err)
```

x/launch/types/message_create_chain_test.go
```diff
@@ -0,0 +1,46 @@
+package types_test
+
+import (
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch/types"
+	"testing"
```
Pantani, line 7:
```suggestion
	"testing"

	"github.com/stretchr/testify/require"
	"github.com/tendermint/spn/testutil/sample"
	"github.com/tendermint/spn/x/launch/types"
```

Pantani (changes requested):

x/launch/keeper/msg_server_create_chain.go
lubtd: Yep, wasn't merged yet

x/launch/keeper/keeper_test.go
lubtd: > We need to add a line after and before all placeholder comment's

That could be the way to go, I may think we don't have many placeholders in the `import` in the templates. This could be an improvement for the templates we could specify in https://github.com/tendermint/starport/issues/1409

x/launch/keeper/msg_server_create_chain.go
lubtd: I removed those functions because no longer necessary

Also I just put a critical error when `codec.NewAnyWithValue(&types.DefaultInitialGenesis{})` fails but not when `codec.NewAnyWithValue(&types.GenesisURL{...})` fails

DefaultInitialGenesis must succeed because it is a constant and must always be converted into any. So an error is unexpected and critical

For GenesisURL, we have an entry from the message sender (URL and Hash) so it is preferable to just make the message fails because a user must never be able to trigger a critical error

x/launch/types/genesis.go
lubtd: Yes that's a good point.
By the way we should in fact also verify that this count is above all the number of the chain id that contains this name but to do this verification properly we should refactor `chainID` from type `string` into `{chainName: string, number: uint64}`
What we can do for now is to add the check you mentioned and replace the current critical error we have for already existing chainID with a simple error

lubtd: Thanks for spotting this. I'm also adding a check to verify the count is above all chain ID name and check chain ID is correct

ilgooz (dismissed):

Pantani (approved):

x/profile/types/message_create_coordinator_test.go
```diff
@@ -1,29 +1,30 @@
-package types
+package types_test
```
Pantani, line 1: Why change the package name? Import cycle?

x/launch/types/key_chain_name_count.go
```diff
@@ -0,0 +1,21 @@
...
+// ChainNameCountKey returns the store key to retrieve a ChainNameCount from the index fields
+func ChainNameCountKey(chainName string) []byte {
+	var key []byte
+
+	chainNameBytes := []byte(chainName)
+	key = append(key, chainNameBytes...)
+	key = append(key, []byte("/")...)
+
+	return key
+}
```
Pantani, line 21:
```suggestion
func ChainNameCountKey(chainName string) []byte {
	return []byte(chainName+"/")}
}
```

x/profile/types/message_create_coordinator_test.go
lubtd: Yes. The `sample` package need the types but the tests need the samples

Pantani (dismissed):

x/profile/types/message_create_coordinator_test.go
lubtd: And also I believe this is a good practice to not expose the test functions in the main package

x/launch/types/key_chain_name_count.go
Pantani: I followed your suggestion in the other PR. to much better for reading too, but it's not a blocker for this PR

x/profile/types/message_create_coordinator_test.go
Pantani: I like it for some situations, but not works if we want to test a non-exported method (lowercase)

Pantani: usually, I like to use tests with another package name for BlackBox tests, to test only the exported methods, and test with the same package name for white-box tests

lubtd: Yeah, sometime it can be useful. I don't see the necessity for `types` so far. For example, I uppercased the `ValidateValidators` and `ValidateCoordinators` for genesis validation. I don't think this is problematic

lubtd: This is particular to types. I don't think we will need to do this for `keeper`

Pantani (approved):

ilgooz (approved):

PR #198, opened 2021-07-23
profile: allow to update validator description

### LINKED ISSUE #188: profile: allow to update validator description
Create a message to associate validator description with an address.

If the address provided is not associated with an existing validator in the store, the entry is created

Pantani:
Closes #188


x/profile/client/cli/tx_update_validator_description.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/profile/types"
+)
+
+const (
+	flagMoniker         = "moniker"
+	flagSecurityContact = "securityContact"
```
lubtd, line 13:
```suggestion
	flagSecurityContact = "security-contact"
```

kebab case for flags

x/profile/client/cli/tx_update_validator_description.go
```diff
@@ -0,0 +1,70 @@
...
+				securityContact,
+				details,
+			)
+			if err := msg.ValidateBasic(); err != nil {
+				return err
+			}
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)
+		},
+	}
+
```
lubtd, line 66: We must add here the flags with description like in `tx_update_coordinator_description.go`

x/profile/keeper/msg_server_update_validator_description_test.go
```diff
@@ -0,0 +1,91 @@
...
+				Description: sample.ValidatorDescription(addr1),
+			},
+		}, {
+			name: "update a existing validator",
+			msg: types.MsgUpdateValidatorDescription{
+				Address:     addr1,
+				Description: sample.ValidatorDescription(addr2),
+			},
+		}, {
+			name: "update and create anotherw validator",
```
lubtd, line 38:
```suggestion
			name: "update and create another validator",
```

x/profile/types/message_update_validator_description.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return []sdk.AccAddress{address}
+}
+
+func (msg *MsgUpdateValidatorDescription) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgUpdateValidatorDescription) ValidateBasic() error {
```
lubtd, line 51: Another check we can have is to check if at least one field of the description is defined.

The benefits of `ValidateBasic()` is that the message is never spread over the network if it fails and so we can avoid to spread a message that has no side effect

x/profile/types/message_update_coordinator_description.go
```diff
@@ -44,5 +44,14 @@ func (msg *MsgUpdateCoordinatorDescription) ValidateBasic() error {
 	if err != nil {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
 	}
+	if msg.Description == nil {
+		return sdkerrors.Wrapf(ErrEmptyDescription, msg.Address)
+	}
+	desc := []byte(msg.Description.Details)
```
lubtd, line 50: Why doing it this way and not just
```
if msg.Description.Details == "" && 
   msg.Description.Identity == "" &&
 ....
```

x/profile/types/message_update_coordinator_description.go
```diff
@@ -44,5 +44,14 @@ func (msg *MsgUpdateCoordinatorDescription) ValidateBasic() error {
 	if err != nil {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
 	}
+	if msg.Description == nil {
```
lubtd, line 47: Good catch, I haven't realize we hadn't this check for `Coordinator`

lubtd (commented):
When a run the command with no flag it succeed with a `moniker` set to `mynode`:

```
spnd q profile list-validator
pagination:
  next_key: null
  total: "0"
validator:
- address: spn1qaprp8cw8nhesfpvdyz39p8yplvhd9fnjffhlu
  description:
    details: ""
    identity: ""
    moniker: mynode
    securityContact: ""
    website: ""
```

It seems like `moniker` is already a defined global flag we use for the `moniker` of the node (although it doesn't show when printing the help of a command).

We must defined another name like `validator-moniker` to avoid collision

lubtd (approved):

ilgooz (approved):

PR #202, opened 2021-07-27
feat(x/profile): add coordinator count check in genesis validate

### LINKED ISSUE #192: profile: add coordinator count check in genesis validate
`Coordinator` has been scaffolded as a `list`. We must apply some fixes reported for the `list` command here: https://github.com/tendermint/starport/issues/1398

Pantani:
Closes #192

### What does this PR does?

Add genesis validation to check if  the state is inconsistent if the value of the count is not strictly above all ids of the element of the list (since this value is incremented for each new value)

### DIFF
--- a/x/profile/types/genesis.go
+++ b/x/profile/types/genesis.go
@@ -37,12 +37,12 @@ func (gs GenesisState) Validate() error {
 
 	// Check for duplicated ID in coordinator
 	coordinatorIDMap := make(map[uint64]bool)
-	count := len(gs.CoordinatorList)
+	count := gs.GetCoordinatorCount()
 	for _, elem := range gs.CoordinatorList {
 		if _, ok := coordinatorIDMap[elem.CoordinatorId]; ok {
 			return fmt.Errorf("duplicated id for coordinator: %d", elem.CoordinatorId)
 		}
-		if elem.CoordinatorId > uint64(count) {
+		if elem.CoordinatorId >= count {
 			return fmt.Errorf("coordinator id %d should be lower or equal than the last id %d",
 				elem.CoordinatorId, count)
 		}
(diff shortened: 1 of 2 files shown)


ilgooz (dismissed):

x/profile/types/genesis.go
```diff
@@ -37,10 +37,15 @@ func (gs GenesisState) Validate() error {
 
 	// Check for duplicated ID in coordinator
 	coordinatorIDMap := make(map[uint64]bool)
+	count := len(gs.CoordinatorList)
```
lubtd, line 40: We must check with the `coordinatorCount` value in `genesisState` because coordinator can be removed and it will not be taken into account if we do `len(gs.CoordinatorList)`

"count" can be actually a bit ambiguous, this is the count of created coordinators so far and not the actual count of current coordinators

x/profile/types/genesis_test.go
```diff
@@ -21,65 +21,78 @@ func TestGenesisStateValidate(t *testing.T) {
 			name: "valid custom genesis",
 			genState: &GenesisState{
 				CoordinatorByAddressList: []*CoordinatorByAddress{
-					{CoordinatorId: 10, Address: "spn1c7gh3kejxm3pzl8fwe65665xncs24x5rl7a8sm"},
-					{CoordinatorId: 11, Address: "spn12330zcy9yez37lzrkm6d7fedcu7hc279sgkh3c"},
-					{CoordinatorId: 12, Address: "spn1d6pd5nk08mu789q4msfpynsuha7yf4wcsvvspr"},
-					{CoordinatorId: 13, Address: "spn1ktzsme3g0ag0236ngvkw62vy9tqrr3xysnhp3g"},
+					{CoordinatorId: 0, Address: "spn1c7gh3kejxm3pzl8fwe65665xncs24x5rl7a8sm"},
```
lubtd, line 24: Based on the comment above, we must define the value `CoordinatorCount` in each genesis state and put a value below an existing coordinator ID for the test to check this condition

lubtd (changes requested):

ilgooz (approved):

lubtd (approved):

PR #205, opened 2021-07-28
feat(x/profile/keeper): assert test data inside keeper for some msgs

Pantani:
### What does this PR does?

Create a better data assertion for the message handler unit tests.

### Solution

Assert the remain data inside the keeper.

## Checklist

- [x] Include tests
- [x] Update documentation (*.md) (if needed)


x/profile/keeper/msg_server_create_coordinator_test.go
```diff
@@ -24,39 +25,48 @@ func msgCreateCoordinator() types.MsgCreateCoordinator {
...
 			name: "already exist address",
 			msg:  msg2,
 			err:  sdkerrors.Wrap(types.ErrCoordAlreadyExist, "coordinatorId: 1"),
 		},
 	}
-	srv, ctx := setupMsgServer(t)
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := srv.CreateCoordinator(ctx, &tt.msg)
+			_, err := srv.CreateCoordinator(wCtx, &tt.msg)
```
lubtd, line 55: Why removing the check with the ID return as a response?

This is actually relevant to check to response is correct

lubtd (commented):
👍 Let's just keep the check with the returned coordinator ID for MsgCreateCoordinator

Also we have many `require` and `assert` melted in the tests (this is not a comment particularly to this PR) I would advocate, we use `require` everywhere.

We can do this in a separate PR, now the tests fail if the assertion is false anyway

x/profile/keeper/msg_server_create_coordinator_test.go
Pantani: because we already assert the coordinator id from the keeper. but we can keep both

lubtd: When a message response is not empty we should perform tests on it

Pantani: done!!

ilgooz (approved):

lubtd (approved):

PR #208, opened 2021-07-28
feat: implement a `critical` package for critical errors

lubtd:
Create the `pkg` folder and  add the `critical` package to handle critical error.

Do the changes with this package for `MsgUpdateCoordinatorAddress`


Pantani:
maybe it's better we have a common `pkg` for default errors, not only critical. Like `ErrNotFound` or `ErrInvalidArgument`

lubtd:
> maybe it's better we have a common `pkg` for default errors, not only critical. Like `ErrNotFound` or `ErrInvalidArgument`

I don't see other errors to set for now since the common ones are generally define with the `sdkerrors` package but I'm not against calling the package `error` to make it more generic

Pantani (approved):

Pantani (approved):

ilgooz (approved):

PR #216, opened 2021-07-28
feat(launch): request to remove a genesis validator

### LINKED ISSUE #164: launch: request to remove a genesis validator
Create the message to send an remove validator request into the request pool for a chain

### LINKED ISSUE #189: profile: allow to delete a validator profile
Implement a message that will delete the validator profile entry from the store

Pantani:
Closes #164

### What does this PR does?

Scaffold `request-remove-validator` msg;
Implements `RequestRemoveValidator` method and tests;


proto/launch/request.proto
```diff
@@ -14,3 +16,7 @@ message Request {
   google.protobuf.Any content = 5 [(cosmos_proto.accepts_interface) = "RequestContent"];
 }
 
+message ContentRemoveValidator {
```
lubtd, line 19: I would call it `ValidatorRemoval`

The content for `MsgRequestAddValidator` will be `GenesisValidator` and so it could be confusing to have the content prefix for some and not for others

Pantani: sure, I didn't like `ContentRemoveValidator` as well, but I didn't found a better name

x/launch/client/cli/tx_request_remove_validator.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+func CmdRequestRemoveValidator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "request-remove-validator [chainID] [validatorAddress]",
```
Pantani, line 13:
```suggestion
		Use:   "request-remove-validator [chainID]",
```

x/launch/client/cli/tx_request_remove_validator.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+func CmdRequestRemoveValidator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "request-remove-validator [chainID]",
+		Short: "Request to remove a genesis validator",
+		Args:  cobra.ExactArgs(2),
```
Pantani, line 15:
```suggestion
		Args:  cobra.ExactArgs(1),
```

x/launch/types/request_content.go
```diff
@@ -1,4 +1,27 @@
...
+		return nil, fmt.Errorf("empty request content for request %d", r.RequestID)
+	}
+	var content RequestContent
+	err := cdc.UnpackAny(r.Content, &content)
+	if err != nil {
+		return nil, err
+	}
+	removeValidator, ok := content.(*ValidatorRemoval)
+	if !ok {
+		return nil, ErrFailedCodecCast
```
lubtd, line 24: Can we have an error like `not a validatorRemoval request` to be more precise?

lubtd: Also we don't need to register the error in `errors.go` if it's only used in `types` package

lubtd (approved):
👍

lubtd:
Hey based on this comment: https://github.com/tendermint/spn/pull/216/files#r681553965

I realize we can in fact already add this check and test it for every request since we already have the `LaunchTriggered` boolean in chains

x/launch/types/request_content_test.go
```diff
@@ -0,0 +1,32 @@
...
+	require.NoError(t, err)
+	require.EqualValues(t, content, result)
+
+	invalidContent := &types.Request{}
+	request.Content, err = codec.NewAnyWithValue(invalidContent)
+	require.NoError(t, err)
+	content, err = request.UnpackValidatorRemoval(cdc)
+	require.Error(t, err)
+	require.Equal(t, err.Error(), "not a validatorRemoval request")
+	require.Nil(t, content)
```
lubtd, line 31: In a general point of view I don't think it's important to assert the other returned value when a method must return an error and it actually returns the error

x/launch/types/message_request_remove_validator_test.go
```diff
@@ -0,0 +1,58 @@
...
+				"invalid address (invalid_address): decoding bech32 failed: invalid index of 1"),
+		}, {
+			name: "invalid chain id",
+			msg: types.MsgRequestRemoveValidator{
+				ValidatorAddress: sample.AccAddress(),
+				ChainID:          "invalid_chain",
+			},
+			err: sdkerrors.Wrapf(types.ErrInvalidChainID, "invalid_chain"),
+		}, {
+			name: "invalid chain name",
```
lubtd, line 34: I may already have commented this on another PR, what's the difference with the above?

lubtd: Well I may understand that here we use a incorrect name but good format `name-number`

In fact the test above is sufficient because we already have tests for the `ParseChainID` methods with all the different kind of invalid chain ID format.

So we just need here a test where we now the chainID fails with `ParseChainID`

x/launch/keeper/msg_server_request_remove_validator_test.go
```diff
@@ -0,0 +1,89 @@
...
+	tests := []struct {
+		name string
+		msg  types.MsgRequestRemoveValidator
+		want uint64
+		err  error
+	}{
+		{
+			name: "invalid chain",
+			msg: types.MsgRequestRemoveValidator{
+				ChainID: "invalid_chain",
```
lubtd, line 31:
```suggestion
				ChainID: "invalid_chain",
```

Can we use `sample.ChainID(0)` to ensure it fails when we use a valid chainID format

lubtd (approved):

lubtd:
https://github.com/tendermint/spn/pull/219#issuecomment-891961176

Let's talk here if we must continue this discussion.

Those information are only used before the chain is launched. Once the network is live, you no longer interact with the `launch` module. Information about validator in the `launch` module are only verified off-chain so you can potentially have wrong information stored that prevent you from starting the chain. The on-chain stored `GenesisValidator` must be removed in order to make the genesis fetched from `spn` correct and so this `ValidatorRemoval` should be sendable by anyone.

We must then have in the `MsgRequestRemoveValidator` and `MsgRequestRemoveAccount` containing two addresses:
- The address of the signer of the message
- The address of the account to remove (that can be the same as the address above)

I approved the PR because it could may be simpler to implement this additional field on both requests on a single PR once both have been merged.

That's as you want. Can you please create an issue for adding this field if you don't implement it on this PR?

lubtd:
And for auto-approved request from coordinator, I'm going to create a separate issue, it will be better to implement it for all request once `MsgSettleRequest` is implemented so that we have defined the logic of approving a request

Pantani:
> [#219 (comment)](https://github.com/tendermint/spn/pull/219#issuecomment-891961176)
> 
> Let's talk here if we must continue this discussion.
> 
> Those information are only used before the chain is launched. Once the network is live, you no longer interact with the `launch` module. Information about validator in the `launch` module are only verified off-chain so you can potentially have wrong information stored that prevent you from starting the chain. The on-chain stored `GenesisValidator` must be removed in order to make the genesis fetched from `spn` correct and so this `ValidatorRemoval` should be sendable by anyone.
> 
> We must then have in the `MsgRequestRemoveValidator` and `MsgRequestRemoveAccount` containing two addresses:
> 
> * The address of the signer of the message
> * The address of the account to remove (that can be the same as the address above)
> 
> I approved the PR because it could may be simpler to implement this additional field on both requests on a single PR once both have been merged.
> 
> That's as you want. Can you please create an issue for adding this field if you don't implement it on this PR?

Great, I will implement both in the current PR. I think it's more quickly. The implementation is simple

x/launch/client/cli/tx_request_remove_validator.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+func CmdRequestRemoveValidator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "request-remove-validator [chainID] [validatorAddress]",
+		Short: "Request to remove a genesis validator",
+		Args:  cobra.MinimumNArgs(1),
```
lubtd, line 15:
```suggestion
		Args:  cobra.MinimumNArgs(2),
```

Pantani: https://github.com/tendermint/spn/pull/220/files#r683780516

ilgooz (dismissed):

x/launch/client/cli/tx_request_remove_validator.go
lubtd: I haven't seen it was about minimum arg👍

x/launch/client/cli/tx_request_remove_validator.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+func CmdRequestRemoveValidator() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "request-remove-validator [chainID] [validatorAddress]",
+		Short: "Request to remove a genesis validator",
+		Args:  cobra.RangeArgs(1, 2),
```
lubtd, line 15: I like this format👍 I'm wondering if we could have a way to precise `[validatorAddress]` is optional. Normally `[...]` is for optional argument and `<...>` for mandatory one but for some reason Cosmos SDK decided to use `[...]` for everything

lubtd (approved):

x/launch/client/cli/tx_request_remove_validator.go
Pantani: It will be cool if we have a way to differentiate both... maybe we can use `(...)`, but we are creating a new pattern. I wouldn't say I like it too much.

lubtd:
Hey, some conflict again🙃

We can merge afterward since it is approved

x/launch/client/cli/tx_request_remove_validator.go
lubtd: Let's keep it like this for now, this is not problematic if people put the address.

In fact, the CLI will mainly be used for testing purposes

PR #219, opened 2021-07-29
feat(profile): allow to delete a validator profile

### LINKED ISSUE #189: profile: allow to delete a validator profile

Pantani:
Closes #189

### What does this PR does?

- Create `delete-validator` structure;
- Implemente `DeleteValidator` method;
- Add `DeleteValidator` tests (`TestMsgDeleteValidator`);


x/profile/keeper/msg_server_delete_validator.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func (k msgServer) DeleteValidator(goCtx context.Context, msg *types.MsgDeleteValidator) (*types.MsgDeleteValidatorResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Check if the validator address is already in the store
+	_, found := k.GetValidator(ctx, msg.Address)
+	if !found {
+		return &types.MsgDeleteValidatorResponse{},
+			sdkerrors.Wrap(types.ErrValidatorNotFound,
```
lubtd, line 19: `sdkerrors` has a `Wrapf` method when the string is formated

x/profile/keeper/msg_server_delete_validator.go
```diff
@@ -0,0 +1,25 @@
...
+				fmt.Sprintf("validator: %s", msg.Address))
```
lubtd, line 20: I think just precising the address is sufficient based on the wrapped error

x/profile/types/message_delete_validator.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (msg *MsgDeleteValidator) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgDeleteValidator) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Address)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid address address (%s)", err)
```
lubtd, line 40:
```suggestion
		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid validator address (%s)", err)
```

x/profile/keeper/msg_server_delete_validator.go
```diff
@@ -0,0 +1,22 @@
+package keeper
+
+import (
+	"context"
```
lubtd, line 4:
```suggestion
	"context"
	
```

lubtd (commented):
Code looks legit to me👍

We will need https://github.com/tendermint/spn/pull/198 to be merged first in order to correctly test the feature from the CLI

lubtd:
We will actually have to change the specs a bit.

We should let the possibility to the coordinator to remove an account or a validator in case it has been mistakenly approved and some data are wrong (some checks must be performed off-chain for the validator).

We need to think about two things:
- Does anyone could send a removal request or either the account itself or the coordinator? In bother case, we will need to add a new field for the message

- Should coordinator request be automatically pre approved? This is something we had in the legacy version although this detail is not mentioned in the specs. We will first need to implement `MsgSettleRequest` for this so this doesn't influence this PR

Pantani:
> We will actually have to change the specs a bit.
> 
> We should let the possibility to the coordinator to remove an account or a validator in case it has been mistakenly approved and some data are wrong (some checks must be performed off-chain for the validator).
> 
> We need to think about two things:
> 
> * Does anyone could send a removal request or either the account itself or the coordinator? In bother case, we will need to add a new field for the message
> * Should coordinator request be automatically pre approved? This is something we had in the legacy version although this detail is not mentioned in the specs. We will first need to implement `MsgSettleRequest` for this so this doesn't influence this PR

If I understand the concept right, the coordinator sends the removal can be better for blockchain itself. He can drop peers with too much downtime or malicious. I think in most cases, the validators just shut down your nodes instead send a removal message before the shutdown.

IMHO, let's keep the automatically pre-approved by the coordinator, and we can add it to the specs

lubtd:
Seems like I put my previous comment above on the wrong PR. It should be here: https://github.com/tendermint/spn/pull/216

lubtd (approved):

ilgooz (approved):

PR #221, opened 2021-08-03
feat(launch): request a genesis account

### LINKED ISSUE #158: launch: request a genesis account
Create the message to send an add account request into the request pool for a chain

Pantani:
Closes #158

### What does this PR does?

Scaffold `request-add-validator` msg;
Implements `RequestAddValidator` method and tests;


x/launch/types/message_request_add_account.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return []sdk.AccAddress{address}
+}
+
+func (msg *MsgRequestAddAccount) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgRequestAddAccount) ValidateBasic() error {
```
lubtd, line 39: We can add that the coins must not be empty

x/launch/types/message_request_add_account_test.go
```diff
@@ -0,0 +1,58 @@
...
+				"invalid address (invalid_address): decoding bech32 failed: invalid index of 1"),
+		}, {
+			name: "invalid chain id",
+			msg: types.MsgRequestAddAccount{
+				Address: sample.AccAddress(),
+				ChainID: "invalid_chain",
+			},
+			err: sdkerrors.Wrapf(types.ErrInvalidChainID, "invalid_chain"),
+		}, {
+			name: "invalid chain name",
```
lubtd, line 34: We don't have chain name for this message, what's the difference with the test above?

x/launch/types/message_request_add_account_test.go
```diff
@@ -0,0 +1,58 @@
...
+		name string
+		msg  types.MsgRequestAddAccount
+		err  error
+	}{
+		{
+			name: "invalid address",
+			msg: types.MsgRequestAddAccount{
+				Address: "invalid_address",
+			},
+			err: sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress,
```
lubtd, line 24:
```suggestion
			err: sdkerrors.Wrap(sdkerrors.ErrInvalidAddress,
```

Same below

x/launch/types/request_content_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"testing"
+
+	codec "github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+func TestValidatorRemovalCodec(t *testing.T) {
```
lubtd, line 12:
```suggestion
func TestGenesisAccountCodec(t *testing.T) {
```

x/launch/types/message_request_add_account_test.go
Pantani: the chainID format is right, but the name is wrong because we have an invalid character (`"wh.thc-10"`)

lubtd (commented):
Works well👍 Just some little details for the tests

x/launch/types/message_request_add_account.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid address (%s)", err)
+	}
+
+	_, _, err = ParseChainID(msg.ChainID)
+	if err != nil {
+		return sdkerrors.Wrap(ErrInvalidChainID, msg.ChainID)
+	}
+
+	if len(msg.Coins) == 0 {
```
lubtd, line 50:
```suggestion
	if msg.Coins.Empty() {
```

x/launch/types/message_request_add_account_test.go
```diff
@@ -0,0 +1,64 @@
...
+				ChainID: "invalid_chain",
+				Coins:   sample.Coins(),
+			},
+			err: sdkerrors.Wrap(types.ErrInvalidChainID, "invalid_chain"),
+		}, {
+			name: "message without coins",
+			msg: types.MsgRequestAddAccount{
+				Address: addr,
+				ChainID: chainID,
+				Coins:   nil,
```
lubtd, line 42:
```suggestion
				Coins:   sdk.NewCoins(),
```

Not 100% if it's the name of the method

x/launch/keeper/msg_server_request_add_account_test.go
```diff
@@ -0,0 +1,101 @@
...
+	k.SetChain(sdkCtx, chains[3])
+	tests := []struct {
+		name string
+		msg  types.MsgRequestAddAccount
+		want uint64
+		err  error
+	}{
+		{
+			name: "invalid chain",
+			msg: types.MsgRequestAddAccount{
```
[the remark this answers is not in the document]
lubtd: Can we use sample.Coins for each

lubtd (dismissed):

x/launch/types/message_request_add_account.go
```diff
@@ -0,0 +1,54 @@
...
+
+	_, _, err = ParseChainID(msg.ChainID)
+	if err != nil {
+		return sdkerrors.Wrap(ErrInvalidChainID, msg.ChainID)
+	}
+
+	if msg.Coins.Empty() {
+		return sdkerrors.Wrap(ErrEmptyCoins, msg.Address)
+	}
+	return nil
```
lubtd, line 53: One additional thing.

We have a method `IsValid` from the type `Coins` that performs various checks like duplicated denom

We should call this method here

Pantani: I added the `IsValid`, but this is not checking if the coin list is empty, so I added both

```go
	if !msg.Coins.IsValid() || msg.Coins.Empty() {
		return sdkerrors.Wrap(ErrInvalidCoins, msg.Address)
	}
```

lubtd: Yes, I was meaning having both sorry if it wasn't clear

x/launch/types/message_request_add_account_test.go
```diff
@@ -0,0 +1,65 @@
...
+		}, {
+			name: "message without coins",
+			msg: types.MsgRequestAddAccount{
+				Address: addr,
+				ChainID: chainID,
+				Coins:   sdk.NewCoins(),
+			},
+			err: sdkerrors.Wrap(types.ErrInvalidCoins, addr),
+		}, {
+			name: "valid message",
```
lubtd, line 47: We can have a test as well for invalid coin by using `Coins: sdk.NewCoins(sdk.Coin{
  Amount: sdk.NewInt(10),
  Denom: "",
}),`

Pantani: Invalid coin venom causes panic to the cosmos SDK library. we must change the test to recover and assert the panic message only for this case:

Screen Shot 2021-08-09 at 13 53 57

Pantani: the `sdk.NewCoins` try to validate and causes this panic, I will create the tests in this way:

`sdk.Coins{sdk.Coin{Denom: "", Amount: sdk.NewInt(10)}}`

lubtd: You mean `sdk.NewCoins` makes this panic?

We can define the value manually then since `Coins` is an alias of `[]Coin`: `[]sdk.Coin{
{...}
}`

Pantani: yes. `sdk.NewCoins` cause panics
I created the tests directly with the object (`sdk.Coins{}`)  instead call the `sdk.NewCoins`

Pantani: also, should we be careful to use `sdk.NewCoins` to not crash the application if the user passes an invalid coin

lubtd: If it is called in `cli` it doesn't matter. We must avoid calling it in `keeper` and `types` but I doubt we call it there

lubtd (dismissed):
Great work🦾

lubtd (approved):

PR #222, opened 2021-08-03
feat(launch): request a vested account

### LINKED ISSUE #159: launch: request a vested account
Create the message to send an add vested account request into the request pool for a chain

Pantani:
Closes #159

### What does this PR does?

Scaffold `request-add-vested-account` msg;
Implements `RequestAddVestedAccount` method and tests;


x/launch/client/cli/tx_request_add_vested_account.go
```diff
@@ -0,0 +1,56 @@
...
+			if err != nil {
+				return err
+			}
+
+			coins, err := sdk.ParseCoinsNormalized(args[1])
+			if err != nil {
+				return fmt.Errorf("failed to parse coins: %w", err)
+			}
+
+			delayedVesting, err := codec.NewAnyWithValue(&types.DelayedVesting{
```
lubtd, line 32: For `DelayedVesting` you have a value `Vesting` which are coins that will be vested at the time `EndTime`

Since for now we only support `DelayedVesting, can we have two additional positional argument:
- `vesting-coins`
- `vesting-end-time`

x/launch/keeper/msg_server_request_add_vested_account_test.go
```diff
@@ -0,0 +1,93 @@
...
+	tests := []struct {
+		name string
+		msg  types.MsgRequestAddVestedAccount
+		want uint64
+		err  error
+	}{
+		{
+			name: "invalid chain",
+			msg: types.MsgRequestAddVestedAccount{
+				ChainID: invalidChain,
```
lubtd, line 32: Same feedback as for `MsgRequestRemoveAccount`

Let's define all the fields for each message test

x/launch/types/message_request_add_vested_account.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid address (%s)", err)
+	}
+
+	_, _, err = ParseChainID(msg.ChainID)
+	if err != nil {
+		return sdkerrors.Wrap(ErrInvalidChainID, msg.ChainID)
+	}
+
+	if len(msg.Coins) == 0 {
```
lubtd, line 53: Can we use the `Empty()` method defined for this type

x/launch/types/message_request_add_vested_account.go
```diff
@@ -0,0 +1,57 @@
...
+		return sdkerrors.Wrap(ErrEmptyCoins, msg.Address)
+	}
```
lubtd, line 55: We should also check that the provided `VestingOption` is one of the possible value for the any type (which now can only be `DelayedVesting`)

It is implemented for `InitialGenesis` for example: https://github.com/tendermint/spn/blob/c815fafa7096e7054027661e5677f659fe5e8298/x/launch/types/message_edit_chain.go#L67

I'm also thinking that we can check if the provided value is valid. For this we can define a method `validate() error` for the `VestingOption` interface and we call it here.
So we just verify that Coins is non empty and timestamp is not 0

I create an issue since we can do the same for `InitialGenesis`

x/launch/types/message_request_add_vested_account_test.go
```diff
@@ -0,0 +1,60 @@
...
+		}, {
+			name: "message without coins",
+			msg: types.MsgRequestAddVestedAccount{
+				Address: addr,
+				ChainID: chainID,
+			},
+			err: sdkerrors.Wrap(types.ErrEmptyCoins, addr),
+		}, {
+			name: "valid message",
+			msg: types.MsgRequestAddVestedAccount{
```
lubtd, line 43: This acutally shouldn't be valid if we don't provide a content

x/launch/client/cli/tx_request_add_vested_account.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	codec "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+func CmdRequestAddVestedAccount() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "request-add-vested-account [chainID] [coins] [options]",
```
lubtd, line 18: What is `options` since it is not used

x/launch/client/cli/tx_request_add_vested_account.go
```diff
@@ -0,0 +1,60 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdRequestAddVestedAccount() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "request-add-vested-account [chainID] [vestingCoins] [vestingEndTime]",
```
lubtd, line 20: We must have two coin arguments:
- `initial-coins` (something like this) that is the initial amount of coins directly accessible to the account like a usual genesis account 
- `vesting-coins` the amount of coins vested at the vesting end time

lubtd: We can use `starting-balance` for the initial coins since this is the name of the field

x/launch/types/message_request_add_vested_account.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid address (%s)", err)
+	}
+
+	_, _, err = ParseChainID(msg.ChainID)
+	if err != nil {
+		return sdkerrors.Wrap(ErrInvalidChainID, msg.ChainID)
+	}
+
+	if msg.Coins.Empty() {
```
lubtd, line 53: We can remove this check because we can actually have an empty starting balance: the account only has vesting coins.

The opposite is not true though because is there is no vesting coins then the user should use a regular genesis account

lubtd (changes requested):

x/launch/types/message_request_add_vested_account.go
Pantani: great, thks!
should I remove this one too?

https://github.com/tendermint/spn/pull/222/files#diff-e3eb87deb7e77fead8c2e3e18dd83ce6edc6dc8d7071257c269dffc2a3d09b58R16-R18

lubtd: No we must have vesting coins otherwise there is no point using a vestedAccount instead of genesisAccount

lubtd:
Same comment as here: https://github.com/tendermint/spn/pull/221#discussion_r685122073

For starting balance (if not empty) and vesting coins

x/launch/types/vesting_options.go
```diff
@@ -1,4 +1,23 @@
...
+}
+
+var _ VestingOptions = &DelayedVesting{}
+
+// Validate check the DelayedVesting object
+func (g DelayedVesting) Validate() error {
+	if g.Vesting.Empty() || !g.Vesting.IsValid() {
+		return sdkerrors.Wrap(ErrInvalidCoins, "invalid vesting coins for DelayedVesting")
+	}
+	if g.EndTime == 0 {
```
lubtd, line 19: Although we can have a empty `StartingBalance` we should still verify the the value provided is valid.

So we can check `IsValid()` without `Empty()`

lubtd (commented):
Tested and it works well👍

Just one last comment

lubtd (dismissed):

ilgooz (approved):

Issue #161, opened 2021-07-16
launch: request a genesis validator


lubtd:
Create the message to send an add validator request into the request pool for a chain

PR #224, opened 2021-08-03
feat(launch): filter accounts and validator list by chain ID flag

### LINKED ISSUE #178: launch: queries to list genesis accounts, vested accounts and validators from a chain ID
The genesis account, vested account and validator structures we scaffold with Starport will only include queries all accouts,.. independently from the chainID or a specific one

We must create queries to be able to query those value from a chain ID (eg. all genesis account for a chain) with the `starport s query` command.

I create a specific issue for this since handling this in a specific PR will make code review easier

Pantani:
Closes #178

### What does this PR does?

- Create a `chain-id` argument to filter `all` requests for genesis and vested accounts, and genesis validators
- remove the unused casting from string to string


lubtd:
I'm personally advocating for having `chainID` as a positional argument and removing the ability to query all requests from all chains. I don't think there is a practical use case of this.

x/launch/keeper/grpc_query_genesis_account.go
```diff
@@ -28,7 +28,9 @@ func (k Keeper) GenesisAccountAll(c context.Context, req *types.QueryAllGenesisA
 			return err
 		}
 
-		genesisAccounts = append(genesisAccounts, &genesisAccount)
+		if req.ChainID == "" || genesisAccount.ChainID == req.ChainID {
```
lubtd, line 31: For performance reason we should use `query.Paginate` with the longest common prefix. So in this case the prefix that include the `chainID`

For this we can add a method that returns the prefix with `chainID` in `types/key_[accounttype].go` then we provide this prefix to the `prefix.NewStore` method above

lubtd (changes requested):

proto/launch/query.proto
```diff
@@ -112,12 +112,14 @@ message QueryGetVestedAccountResponse {
...
 message QueryAllVestedAccountRequest {
-	cosmos.base.query.v1beta1.PageRequest pagination = 1;
+	string chainID = 1;
+	cosmos.base.query.v1beta1.PageRequest pagination = 2;
 }
 
 message QueryAllVestedAccountResponse {
-	repeated VestedAccount vestedAccount = 1;
-	cosmos.base.query.v1beta1.PageResponse pagination = 2;
+	string chainID = 1;
```
lubtd, line 120: We don't need the `chainID` in the response

x/launch/client/cli/query_genesis_account_test.go
```diff
@@ -31,7 +31,7 @@ func networkWithGenesisAccountObjects(t *testing.T, n int) (*network.Network, []
 	for i := 0; i < n; i++ {
 		state.GenesisAccountList = append(
 			state.GenesisAccountList,
-			sample.GenesisAccount(strconv.Itoa(i), strconv.Itoa(i)),
+			sample.GenesisAccount("foo", strconv.Itoa(i)),
```
lubtd, line 34: We should test with a valid `chainID` from `sample`

x/launch/client/cli/query_vested_account.go
```diff
@@ -12,8 +12,8 @@ import (
 func CmdListVestedAccount() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "list-vested-account",
-		Short: "list all vestedAccount",
-		Args:  cobra.NoArgs,
+		Short: "list all vestedAccount [chainID]",
```
lubtd, line 15:
```suggestion
		Short: "list all vestedAccount",
```

x/launch/client/cli/query_vested_account.go
```diff
@@ -12,8 +12,8 @@ import (
 func CmdListVestedAccount() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "list-vested-account",
```
lubtd, line 14:
```suggestion
		Use:   "list-vested-account [chainID]",
```

lubtd (commented):
Just a single fix and that's it

lubtd:
The attached issue https://github.com/tendermint/spn/issues/178 also include list validators by `chainID` so we will have to handle this to close the issue. This can be completed in a separate PR

Pantani:
> The attached issue #178 also include list validators by `chainID` so we will have to handle this to close the issue. This can be completed in a separate PR

we can do it in the same PR we will create the request to add the validator

lubtd:
> we can do it in the same PR we will create the request to add the validator

Mmh so do you plan to do it on this PR?

Because you requested for a review but the validator query still doesn't include `chainID`

Pantani:
> > we can do it in the same PR we will create the request to add the validator
> 
> Mmh so do you plan to do it on this PR?
> 
> Because you requested for a review but the validator query still doesn't include `chainID`

Done!! 
I made a mistake. I was thought we need to implement #161 before it. Forget my last comment

lubtd:
> > > we can do it in the same PR we will create the request to add the validator
> > 
> > 
> > Mmh so do you plan to do it on this PR?
> > Because you requested for a review but the validator query still doesn't include `chainID`
> 
> Done!!
> I made a mistake. I was thought we need to implement #161 before it. Forget my last comment

Nope

Well we will need `MsgSettleRequest` in order to test the queries from the CLI but this is not a very big concern as long as we have all the tests

lubtd (approved):
👍

ilgooz (approved):

PR #229, opened 2021-08-05
feat: add `Validate` method to `InitialGenesis`

lubtd:
Add a method for validation in the `InitialGenesis` interface to have a stateless check on the values (non empty genesis URL for example)
Add additional tests for this


testutil/sample/launch.go
```diff
@@ -136,3 +138,9 @@ func MsgEditChain(
 		initialGenesis,
 	)
 }
+
+// GenesisHash returns a sample sha256 hash of custom genesis for GenesisURL
+func GenesisHash() string {
```
Pantani, line 143: If we merge the #230 first, we can reuse this function:

https://github.com/tendermint/spn/pull/230/files#diff-6619c7d1a9a468c88892c0e3dcd2961584eae593aae45d7071bc902b9eb59896R13-R18

x/launch/types/initial_genesis.go
```diff
@@ -1,4 +1,28 @@
...
+// Validate implements InitialGenesis
+func (g GenesisURL) Validate() error {
+	if g.Url == "" {
+		return errors.New("no url provided")
+	}
+	if len(g.Hash) != HashLength {
+		return errors.New("hash must be sha256")
+	}
+	return nil
+}
```
Pantani, line 28: Should we create separate tests for it? What do you think?

Pantani (approved):
🙌🏽
Only a few suggestions

x/launch/types/initial_genesis.go
lubtd: Yep good point. We should maybe think about having a test coverage CI step

Pantani: we can use the [codecov](https://about.codecov.io/language/go/)

ilgooz (dismissed):

Pantani (approved):
👏🏼

PR #230, opened 2021-08-05
feat(launch): automatically compute custom genesis hash in CLI

lubtd:
Define the hash function for initial genesis as SHA256 hash

For `MsgCreateChain` and `MsgEditChain`, when a genesis URL for a custom initial genesis is provided, we automatically compute the corresponding hash


x/launch/client/cli/tx_create_chain.go
```diff
@@ -52,8 +56,29 @@ func CmdCreateChain() *cobra.Command {
...
 
 	cmd.Flags().String(flagGenesisURL, "", "URL for a custom genesis")
-	cmd.Flags().String(flagGenesisHash, "", "hash of the content of the custom genesis")
 	flags.AddTxFlagsToCmd(cmd)
 
 	return cmd
 }
+
+// getHashFromURL fetches content from url and returns the hash based on the genesis hash method
+func getHashFromURL(ctx context.Context, url string) (string, error) {
```
Pantani, line 65: Maybe we should move it to `types/initial_genesis.go` and keep here only the logic for CLI. What do you think?

Pantani (approved):
👌🏽

x/launch/client/cli/tx_create_chain.go
lubtd: This is just a small helper for the CLI
Also I would prefer keeping only stateless and "pure" methods in `types`

lubtd: I said we will only use it for the CLI because we can't in the blockchain logic verify the content of the url. The hash is always verified off-chain

ilgooz (approved):

Pantani (approved):

PR #233, opened 2021-08-06
feat(CI): test coverage

### LINKED ISSUE #231: Add a code test coverage CI action
Tests are important for blockchains, so is code coverage

Pantani:
closes #231


lubtd:
Thanks for already have looked into this👍

Also I personally think that we shouldn't set a particular threshold for the coverage or try a specific value.
To me it would be more like a support when we implement a new feature to have a good sight of the scope of our tests

Pantani:
> Thanks for already have looked into this👍
> 
> Also I personally think that we shouldn't set a particular threshold for the coverage or try a specific value.
> To me it would be more like a support when we implement a new feature to have a good sight of the scope of our tests

sure, we can define or remove the thresholds, I added a codecov config file

https://github.com/tendermint/spn/blob/feat/code-coverage/.codecov.yml

lubtd (approved):
This is awesome, all the config seems legit to me👍

Do we need the report on `develop` first to be able to access the full report?
Is there a way we can access the precise lines that are covered?

ilgooz (approved):

Pantani:
> This is awesome, all the config seems legit to me👍
> 
> Do we need the report on `develop` first to be able to access the full report?
> Is there a way we can access the precise lines that are covered?

I'm not sure, but I think so

https://docs.codecov.com/docs/error-reference

lubtd (approved):

PR #235, opened 2021-08-07
feat(launch): allow to settle requests

### LINKED ISSUE #165: launch: allow to settle requests
Give the ability to coordinators to settle a request in the request pool:

- Approve a request
- Reject a request

When the request is rejected, it is simply removed from the request pool.
If it is approved, the changes must be done for the chain launch to include this approved request like account or validator

Some on-check verification must be done:
- An account can't be approved if an account of the same address already exist: it must be removed with a remove account request before
- We can't have a genesis account and a vested account with the same address

When approving a validator, we don't check if an account of the address exist because the account can be set through a custom genesis. This check must be performed off-chain

Pantani:
Closes #165

### What does this PR does?

- Create `MsgSettleRequest`;
- Implements `SettleRequest` keeper method and test;
- Implement validation;
- Create `ApplyRequest` method;


x/launch/client/cli/tx_settle_request.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdSettleRequest() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "settle-request [chainID] [requestID] [approve]",
```
lubtd, line 19: I think it can be dangerous to use a boolean value for `approve` because I may think `ParseBool` can give the wrong value from a simple typing mistake

What I would do is to have, as the first positional argument, a string that can be either `approve` or `reject`
```
settle-request [approve|reject] [chainID] [requestID]
```

Another solution can be to have two commands `approve-request` and `reject-request` but I think the solution above is sufficient

x/launch/client/cli/tx_settle_request.go
```diff
@@ -0,0 +1,108 @@
...
+	"strings"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+var _ = strconv.Itoa(0)
```
lubtd, line 15:
```suggestion
```

Some mini-hack from templates we should figure out how to remove 😅

x/launch/client/cli/tx_settle_request.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	}
+
+	flags.AddTxFlagsToCmd(cmd)
+
+	return cmd
+}
+
+// parseList parses comma separated numbers and range to []uint64.
+func parseList(arg string) ([]uint64, error) {
```
lubtd, line 58: We can save this function aside for `starport` CLI if it manages ranges

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,92 @@
...
+	types.RegisterInterfaces(cdc)
+
+	var content types.RequestContent
+	if err := cdc.UnpackAny(request.Content, &content); err != nil {
+		return sdkerrors.Wrap(types.ErrInvalidRequestContent, err.Error())
+	}
+
+	switch c := content.(type) {
+	case *types.GenesisAccount:
+		k.SetGenesisAccount(ctx, *c)
```
lubtd, line 77: Must fails if the account already set as GenesisAccount or VestedAccount

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,92 @@
...
+	case *types.VestedAccount:
+		k.SetVestedAccount(ctx, *c)
```
lubtd, line 79: Must fails if the account already set as GenesisAccount or VestedAccount

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,92 @@
...
+	case *types.AccountRemoval:
+		k.RemoveVestedAccount(ctx, msg.ChainID, c.Address)
```
lubtd, line 81: Must fails if the account is not set neither as GenesisAccount nor as VestedAccount

Must throw a critical error if set as both

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,92 @@
...
+		k.RemoveGenesisAccount(ctx, msg.ChainID, c.Address)
+	case *types.GenesisValidator:
+		k.SetGenesisValidator(ctx, *c)
```
lubtd, line 84: Must fails if already set

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,92 @@
...
+	case *types.ValidatorRemoval:
+		k.RemoveGenesisValidator(ctx, msg.ChainID, c.ValAddress)
```
lubtd, line 86: Must fails if not set

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,92 @@
...
+	cdc := codectypes.NewInterfaceRegistry()
+	types.RegisterInterfaces(cdc)
+
+	var content types.RequestContent
+	if err := cdc.UnpackAny(request.Content, &content); err != nil {
+		return sdkerrors.Wrap(types.ErrInvalidRequestContent, err.Error())
+	}
+
+	switch c := content.(type) {
+	case *types.GenesisAccount:
```
lubtd, line 76: Instead of using a switch, I think we can add a `Apply(context) error` method to the `RequestContent` interface

lubtd (commented):
Two main feedbacks:

- The message must take only one request ID as argument to make the message as minimal as possible
- We must add checks for request applications. We should avoid the coordinator seeing an approved request while it had no consequence on the state of the genesis

The fact that Add... requests don't fail but update the value if it is already present can be discussed. Although I think it would be more explicit to be required to send a Remove request first then send another Add when an account must be updated

testutil/sample/launch.go
```diff
@@ -87,6 +97,36 @@ func Request(chainID string) *launch.Request {
 	}
 }
 
+func CreateAllRequestContents(chainID, genesis, vested, validator string) []*types.Any {
```
lubtd, line 100: Can we document the method

lubtd:
```suggestion
func AllRequestContents(chainID, genesis, vested, validator string) []*types.Any {
```

x/launch/client/cli/tx_settle_request.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+func CmdSettleRequest() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "settle-request [approve/reject] [chainID] [requestID]",
```
lubtd, line 15:
```suggestion
		Use:   "settle-request [approve|reject] [chainID] [requestID]",
```

I would prefer using `|` that means or

x/launch/client/cli/tx_settle_request.go
```diff
@@ -0,0 +1,50 @@
...
+		Short: "Approve or reject a pending request",
+		Args:  cobra.ExactArgs(3),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			approve, err := strconv.ParseBool(args[0])
```
lubtd, line 19: I meant in my previous comment that I think it can be more explicit to either provide `"approve"` or `"reject"` string as the first argument instead of using `ParseBool`  and returning an error if none is provided

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+	chain, found := k.GetChain(ctx, msg.ChainID)
+	if !found {
+		return nil, sdkerrors.Wrap(types.ErrChainNotFound, msg.ChainID)
+	}
+
+	if chain.LaunchTriggered {
+		return nil, sdkerrors.Wrap(types.ErrTriggeredLaunch, msg.ChainID)
+	}
+
+	coordAddress := k.profileKeeper.GetCoordinatorAddressFromID(ctx, chain.CoordinatorID)
```
lubtd, line 29: Can we handle the case when the coordinator address is not found

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+	if msg.Coordinator != coordAddress {
+		return nil, sdkerrors.Wrap(types.ErrNoAddressPermission, msg.Coordinator)
+	}
+
+	// first check if the request exist
```
lubtd, line 34:
```suggestion
	// first check if the request exists
```

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := cdc.UnpackAny(request.Content, &content); err != nil {
+		return spnerrors.Critical(err.Error())
+	}
+
+	switch c := content.(type) {
+	case *types.GenesisAccount:
+		_, foundGenesis := k.GetGenesisAccount(ctx, msg.ChainID, c.Address)
+		_, foundVested := k.GetVestedAccount(ctx, msg.ChainID, c.Address)
+		if foundVested || foundGenesis {
+			return spnerrors.Critical(fmt.Sprintf(
```
lubtd, line 77: This is a normal error if one of them is set

This is a critical error if both are set `foundVested && foundGenesis` because it must never be the case

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+				"account %s for chain %s already exist",
+				c.Address, msg.ChainID),
+			)
+		}
+		k.SetGenesisAccount(ctx, *c)
+	case *types.VestedAccount:
+		_, foundGenesis := k.GetGenesisAccount(ctx, msg.ChainID, c.Address)
+		_, foundVested := k.GetVestedAccount(ctx, msg.ChainID, c.Address)
+		if foundVested || foundGenesis {
+			return spnerrors.Critical(fmt.Sprintf(
```
lubtd, line 87: Same as above

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+		_, foundGenesis := k.GetGenesisAccount(ctx, msg.ChainID, c.Address)
+		_, foundVested := k.GetVestedAccount(ctx, msg.ChainID, c.Address)
+		if foundGenesis {
+			k.RemoveGenesisAccount(ctx, msg.ChainID, c.Address)
+		}
+		if foundVested {
+			k.RemoveVestedAccount(ctx, msg.ChainID, c.Address)
+		}
+		if !foundGenesis && !foundVested {
+			return spnerrors.Critical(fmt.Sprintf(
```
lubtd, line 103: This is a normal error

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+				"account %s for chain %s not found",
+				c.Address, msg.ChainID),
+			)
+		}
+	case *types.GenesisValidator:
+		if _, found := k.GetGenesisValidator(ctx, msg.ChainID, c.Address); found {
+			return spnerrors.Critical(fmt.Sprintf(
```
lubtd, line 110: Normal error

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+				"genesis validator %s for chain %s already exist",
+				c.Address, msg.ChainID),
+			)
+		}
+		k.SetGenesisValidator(ctx, *c)
+	case *types.ValidatorRemoval:
+		if _, found := k.GetGenesisValidator(ctx, msg.ChainID, c.ValAddress); !found {
+			return spnerrors.Critical(fmt.Sprintf(
```
lubtd, line 118: Normal error

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+			)
+		}
+		k.SetVestedAccount(ctx, *c)
+	case *types.AccountRemoval:
+		_, foundGenesis := k.GetGenesisAccount(ctx, msg.ChainID, c.Address)
+		_, foundVested := k.GetVestedAccount(ctx, msg.ChainID, c.Address)
+		if foundGenesis {
+			k.RemoveGenesisAccount(ctx, msg.ChainID, c.Address)
+		}
+		if foundVested {
```
lubtd, line 99: Same as for the Add request. If both are set at the same time we must throw a critical error because of store inconsistency

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,268 @@
...
+		name string
+		msg  types.MsgSettleRequest
+		want uint64
+		err  error
+	}{
+		{
+			name: "invalid chain",
+			msg: types.MsgSettleRequest{
+				ChainID:     invalidChain,
+				Coordinator: addr1,
```
lubtd, line 39: Can we use `pk.GetCoordinatorAddressFromID(sdkCtx, chains[1].CoordinatorID)` as well if the test is not about invalid coordinator

x/launch/keeper/msg_server_settle_request.go
```diff
@@ -0,0 +1,128 @@
...
+		err := applyRequest(ctx, k.Keeper, msg, request)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return &types.MsgSettleRequestResponse{}, nil
+}
+
+// applyRequest approve the request and perform
```
lubtd, line 56:
```suggestion
// applyRequest approves the request and performs
```

lubtd: Can we move this method to `request.go` since it will not only be called by this specific message

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+			require.NoError(t, err)
+
+			_, found := k.GetRequest(sdkCtx, tt.msg.ChainID, tt.msg.RequestID)
+			require.False(t, found, "request not removed")
+		})
+	}
+}
+
+func TestApplyRequest(t *testing.T) {
```
lubtd, line 117: Same as for `applyRequest`, we can move this test into `request_test.go`

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,268 @@
...
+		requests                 = createNRequest(k, sdkCtx, 6)
+	)
+	chains[0].LaunchTriggered = true
+	k.SetChain(sdkCtx, chains[0])
+	chains[1].ChainID = "foo"
+	k.SetChain(sdkCtx, chains[1])
+	tests := []struct {
+		name string
+		msg  types.MsgSettleRequest
+		want uint64
```
lubtd, line 32:
```suggestion
```

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,268 @@
...
+		err  error
+	}{
+		{
```
lubtd, line 35: We should have one case where the `applyRequest` fails to ensure the message fails if the request application fails.

Just one case since all the cases are tested in `TestApplyRequest`

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,268 @@
...
+		k, _, _, _, sdkCtx, _ = setupMsgServer(t)
+		chainID, _            = sample.ChainID(10)
+		contents              = sample.CreateAllRequestContents(chainID, genesis, vested, validator)
+		requests              = createRequests(k, sdkCtx, chainID, contents)
+		invalidContent, _     = codec.NewAnyWithValue(&types.Request{})
+		invalidContentID      = k.AppendRequest(sdkCtx, *sample.RequestWithContent(chainID, invalidContent))
+	)
+	tests := []struct {
+		name string
+		msg  types.MsgSettleRequest
```
lubtd, line 131: Can we have more generic tests for `applyRequest` where `types.Request` is directly given in the test case

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,268 @@
...
+			err: spnerrors.Critical(
+				"no concrete type registered for type URL /tendermint.spn.launch.Request against interface *types.RequestContent"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			request, found := k.GetRequest(sdkCtx, tt.msg.ChainID, tt.msg.RequestID)
+			require.True(t, found)
+
+			err := applyRequest(sdkCtx, *k, &tt.msg, request)
```
lubtd, line 259: We should assert the request is correctly applied into the store when it is successful (account exists if add account, account is removed if no account, etc.)

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,268 @@
...
+			_, err := srv.SettleRequest(ctx, &tt.msg)
+			if tt.err != nil {
+				require.ErrorIs(t, err, tt.err)
+				return
+			}
+			require.NoError(t, err)
+
+			_, found := k.GetRequest(sdkCtx, tt.msg.ChainID, tt.msg.RequestID)
+			require.False(t, found, "request not removed")
+		})
```
lubtd, line 113: We should assert the request is applied when approve is true.

Since request application is already tested in `TestApplyRequest`. We can consider all request in `TestMsgSettleRequest` as `AddAccount` request and therefore simply verify the account is set in the store if approve is `true` and not if approve is `false`

lubtd (changes requested):

x/launch/keeper/msg_server_settle_request_test.go
Pantani: We don't have the chain list object for this method. We don't need them. The chain validations regarding coordinator address are in the parent method. This one is only for apply the request. I created an object to represent the coordinator account in the tests to be easier

x/launch/keeper/msg_server_settle_request_test.go
Pantani: we already have all failures cases for the `applyRequest` method

and also all cases for msg failures:
https://github.com/tendermint/spn/pull/235/files#diff-ab80113e0f271d47bc52c73c901a3c0bafdb4d2ef481d8f5ca63a2a011e1c152R13

lubtd: Yes all the failure cases for request application should be in `TestApplyRequest`

We should still assert that this specific message fails if `applyRequest` fails

x/launch/keeper/request.go
```diff
@@ -112,3 +116,81 @@ func (k Keeper) GetAllRequest(ctx sdk.Context) (list []types.Request) {
...
+	var content types.RequestContent
+	if err := cdc.UnpackAny(request.Content, &content); err != nil {
+		return spnerrors.Critical(err.Error())
+	}
+
+	switch c := content.(type) {
+	case *types.GenesisAccount:
+		_, foundGenesis := k.GetGenesisAccount(ctx, chainID, c.Address)
+		_, foundVested := k.GetVestedAccount(ctx, chainID, c.Address)
+		if foundVested || foundGenesis {
```
lubtd, line 140: We should also return a critical error when `foundGenesis && foundVested`  the same as `AccountRemoval` and also `VestedAccount`

To factorize things, we could implement a local method `CheckAccountExistence(...) (bool, err)` that returns `false` is neither GenesisAccount nor VestedAccount exists, `true` if one of them exists and return the critical error if both exist

x/launch/keeper/request.go
```diff
@@ -112,3 +116,81 @@ func (k Keeper) GetAllRequest(ctx sdk.Context) (list []types.Request) {
...
+			return spnerrors.Critical(
+				fmt.Sprintf("account %s for chain %s found in vested and genesis accounts",
+					c.Address, chainID),
+			)
+		}
+		if foundGenesis {
+			k.RemoveGenesisAccount(ctx, chainID, c.Address)
+		} else if foundVested {
+			k.RemoveVestedAccount(ctx, chainID, c.Address)
+		}
```
lubtd, line 175:
```suggestion
		k.RemoveGenesisAccount(ctx, chainID, c.Address)
		k.RemoveVestedAccount(ctx, chainID, c.Address)
```

x/launch/keeper/request_test.go
```diff
@@ -76,3 +93,111 @@ func TestRequestCount(t *testing.T) {
...
+			case *types.GenesisAccount:
+				_, found := k.GetGenesisAccount(sdkCtx, chainID, c.Address)
+				require.True(t, found, "genesis account not found")
+			case *types.VestedAccount:
+				_, found := k.GetVestedAccount(sdkCtx, chainID, c.Address)
+				require.True(t, found, "vested account not found")
+			case *types.AccountRemoval:
+				_, foundGenesis := k.GetGenesisAccount(sdkCtx, chainID, c.Address)
+				_, foundVested := k.GetVestedAccount(sdkCtx, chainID, c.Address)
+				require.False(t, foundGenesis && foundVested, "account not removed")
```
lubtd, line 193:
```suggestion
				require.False(t, foundGenesis || foundVested, "account not removed")
```

Should be removed from both places. Here the assertion would succeed is one exists

x/launch/types/expected_keepers.go
```diff
@@ -5,6 +5,7 @@ import (
 )
 
 type ProfileKeeper interface {
+	HasCoordinator(ctx sdk.Context, id uint64) bool
```
lubtd, line 8: I would instead returns `(address string, found bool)` in `GetCoordinatorAddressFromID` to make both methods consistent

And btw I think we can rename `GetCoordinatorAddressFromID` into `CoordinatorAddressFromID` for consistency as well

lubtd:
Tested and it works very well. That's very exciting to see it's taking form🚀

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -0,0 +1,140 @@
...
+				require.ErrorIs(t, err, tt.err)
+				return
+			}
+			require.NoError(t, err)
+
+			_, found := k.GetRequest(sdkCtx, tt.msg.ChainID, tt.msg.RequestID)
+			require.False(t, found, "request not removed")
+
+			_, found = k.GetGenesisAccount(sdkCtx, tt.msg.ChainID, tt.checkAddr)
+			require.True(t, found, "request apply not found")
```
lubtd, line 137: We should assert it exists when `msg.Approve == true` and it doesn't when `msg.Approve == false`

x/launch/keeper/msg_server_settle_request_test.go
Pantani: done on `b0ddecc`

testutil/sample/launch.go
```diff
@@ -87,6 +99,45 @@ func Request(chainID string) *launch.Request {
 	}
 }
 
+// AllRequestContents creates all contents types for request and
+// returns a list of all pack contents converted to `types.Any` object
+func AllRequestContents(chainID, genesis, vested, validator string) []*types.Any {
+	contents := make([]proto.Message, 0)
+	contents = append(contents, GenesisAccount(chainID, genesis))
```
ilgooz, line 106: Can't we use a single append for all?

```
append(contents,
GenesisAccount(chainID, genesis),
AccountRemoval(genesis),
...)
```

Aren't we gonna drop Any yet?

lubtd: > Aren't we gonna drop Any yet?

No, we're waiting Frojdi coming back to have his mind about this.
For now we finish the messages implementation using Any

x/profile/keeper/coordinator.go
```diff
@@ -103,8 +103,11 @@ func (k Keeper) GetAllCoordinator(ctx sdk.Context) (list []types.Coordinator) {
 }
 
 // GetCoordinatorAddressFromID returns a coordinator address from its id
-func (k Keeper) GetCoordinatorAddressFromID(ctx sdk.Context, id uint64) string {
-	return k.GetCoordinator(ctx, id).Address
+func (k Keeper) GetCoordinatorAddressFromID(ctx sdk.Context, id uint64) (string, bool) {
+	if !k.HasCoordinator(ctx, id) {
```
lubtd, line 107: I may need to see if we can remove this method as well from Starport template

lubtd (approved):
Great work👏

testutil/sample/launch.go
Pantani: @ilgooz, done on [`5906f90`](https://github.com/tendermint/spn/pull/235/commits/5906f90b4fc8f4214db3683ec2cd7fcdb15441d6)

x/profile/keeper/coordinator.go
Pantani: Are you mean the `Has<Object>` method? I think we can put it together with the `Get<Object>` method, as the map type. What do you think?

E.g:
```golang
GetCoordinator(ctx sdk.Context, id uint64) (coord types.Coordinator, found bool)
```

lubtd (approved):

ilgooz (approved):

x/profile/keeper/coordinator.go
lubtd: Yes I was thinking standardising the generated method list independently from the type. This is not a big priority

PR #237, opened 2021-08-09
refactor(launch): restrict chain name to only lowercase character

lubtd:
Restrict the chain name to only lowercase character

### DIFF
--- a/x/launch/types/chain.go
+++ b/x/launch/types/chain.go
@@ -9,7 +9,11 @@ import (
 	codec "github.com/cosmos/cosmos-sdk/codec/types"
 )
 
-const chainIDSeparator = "-"
+const (
+	chainIDSeparator = "-"
+	chainNameMaxLength = 30
+)
+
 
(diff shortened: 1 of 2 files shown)


x/launch/types/chain.go
```diff
@@ -9,7 +9,10 @@ import (
 	codec "github.com/cosmos/cosmos-sdk/codec/types"
 )
 
-const chainIDSeparator = "-"
+const (
+	ChainIDSeparator   = "-"
+	ChainNameMaxLength = 30
```
Pantani, line 14: Is this const need to be public/exported?

Pantani (approved):
💪🏽

only one doubt

x/launch/types/chain.go
lubtd: In order to use it in tests for edge case (trying to create a name with `ChainNameMaxLength +1`)

ilgooz (approved):

PR #240, opened 2021-08-11
feat(launch): Approve automatically requests sent by the coordinator

### LINKED ISSUE #227: Approve automatically requests sent by coordinator
Requests sent by coordinator (if the message is valid) are not added to the request pool but directly approved.

We should implement this once all request and `MsgSettleRequest` messages are implemented

Pantani:
Closes #227

### What does this PR does?

Automatically approve requests sent by the coordinator.


proto/launch/tx.proto
```diff
@@ -36,6 +36,7 @@ message MsgRequestAddValidator {
 
 message MsgRequestAddValidatorResponse {
   uint64 requestID = 1;
+  bool autoApproved = 2;
```
lubtd, line 39: Haven't thought about it.
That can actually be useful if we add later different logic for auto approval like allowances

proto/launch/tx.proto
```diff
@@ -36,6 +36,7 @@ message MsgRequestAddValidator {
 
 message MsgRequestAddValidatorResponse {
```
lubtd, line 37: I think we can create a single proto message `MsgRequestResponse` for all request msgs

x/launch/keeper/msg_server_request_add_account.go
```diff
@@ -33,14 +35,33 @@ func (k msgServer) RequestAddAccount(
 		return nil, sdkerrors.Wrap(types.ErrCodecNotPacked, msg.String())
 	}
 
-	requestID := k.AppendRequest(ctx, types.Request{
+	coordAddress, found := k.profileKeeper.GetCoordinatorAddressFromID(ctx, chain.CoordinatorID)
+	if !found {
+		return nil, spnerrors.Critical(
```
lubtd, line 40: This one is actually not critical because with `MsgDeleteCoordinator`. The coordinator associated to a chain can be delete

We should still return an error since the request will never beable to be approved since the associated coordinator has been deleted.
We can have a message like 
```
sdkerrors.Wrap(types.ErrChainInactive, "the chain coordinator has been deleted")
```

x/launch/keeper/msg_server_request_add_validator.go
```diff
@@ -33,14 +35,33 @@ func (k msgServer) RequestAddValidator(goCtx context.Context, msg *types.MsgRequ
 		return nil, sdkerrors.Wrap(types.ErrCodecNotPacked, msg.String())
 	}
 
-	requestID := k.AppendRequest(ctx, types.Request{
+	coordAddress, found := k.profileKeeper.GetCoordinatorAddressFromID(ctx, chain.CoordinatorID)
+	if !found {
+		return nil, spnerrors.Critical(
```
lubtd, line 40: Same as above and for all requests

x/launch/keeper/msg_server_request_add_account.go
lubtd: Based on this. We can also add a test for each request when the message fails when the coordinator ID is not associated to an address

lubtd (commented):
Tested every commands and it work well👍

Just few comments

proto/launch/tx.proto
Pantani: the reason is that when we auto-approve the request, we don't have the request-id. I think it is a good way to show the request is already approved

x/launch/keeper/msg_server_request_add_account.go
Pantani: solves on [ddcd930](https://github.com/tendermint/spn/pull/240/commits/ddcd930d7be0abf48ffb6fb4a49798b59fa299fc)

lubtd (approved):

ilgooz (approved):

PR #245, opened 2021-08-17
test: add tests for genesis export/import and create `keeper_test` package

lubtd:
Test the `InitGenesis` and `ExportGenesis` methods of each module.
We simply test that, for a given genesis, the exported state is the same as the one used for initialization. This template can be used for https://github.com/tendermint/starport/issues/1419

This change requires to do some changes and use keeper from the `keeper` package.
Therefore, the PR also introduces the `testutils/keeper` package to create keepers for testing purpose.
All the tests under `keeper` are moved in the `keeper_test` package


x/launch/genesis_test.go
```diff
@@ -0,0 +1,42 @@
+package launch_test
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+
+	testkeeper "github.com/tendermint/spn/testutil/keeper"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch"
```
Pantani, line 9:
```suggestion
	"testing"

	"github.com/stretchr/testify/require"
	testkeeper "github.com/tendermint/spn/testutil/keeper"
	"github.com/tendermint/spn/testutil/sample"
	"github.com/tendermint/spn/x/launch"
```

x/launch/keeper/genesis_validator_test.go
```diff
@@ -1,19 +1,22 @@
-package keeper
+package keeper_test
 
 import (
 	"strconv"
 	"testing"
 
+	testkeeper "github.com/tendermint/spn/testutil/keeper"
+
```
Pantani, line 8:
```suggestion
```

x/launch/keeper/request.go
```diff
@@ -133,7 +133,7 @@ func checkAccount(ctx sdk.Context, k Keeper, chainID, address string) (bool, err
 
 // applyRequest approves the request and performs
```
Pantani, line 134:
```suggestion
// ApplyRequest approves the request and performs
```

x/launch/keeper/request.go
```diff
@@ -133,7 +133,7 @@ func checkAccount(ctx sdk.Context, k Keeper, chainID, address string) (bool, err
...
 // the launch information changes
-func applyRequest(
+func ApplyRequest(
```
Pantani, line 136: I like the idea of a separete package for tests, but only for black-box tests, because if we do it for everything, we need to export (make public) all methods. It can be harder in the future when the application starts to grow a lot.

Pantani (changes requested):

x/launch/keeper/request.go
lubtd: Yes but I think most methods can be made public in the keeper. For `ApplyRequest ` it totally makes sense in my mind since this is an action that could be performed by another module in theory.

Internal methods should be used as helper, and if there are few we can cover them by testing the high-level methods

lubtd: Also using the same package would require having the method returning a keeper instance for testing purpose in the same package, I'm not a big fan of this idea

Pantani (approved):

testutil/sample/launch.go
```diff
@@ -207,3 +216,67 @@ func GenesisHash() string {
...
+}
+
+// LaunchGenesisState returns a sample genesis state for the launch module
+func LaunchGenesisState() launch.GenesisState {
+	chainID1, _ := ChainID(0)
+	chainID2, _ := ChainID(0)
+
+	return launch.GenesisState{
+		ChainList: []launch.Chain{
+			*Chain(chainID1, Uint64()),
```
ilgooz, line 238: Does `Chain`, `GenesisAccount` and others actually needs to return pointers in the first place?

ilgooz (approved):

testutil/sample/launch.go
lubtd: It's a good point. It used to be more convenient to use pointer since we used them but now we may have some type where it's better to return the value https://github.com/tendermint/spn/issues/254

PR #247, opened 2021-08-17
refactor(test): use `require` package instead `assert`

### LINKED ISSUE #213: tests: use a single package (`require` or `assert`) for `testify`
We have occurrences of `require` and `assert` in the tests.

For consistency I propose we use only one of them everywhere. My preference would be for `require` that sounds cleaner since tests stop under a negative assertion

Pantani:
Closes #213

### What does this PR does?

Use the `require` package from `testify` instead the `assert` package;

### How to test?

```shell
$ make test
```


lubtd (approved):
I assert this PR is legit

ilgooz (approved):

PR #250, opened 2021-08-19
refactor: move from proto `Any` to `oneof`

lubtd:
Migrate the interface representations from proto `Any` to `oneof`

It took a bit more time than expected because there is a major difference with `oneof`:
The getter is automatically generated from proto but we need to implement our own setter (unlike `Any` where we set the type with `NewAnyFrom` but we implemented our own getter (unpack))

So we also implemented setters and tests for those


Pantani (dismissed):

ilgooz (approved):

x/launch/keeper/msg_server_request_add_vested_account_test.go
```diff
@@ -14,13 +13,13 @@ import (
...
-		addr2                      = sample.AccAddress()
-		addr3                      = sample.AccAddress()
-		k, pk, srv, _, sdkCtx, cdc = setupMsgServer(t)
-		ctx                        = sdk.WrapSDKContext(sdkCtx)
+		invalidChain, _          = sample.ChainID(0)
+		coordAddr                = sample.AccAddress()
+		addr1                    = sample.AccAddress()
+		addr2                    = sample.AccAddress()
+		addr3                    = sample.AccAddress()
+		k, pk, srv, _, sdkCtx, _ = setupMsgServer(t)
```
Pantani, line 21: Maybe we should remove the `codec.Marshaler` parameter from the `setupMsgServer` method. We will not use it anymore

Pantani (approved):
🤙

x/launch/keeper/msg_server_request_add_vested_account_test.go
lubtd: We can actually check for this. I will merge this PR so that we can already resolve potential conflicts

PR #253, opened 2021-08-19
refactor(tests): Assert only the error type using the `require.ErrorIs` method

Pantani:
### What does this PR does?

Assert only the error type using the `require.ErrorIs` method from `testify` and remove the error message assertion.

### How to test?

```shell
$ make test
```


ilgooz (approved):

lubtd (approved):
This is great, I think we're with a good design now, where we also test the reason of the error.

PR #290, opened 2021-09-06
feat(campaign): create `shares` type

lubtd:
`Shares` simply reuse the type `sdk.Coins` since it represents denoms associated with an amount.

We add here the prefix `s/` to distinguish it from other coins in the store

### DIFF
--- a/x/campaign/types/shares.go
+++ b/x/campaign/types/shares.go
@@ -3,8 +3,9 @@ package types
 import (
 	"errors"
 	"fmt"
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	"strings"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // Shares represents the portion of a supply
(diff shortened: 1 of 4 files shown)


x/campaign/types/shares.go
```diff
@@ -0,0 +1,98 @@
...
+	return Shares(decreasedCoins), nil
+}
+
+// IsTotalReached checks if the provided shares overflow the total number of shares
+// Denoms not specified in totalShares uses DefaultTotalShareNumber as the default number of total shares
+func IsTotalReached(shares, totalShares Shares) bool {
+	// Check the explicitly defined total shares
+	totalMap := make(map[string]uint64)
+	for _, coin := range totalShares {
+		totalMap[coin.Denom] = coin.Amount.Uint64()
```
Pantani, line 78: the `shares` can have duplicate coins into the `totalShares`?

```suggestion
		totalMap[coin.Denom] += coin.Amount.Uint64()
```

lubtd: Nope.

The `sdk.Coins` can theoretically have duplicated denom and it can be handled by `IsValid()`. However, this `IsTotalReached` method is a low-level function so we shouldn't do this check here. It would rather be checks in methods in `keeper` that returns a critical error

Pantani (approved):

ilgooz (approved):

PR #293, opened 2021-09-08
fix: add placeholders to support genesis test scaffolding

lubtd:
Add necessary placeholders in `genesis_test.go` and `types/genesis_test.go` to support genesis test scaffolding from Starport

One placeholder in `genesis_test.go` is placed in a comment block because we don't need it (we use a sample genesis state from the `sample` library) we may need to think about the capability to skip placeholders in Starport


Pantani (approved):
LGTM 👌🏽

ilgooz (approved):

PR #298, opened 2021-09-08
feat(campaign): initialize mainnet account

### LINKED ISSUE #258: campaign: initialize mainnet accounts
Initialize the structure for mainnet accounts.

This is a map indexed by the campaignID and the address of the account

Pantani:
Closes #258

### What does this PR does?

Initialize the map for mainnet accounts


x/campaign/genesis_test.go
```diff
@@ -9,15 +9,36 @@ import (
...
+	{
+		CampaignID: 0,
+Address: "0",
+},
+	{
+		CampaignID: 1,
+Address: "1",
+},
+},
+
```
lubtd, line 25:
```suggestion
```

x/campaign/genesis_test.go
```diff
@@ -9,15 +9,36 @@ import (
...
 	got := *campaign.ExportGenesis(ctx, *keeper)
 
-	require.Len(t, got.CampaignList, len(original.CampaignList))
-	require.Subset(t, original.CampaignList, got.CampaignList)
+	// this line is used by starport scaffolding # genesis/test/assert
+	require.Len(t, got.CampaignList, len(genesisState.CampaignList))
+	require.Subset(t, genesisState.CampaignList, got.CampaignList)
+
+	require.Equal(t, genesisState.CampaignCount, got.CampaignCount)
+	require.Equal(t, genesisState, got)
```
lubtd, line 40:
```suggestion
```

x/campaign/keeper/mainnet_account_test.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// Prevent strconv unused error
+var _ = strconv.IntSize
+
+func createNMainnetAccount(keeper *keeper.Keeper, ctx sdk.Context, n int) []types.MainnetAccount {
+	items := make([]types.MainnetAccount, n)
+	for i := range items {
+		items[i].CampaignID = uint64(i)
+		items[i].Address = strconv.Itoa(i)
```
lubtd, line 21: We should test it with a real address like with `sample.AccAddress()`

One issue if we do this is that the `TestMainnetAccountQueryPaginated` above will fail because the order of the elements in the store will not be predictable.
I'm personally in the opinion that `TestMainnetAccountQueryPaginated` can be removed for now if it is problematic. Test for the `cli` package are not the most important priority

x/campaign/keeper/mainnet_account_test.go
```diff
@@ -0,0 +1,60 @@
...
+			item.Address,
+		)
+		require.False(t, found)
+	}
+}
+
+func TestMainnetAccountGetAll(t *testing.T) {
+	k, ctx := keepertest.Campaign(t)
+	items := createNMainnetAccount(k, ctx, 10)
+	require.Equal(t, items, k.GetAllMainnetAccount(ctx))
```
lubtd, line 59: This is something we must also change in Starport template.

Since order of elements in the store is in theory not obvisouse the same as the order from `createNMainnetAccount` we should use the combination
```
require.Len...
require.Subset...
```
To test both arrays contain same elements independently from the orders

lubtd (commented):
👍

x/campaign/keeper/mainnet_account_test.go
lubtd: Well for `TestMainnetAccountQueryPaginated ` we can in fact just check if each paginated result is a subset of the whole testcase array

Pantani: Great, I made the changes here. Also, I check if the result count is less or equal to the step

https://github.com/tendermint/spn/pull/298/commits/0344a0b7faaee8b1dd0dc7c737679b4aebb2e2e0

x/campaign/keeper/mainnet_account_test.go
Pantani: we can use `require.ElementsMatch` for this case

lubtd: WWOOO I didn't know this one😅

lubtd (approved):

ilgooz (approved):

PR #306, opened 2021-09-11
refactor(test): improve test templates

Pantani:
### What does this PR does?

- Check if the paginated result is a subset of all messages; 
- Check element from get all methods matches independently of the sort;


lubtd (approved):

lubtd:
Everything looks good. I would suggest to fix this one first: https://github.com/tendermint/spn/issues/312 since the issue seems to be related using `sample.AccAddress()` in the genesis of the cli tests

PR #314, opened 2021-09-27
feat(campaign): create the voucher methods

### LINKED ISSUE #313: campaign: initialize voucher type
Initialize the voucher type and create methods to create vouchers from denom, validate vouchers, etc...

Pantani:
Closes #313

### What does this PR does?

- Initialize the voucher type;
- Create `NewVoucher` methods;
- Create validate method;
- Create base math methods;


lubtd (commented):
I may have created the https://github.com/tendermint/spn/issues/313 issue a bit too quickly but I don't think we would need a new type at the end because the vouchers will always be handled as `sdk.Coins`. The mint/burn will use bank methods so we must use `sdk.Coins` type also we never store this type somewhere in the store unlike shares.

I think we can implement methods that uses `sdk.Coins`. From what I can see the only required methods would be:
- `CheckVouchers` a method that takes a `sdk.Coins` and checks all coins has the voucher prefic in the denom
- `SharesToVouchers` converts shares with `s/` to vouchers with `v/`, necessary for `MintVouchers`
- `VouchersToShares` converts voucher coins into shares, necessary for redeem and burn

And normally that should do.

Also we can implement everything with `sdk.Coins` instead of `sdk.Coin` since all messages handle set of vouchers

Pantani:
> * VouchersToShares

Great, it's better if we don't need to create a new type. If we are using `sdk.Coins`, I don't see any other method we need. The Coins object itself already implements all necessary methods.

x/campaign/types/voucher.go
```diff
@@ -0,0 +1,56 @@
...
+		splitDenom := strings.Split(coin.Denom, "/")
+		coinName := splitDenom[len(splitDenom)-1]
+		coin.Denom = SharePrefix + coinName
+		shares[i] = coin
+	}
+	return shares
+}
+
+// VoucherName returns the Voucher name with prefix
+func VoucherName(campaignID uint64, coin string) string {
```
lubtd, line 39: Very good point, we also need to include the campaign ID in the prefix

x/campaign/types/voucher.go
```diff
@@ -0,0 +1,56 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+const (
+	// VoucherPrefix is the prefix used to represent a voucher denomination
+	VoucherPrefix = "v/"
+)
+
+// SharesToVouchers returns new Coins vouchers from the Shares representation
+func SharesToVouchers(campaignID uint64, shares Shares) sdk.Coins {
```
lubtd, line 16: For both conversion methods, I think we should include a verification that shares are valid shares and same for vouchers and return errors.

We can then use critical errors if shares in the store are invalid shares

x/campaign/types/voucher.go
```diff
@@ -0,0 +1,56 @@
...
+		vouchers[i] = coin
+	}
+	return vouchers
+}
+
+// VouchersToShares returns new Shares from the Coins vouchers representation
+func VouchersToShares(vouchers sdk.Coins) Shares {
+	shares := make(Shares, len(vouchers))
+	for i, coin := range vouchers {
+		splitDenom := strings.Split(coin.Denom, "/")
```
lubtd, line 30: I think it would be proper to have a common method to compute the prefix of the voucher and use `TrimPrefix`

x/campaign/types/voucher.go
```diff
@@ -0,0 +1,56 @@
...
+	return shares
+}
+
+// VoucherName returns the Voucher name with prefix
+func VoucherName(campaignID uint64, coin string) string {
+	return fmt.Sprintf("%s%d/%s", VoucherPrefix, campaignID, coin)
+}
+
+// CheckVoucher checks if given Voucher are valid Voucher
+func CheckVoucher(campaignID uint64, vouchers sdk.Coins) error {
```
lubtd, line 44:
```suggestion
func CheckVouchers(campaignID uint64, vouchers sdk.Coins) error {
```

x/campaign/types/voucher.go
lubtd:
```suggestion
func VoucherName(campaignID uint64, denom string) string {
```

x/campaign/types/voucher_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+	campaign "github.com/tendermint/spn/x/campaign/types"
+)
+
+var (
+	campaignID         = uint64(10)
+	prefixedVoucherFoo = fmt.Sprintf("%s%d/foo", campaign.VoucherPrefix, campaignID)
```
lubtd, line 15: We should use `VoucherName`

lubtd:
> The Coins object itself already implements all necessary methods.

that's totally fine then🙂

x/campaign/types/voucher.go
```diff
@@ -0,0 +1,69 @@
...
+
+const (
+	// VoucherPrefix is the prefix used to represent a voucher denomination
+	VoucherPrefix = "v/"
+)
+
+// SharesToVouchers returns new Coins vouchers from the Shares representation
+func SharesToVouchers(campaignID uint64, shares Shares) (sdk.Coins, error) {
+	err := CheckShares(shares)
+	if err != nil {
```
lubtd, line 18:
```suggestion
	if err := CheckShares(shares); err != nil {
```

x/campaign/types/voucher.go
```diff
@@ -0,0 +1,69 @@
...
+		coin.Denom = VoucherDenom(campaignID, denom)
+		vouchers[i] = coin
+	}
+	return vouchers, nil
+}
+
+// VouchersToShares returns new Shares from the Coins vouchers representation
+func VouchersToShares(campaignID uint64, vouchers sdk.Coins) (Shares, error) {
+	err := CheckVouchers(campaignID, vouchers)
+	if err != nil {
```
lubtd, line 33:
```suggestion
	if err := CheckVouchers(campaignID, vouchers); err != nil {
```

x/campaign/types/voucher.go
```diff
@@ -0,0 +1,69 @@
...
+
+// VoucherToShareDenom remove the voucher prefix and add the share prefix
+func VoucherToShareDenom(campaignID uint64, denom string) string {
+	prefix := VoucherDenom(campaignID, "")
+	shareDenom := strings.TrimPrefix(denom, prefix)
+	return SharePrefix + shareDenom
+}
+
+// CheckVouchers checks if given Vouchers are valid
+func CheckVouchers(campaignID uint64, vouchers sdk.Coins) error {
```
lubtd, line 57:
```suggestion
func CheckVouchers(vouchers sdk.Coins, campaignID uint64) error {
```

lubtd (approved):

ilgooz (approved):

lubtd (approved):

PR #315, opened 2021-09-28
feat(invariants): add module invariants

Pantani:
### What does this PR does?

Add `profile`, `launch` and `campaign` module [invariants](https://docs.cosmos.network/master/building-modules/invariants.html)


lubtd (approved):
Thanks for having investigated into this👍

I think those current ones are good, though I'm wondering if the unknown request is critical to have but we can finalize all of them at the end of the Milestone

lubtd:
For Unknow Request, I'm wondering this because too much invariants would theoretically slow down the simulation process

Pantani:
> For Unknow Request, I'm wondering this because too much invariants would theoretically slow down the simulation process

I thought the same, but I don't know the actual cost of have invariants. We can delete the entire method or keep only the interface validate call:

```go

// UnknownRequestTypeInvariant invariant that checks if the Request
// type is valid
func UnknownRequestTypeInvariant(k Keeper) sdk.Invariant {
	return func(ctx sdk.Context) (string, bool) {
		all := k.GetAllRequest(ctx)
		for _, request := range all {
			if err := request.Content.Validate(); err != nil {
				return sdk.FormatInvariant(
					types.ModuleName, "unknown-request-type",
					"invalid request",
				), true
			}
		}
		return "", false
	}
}
```

lubtd:
> I thought the same, but I don't know the actual cost of have invariants. We can delete the entire method or keep only the interface validate call

We can check this afterward since it's trivial to disable a invariant

x/launch/keeper/invariants.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+// RegisterInvariants registers all module invariants
+func RegisterInvariants(ir sdk.InvariantRegistry, k Keeper) {
+	ir.RegisterRoute(types.ModuleName, "zero-launch-timestamp",
```
ilgooz, line 12: Maybe we can move routes to constants and reuse in the other part of the code?

Pantani: I like the solution, but the methods are hard for we can reuse. There is too much specific logic for being reusable. I used the same pattern from the [cosmos-sdk modules](https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/keeper/invariants.go).

Pantani:
added invariants for the campaign module too

lubtd (approved):

ilgooz (approved):

PR #321, opened 2021-09-29
feat(launch): allows to provide a campaign ID to a chain

### LINKED ISSUE #261: launch: allows to provide a campaign ID to a chain
When a chain is created with `MsgCreateChain`. The coordinator can provide a campaign ID.

The coordinator of the chain must be the coordinator of the campaign.

The chainID is appended to the campaignChains list in `campaign` module

Pantani:
Closes #261

### What does this PR does?

associate a chain to campaign passing a campaign id to `MsgCreateChain` method


ilgooz (approved):

x/launch/keeper/msg_server_create_chain.go
```diff
@@ -35,6 +36,18 @@ func (k msgServer) CreateChain(goCtx context.Context, msg *types.MsgCreateChain)
...
+		if !found {
+			campaign = campaigntypes.CampaignChains{
+				CampaignID: msg.CampaignID,
+				Chains:     []uint64{},
+			}
+		}
+		campaign.Chains = append(campaign.Chains, id)
+		k.campaignKeeper.SetCampaignChains(ctx, campaign)
+	}
+
```
lubtd, line 50:
```suggestion
```

The last argument of `CreateNewChain` are `hasCampaign`, `campaignID` and `isMainnet`. We must provide those values to this function with mainnet = false. The chain is added to the list in `cmapaign` inside this function

x/launch/keeper/msg_server_create_chain_test.go
```diff
@@ -79,6 +90,12 @@ func TestMsgCreateChain(t *testing.T) {
 
 			// Chain created from MsgCreateChain is never a mainnet
 			require.False(t, chain.IsMainnet)
+
+			if tc.msg.HasCampaign {
```
lubtd, line 94: We can add the check `require.False(t, chain.HasCampaign)` and `require.Equals(t, tc.msg.CampaignID, chain.CampaignID)`

ilgooz (approved):

x/launch/keeper/msg_server_create_chain.go
```diff
@@ -27,8 +27,8 @@ func (k msgServer) CreateChain(goCtx context.Context, msg *types.MsgCreateChain)
 		msg.SourceHash,
 		msg.GenesisURL,
 		msg.GenesisHash,
-		false,
```
lubtd, line 30: The call of `CreateNewChain` had previously all the preconditions so that it is supposed to never fail. Hence, the fact we were handling the error with a critical error.

However, now the method can fail because of a non-existing campaign.

Since a critical error should not be able to be triggered by a user, we should either handle this method with a normal error or add a check before to check the campaign exist and let the critical error here (because the preconditions would then be met again)

I'm in favor of the 2nd personally

lubtd: Also, the method can fail because the coordinator of the chain and the campaign don't match. So maybe it can be simpler to just directly handle this method with an normal error

x/launch/keeper/msg_server_create_chain_test.go
```diff
@@ -29,22 +35,27 @@ func TestMsgCreateChain(t *testing.T) {
...
+			msg:           sample.MsgCreateChain(coordAddress, "", false, 0),
 			wantedChainID: 1,
 		},
 		{
 			name:          "valid message with genesis url",
-			msg:           sample.MsgCreateChain(coordAddress, "foo.com"),
+			msg:           sample.MsgCreateChain(coordAddress, "foo.com", false, 0),
 			wantedChainID: 2,
 		},
+		{
```
lubtd, line 51: We can have two additional test cases:
- The campaign doesn't exist
- The coordinator of the campaign and the chain doesn't match

ilgooz (approved):

PR #325, opened 2021-10-01
feat(campaign): implement `MsgUnredeemVouchers`

lubtd:
Implement the message `MsgUnredeemVouchers` that allows a user to unredeem vouchers that have been redeemed into shares for the account.

It basically moves back the shares from an existing account into voucher coins to the account address on `spn`

This message has been implemented first because it can be tested in the CLI with `MsgAddShares`

To test it:

```
spnd tx profile create-coordinator --from alice
spnd tx campaign create-campaign toast 5000foo --from alice
spnd tx campaign add-shares 0 <bob_address> 1000foo --from alice
spnd tx campaign unredeem-vouchers  0 500foo --from bob

spnd q bank balances <bob_address>

balances:
- amount: "100000000"
  denom: stake
- amount: "10000"
  denom: token
- amount: "500"
  denom: v/0/foo
pagination:
  next_key: null
  total: "0"
🎉
```


x/campaign/client/cli/tx_unredeem_vouchers.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/tendermint/spn/x/campaign/types"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdUnredeemVouchers() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unredeem-vouchers [campaign-id] [shares]",
+		Short: "Unredeem vouchers that has been redeemed into an account and get vouchers back",
```
Pantani, line 20:
```suggestion
		Short: "Unredeem vouchers that have been redeemed into an account and get vouchers back",
```

x/campaign/keeper/msg_server_test.go
```diff
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
```
Pantani, line 4: wrong lint import sort

x/campaign/keeper/msg_server_unredeem_vouchers.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, sdkerrors.Wrap(types.ErrSharesDecrease, err.Error())
+	}
+
+	// If the account no longer has shares, it can be removed from the store
+	if types.IsEqualShares(account.Shares, types.EmptyShares()) {
+		k.RemoveMainnetAccount(ctx, msg.CampaignID, msg.Sender)
+	} else {
+		k.SetMainnetAccount(ctx, account)
+	}
```
Pantani, line 37: same comment as [here](https://github.com/tendermint/spn/pull/326/files#r720307032). we should save the account only if the account were already minted

x/campaign/types/message_unredeem_vouchers_test.go
```diff
@@ -0,0 +1,64 @@
+package types_test
+
+import (
+	"github.com/tendermint/spn/x/campaign/types"
+	"testing"
+
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
```
Pantani, line 9:
```suggestion
	"testing"

	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
	"github.com/stretchr/testify/require"
	"github.com/tendermint/spn/x/campaign/types"
	"github.com/tendermint/spn/testutil/sample"
```

x/campaign/keeper/msg_server_unredeem_vouchers.go
lubtd: As mentioned, the context is only saved if the message is successful. In this case here, I think it's more visual to make the verification just after having decreased the shares

ilgooz (approved):

Pantani (dismissed):

PR #332, opened 2021-10-05
feat(campaign): add shares invariant

Pantani:
### What does this PR does?

Add invariant to check if all account shares for a campaign are equal to the campaign shares.

### DIFF
--- a/x/campaign/keeper/invariants.go
+++ b/x/campaign/keeper/invariants.go
@@ -11,6 +11,7 @@ const (
 	campaignChainsWithoutCampaign = "campaign-chains-without-campaign"
 	accountWithoutCampaign        = "account-without-campaign"
 	vestingAccountWithoutCampaign = "vesting-account-without-campaign"
+	campaignShares                = "campaign-shares"
 )
 
 // RegisterInvariants registers all module invariants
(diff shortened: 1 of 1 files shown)


x/campaign/keeper/invariants.go
```diff
@@ -88,3 +91,48 @@ func VestingAccountWithoutCampaignInvariant(k Keeper) sdk.Invariant {
...
+				acc.Shares,
+			)
+		}
+		for _, acc := range vestingAccounts {
+			if _, ok := shares[acc.CampaignID]; !ok {
+				shares[acc.CampaignID] = types.EmptyShares()
+			}
+			shares[acc.CampaignID] = types.IncreaseShares(
+				shares[acc.CampaignID],
+				acc.Shares,
```
lubtd, line 118: We must use `GetTotalShares()`

x/campaign/keeper/invariants.go
```diff
@@ -88,3 +91,48 @@ func VestingAccountWithoutCampaignInvariant(k Keeper) sdk.Invariant {
...
+		}
+		for campaignID, share := range shares {
+			campaign, found := k.GetCampaign(ctx, campaignID)
+			if !found {
+				return sdk.FormatInvariant(
+					types.ModuleName, campaignShares,
+					fmt.Sprintf("%s: %d", types.ErrCampaignNotFound, campaignID),
+				), true
+			}
+			if !types.IsEqualShares(share, campaign.AllocatedShares) {
```
lubtd, line 129: We must also take into account the circulating vouchers.

We normally can query emitted tokens in `bank` from a denom with the `SupplyOf` query https://github.com/cosmos/cosmos-sdk/blob/1c468de77216e1da4c79c2a87a8e0ad307060185/x/bank/keeper/grpc_query.go#L90

lubtd (changes requested):

PR #374, opened 2021-11-01
chore(config): add chain-id

lubtd:
Add a default `chain-id` for `spn` that uses the `spn` format: `<name>-<networkID>`

Need to check if the `starport network` remains compatible after this change otherwise we need to make the change on `starport` before merging


ilgooz (approved):

lubtd:
We gotta merge this one to be able to experiment IBC connection

Pantani (approved):

PR #375, opened 2021-11-01
fix(`launch`): prevent hiding invalid chain id error

lubtd:
Make error messages more explicit in `chainid.go` and prevent hiding underlying chain ID error for `MsgCreateChain` and `MsgEditChain`


Pantani (approved):

pkg/chainid/chainid.go
```diff
@@ -24,14 +24,14 @@ func NewGenesisChainID(chainName string, networkNumber uint64) string {
 func ParseGenesisChainID(genesisChainID string) (string, uint64, error) {
 	parsed := strings.Split(genesisChainID, ChainIDSeparator)
 	if len(parsed) != 2 {
-		return "", 0, errors.New("incorrect chain ID format")
+		return "", 0, fmt.Errorf("%s: invalid chain ID, format must be <chain-name>-<network-number>", genesisChainID)
```
ilgooz, line 27: I think in general, it might be ideal to assign the underlying error `errors.New("incorrect chain ID format")` to a variable and use `errors.Wrapf(err, "custom %s", id)` so, in the client side it can be unwrapped to determine the exact error easily.

Pantani (approved):

ilgooz (approved):

ilgooz (approved):

PR #380, opened 2021-11-08
refactor(`campaign`): start with ID `1` for `campaign`

lubtd:
Set the campaign count value to `1` to start with `1` for the campaign ID

We should rename "count" into "counter" as it serves more as a counter to track IDs rather than count of created objects but we can handle it in another PR to have a fix for Starport CLI

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/google/go-cmp v0.5.6
 	github.com/gorilla/mux v1.8.0
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
+	github.com/pkg/errors v0.9.1
 	github.com/spf13/cast v1.3.1
 	github.com/spf13/cobra v1.1.3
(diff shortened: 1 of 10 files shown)


Pantani (approved):

Pantani:
we can also remove the `hasCampaign` field for Chain object

lubtd:
> we can also remove the [`hasCampaign` field](https://github.com/tendermint/spn/blob/87d0437c12e82af0f111471efae7806bc7a322f5/proto/launch/chain.proto#L24-L24) for the Chain object

I would prefer to keep it for two reasons
- The interface is kept explicit, we made the choice of keeping 0 out because of Go language but any app could do a GET request to get chain information and this would be too implicit
- We keep a design where the ID of the campaign is not related to other side effects, we can just update CampaignCount value without side-effect

Having an additional boolean for a chain is not significant given that kb of gentxs will be stored for a chai

PR #399, opened 2021-11-30
feat(launch): Setting request IDs starting from 1

Pantani:
Closes ##396

### What does this PR does?

avoid the request ID starting with zero.


lubtd (changes requested):
I think we should rather change `GetRequestCounter` and returns `1` when the counter is not initialized https://github.com/tendermint/spn/blob/aa02296826a8a6970f52a52b26ed6126042451a7/x/launch/keeper/request.go#L21

`Request` is an indexed list unlike `Chain` and there is not initial counter params like a regular list

ilgooz (approved):

lubtd (approved):
Works like a charm

PR #413, opened 2021-12-07
test: increase campaign module tests coverage

ivanovpetr:


x/campaign/keeper/msg_server_mint_vouchers.go
```diff
@@ -43,7 +43,7 @@ func (k msgServer) MintVouchers(goCtx context.Context, msg *types.MsgMintVoucher
 		return nil, spnerrors.Criticalf("verified shares are invalid %s", err.Error())
 	}
 	if err := k.bankKeeper.MintCoins(ctx, types.ModuleName, vouchers); err != nil {
-		return nil, sdkerrors.Wrap(types.ErrVouchersMinting, err.Error())
+		return nil, spnerrors.Criticalf("can't mint vouchers %s", err.Error())
```
lubtd, line 46: Same

lubtd (commented):
Nice work👍
Though mentioned in the issue, the coverage of  `MsgMintVouchers` and `MsgUnredeemVouchers` are already good.
I would let the current error for now, we should check if we can set it as a Critical Error: https://github.com/tendermint/spn/issues/414

x/campaign/keeper/msg_server_unredeem_vouchers.go
```diff
@@ -42,7 +42,7 @@ func (k msgServer) UnredeemVouchers(goCtx context.Context, msg *types.MsgUnredee
 		return nil, spnerrors.Criticalf("verified shares are invalid %s", err.Error())
 	}
 	if err := k.bankKeeper.MintCoins(ctx, types.ModuleName, vouchers); err != nil {
-		return nil, sdkerrors.Wrap(types.ErrVouchersMinting, err.Error())
+		return nil, spnerrors.Criticalf("can't mint vouchers %s", err.Error())
```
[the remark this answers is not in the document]
ivanovpetr: Okay. I revert it and create an issue for research.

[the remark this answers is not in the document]
lubtd: Already created https://github.com/tendermint/spn/issues/414 👍

lubtd (dismissed):
👍

x/campaign/types/genesis_test.go
```diff
@@ -140,11 +138,13 @@ func TestGenesisState_Validate(t *testing.T) {
 				},
 				CampaignCounter: 2,
 				MainnetAccountList: []types.MainnetAccount{
-					sample.MainnetAccount(330, sample.Address()),
-					sample.MainnetAccount(434, sample.Address()),
+					{
+						CampaignID: 330,
+						Address:    "330",
+					},
```
Pantani, line 144:
```suggestion
					sample.MainnetAccount(330, sample.Address()),
```

ivanovpetr: If `sample.Address()` is used I cannot assert exact error message which contain the exact address. I will use `sample.MainnetAccount()` but without `simple.Address()`

lubtd (approved):

x/campaign/types/genesis_test.go
Pantani: you can create the address in the top level of the function and reuse it in the test and the log

Pantani (approved):

PR #417, opened 2021-12-08
feat: adapt delayed vesting options to the chain command interface

lubtd:
For `DelayedVesting` in `launch` and `ShareDelayedVesting` in `campaign`:
- Remove `StartingBalance` value
- Add `TotalBalance` in the option

This is to adapt to the `add-genesis-account` command interface that needs the total balance and vesting balance as argument


ilgooz (approved):

x/campaign/types/mainnet_vesting_account.go
```diff
@@ -1,11 +1,12 @@
...
+import "errors"
+
 // GetTotalShares return total shares for account and delayed vesting options
 func (m MainnetVestingAccount) GetTotalShares() (Shares, error) {
-	vestingShares, err := m.VestingOptions.GetDelayedVestingShare()
-	if err != nil {
-		return nil, err
+	dv := m.VestingOptions.GetDelayedVesting()
+	if dv == nil {
+		return nil, errors.New("invalid vesting options type")
```
Pantani, line 9:
```suggestion
		return nil, ErrInvalidVestingOption
```

maybe should we create the `ErrInvalidVestingOption` error?

Pantani (approved):
💪🏽

x/campaign/types/mainnet_vesting_account.go
lubtd: I think it's okay. Methods in `types` are last-layer called methods so they are usually wrapped inside a higher-level error.
We generally define errors for `keeper` logic

PR #439, opened 2022-01-04
feat: initialize monitoring modules

lubtd:
- initialize the monitoring consumer module as `monitoringc`
- initialize the monitoring provider module as `monitoringp` with no app registration
- create test keepers
- refactor `test-keeper` package

Names for the monitoring modules can be discussed further, I would still propose to move forward for now

In my opinion, the best would be the following architecture
```
x/
  monitoring/
    consumer/
      ...
    provider/
      ...
```
However, we would not be able to use Starport scaffolding command for now


Pantani (approved):
👌🏽 LGMT

PR #440, opened 2022-01-04
feat: import `fundraising` module

lubtd:
Import latest version `fundraising` and set basic config

For testing: https://github.com/tendermint/fundraising/tree/main/docs/How-To/cli


Pantani (changes requested):
add the fundraising to `SetOrderEndBlockers` is missing
```go
app.mm.SetOrderEndBlockers(
		crisistypes.ModuleName,
		govtypes.ModuleName,
		stakingtypes.ModuleName,
		feegrant.ModuleName,
		fundraisingtypes.ModuleName,
	)
```

lubtd:
Very good point, it actually has an End-Blocker for distribution. It seems a lot of logic, I think we will need to discuss it

Pantani (approved):

Issue #442, opened 2022-01-07
Switch from `spm` to `starport`


lubtd:
Switch `spm` packages to `starport` once `v0.19.2` is released

PR #444, opened 2022-01-11
feat(`launch`): implement request-selection query

ivanovpetr:
When query contains id of any non-existent request all found request will be returned


proto/launch/query.proto
```diff
@@ -16,7 +16,12 @@ option go_package = "github.com/tendermint/spn/x/launch/types";
 
 // Query defines the gRPC querier service.
 service Query {
-  // this line is used by starport scaffolding # 2
+  // Queries a list of requestSelection items.
+	rpc RequestSelection(QueryRequestSelectionRequest) returns (QueryRequestSelectionResponse) {
+		option (google.api.http).get = "/tendermint/spn/launch/requestSelection";
+	}
+
+// this line is used by starport scaffolding # 2
```
Pantani, line 24:
```suggestion
	// Queries a list of requestSelection items.
	rpc RequestSelection(QueryRequestSelectionRequest) returns (QueryRequestSelectionResponse) {
		option (google.api.http).get = "/tendermint/spn/launch/requestSelection";
	}

	// this line is used by starport scaffolding # 2
```

x/launch/client/cli/query.go
```diff
@@ -35,6 +35,8 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 	cmd.AddCommand(CmdShowRequest())
 	cmd.AddCommand(CmdListRequest())
 	cmd.AddCommand(CmdQueryParams())
+	cmd.AddCommand(CmdRequestSelection())
```
Pantani, line 38: `AddCommand` is a variadic function. Can you put all `AddCommand` calls in the same one?
```go
	cmd.AddCommand(
		CmdShowChain(),
		CmdListChain(),
		CmdShowGenesisValidator(),
		CmdListGenesisValidator(),
		CmdShowVestingAccount(),
		CmdListVestingAccount(),
		CmdShowGenesisAccount(),
		CmdListGenesisAccount(),
		CmdShowRequest(),
		CmdListRequest(),
		CmdQueryParams(),
		CmdRequestSelection(),
	)
```

x/launch/keeper/grpc_query_request_selection.go
```diff
@@ -0,0 +1,57 @@
+package keeper
+
+import (
+	"context"
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/tendermint/spn/x/launch/types"
+	"github.com/tendermint/starport/starport/pkg/numbers"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
Pantani, line 10:
```suggestion
	"context"
	
	"github.com/cosmos/cosmos-sdk/store/prefix"
	sdk "github.com/cosmos/cosmos-sdk/types"
	"github.com/tendermint/spn/x/launch/types"
	"github.com/tendermint/starport/starport/pkg/numbers"
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
```

x/launch/keeper/grpc_query_request_selection_test.go
```diff
@@ -0,0 +1,90 @@
+package keeper_test
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"testing"
```
Pantani, line 10:
```suggestion
	"testing"
	
	sdk "github.com/cosmos/cosmos-sdk/types"
	"github.com/stretchr/testify/require"
	"github.com/tendermint/spn/testutil/sample"
	"github.com/tendermint/spn/x/launch/types"
	"google.golang.org/grpc/codes"
	"google.golang.org/grpc/status"
```

x/launch/client/cli/query.go
```diff
@@ -24,17 +24,21 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
...
+		CmdShowGenesisValidator(),
+		CmdListGenesisValidator(),
+		CmdShowVestingAccount(),
+		CmdListVestingAccount(),
+		CmdShowGenesisAccount(),
+		CmdListGenesisAccount(),
+		CmdShowRequest(),
+		CmdListRequest(),
+		CmdQueryParams(),
+		CmdRequestSelection(),
```
lubtd, line 39:
```suggestion
		CmdRequestSelection(),
		CmdQueryParams(),
```

x/launch/client/cli/query_request_selection.go
```diff
@@ -0,0 +1,52 @@
...
+	"strconv"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/tendermint/spn/x/launch/types"
+)
+
+var _ = strconv.Itoa(0)
```
lubtd, line 13:
```suggestion
```

This is generated by Starport CLI for some utilities like not throwing errors because of unused packages
It's not a big deal to have these in the code but don't hesitate to remove them when you spot them

x/launch/keeper/grpc_query_request_selection.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (k Keeper) RequestSelection(goCtx context.Context, req *types.QueryRequestSelectionRequest) (*types.QueryRequestSelectionResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	ids, err := numbers.ParseList(req.RequestIDs)
+	if err != nil || len(ids) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "invalid argument")
```
lubtd, line 22: We should include `err` in the returned error for better debugging

x/launch/keeper/grpc_query_request_selection.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func (k Keeper) RequestSelection(goCtx context.Context, req *types.QueryRequestSelectionRequest) (*types.QueryRequestSelectionResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	ids, err := numbers.ParseList(req.RequestIDs)
+	if err != nil || len(ids) == 0 {
```
lubtd, line 21: `len(ids) == 0`: returns an error `no request id provided`

x/launch/keeper/grpc_query_request_selection.go
```diff
@@ -0,0 +1,58 @@
...
+		var request types.Request
+		err := k.cdc.Unmarshal(iterator.Value(), &request)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		if _, ok := selectionMap[request.RequestID]; ok {
+			result = append(result, request)
+			selectionMap[request.RequestID] = struct{}{}
+		}
+	}
```
lubtd, line 47: The requests are directly indexed by their own ids. Therefore we can directly fetch them from the id like `Request` in `grpc_query_request.go`

proto/launch/query.proto
```diff
@@ -16,6 +16,11 @@ option go_package = "github.com/tendermint/spn/x/launch/types";
 
 // Query defines the gRPC querier service.
 service Query {
+  // Queries a list of requestSelection items.
+  rpc RequestSelection(QueryRequestSelectionRequest) returns (QueryRequestSelectionResponse) {
+    option (google.api.http).get = "/tendermint/spn/launch/requestSelection";
```
lubtd, line 21:
```suggestion
    option (google.api.http).get = "/tendermint/spn/launch/requestSelection/{launchID}/{requestIDs}";
```
We should put the rpc def just below `RequestAll` for keeping an intuitive order

proto/launch/query.proto
```diff
@@ -172,4 +177,13 @@ message QueryParamsResponse {
...
 
+message QueryRequestSelectionRequest {
+  uint64 launchID = 1;
+  string requestIDs = 2;
+}
+
+message QueryRequestSelectionResponse {
+  repeated Request request = 1 [(gogoproto.nullable) = false];
+}
+
```
lubtd, line 188: Same, we should put these messages definition right below `AllRequest` ones

lubtd (changes requested):
Some feedbacks
The query name should reflect an action: what about `SelectRequest` for the name of the query?

The query should fail if at list one id has not been retrieved because it means the user didn't get the data requested

x/launch/keeper/grpc_query_request_selection.go
ivanovpetr: Oh yeah! It is way easier this way.

ivanovpetr:
Speaking about naming I followed the idea the first word of the name is what type of entity you will get and the second word is form of the result. Querying one request the form is omited, querying all request you calling method RequestAll and querying some selection of request you calling request selection. What do you think about this consistency theory.

I do agree returning an error in case of some requested ids not found is better behaviour. But what error should look like in that case. It should contain no details, just "Not found" or it should return the first id in query which was not found back "id 5 i not found" or it should return all not found ids "ids 1,2,6,8 are not found". What do you think?

Pantani (dismissed):

x/launch/keeper/grpc_query_request_selection.go
```diff
@@ -0,0 +1,51 @@
...
+
+	result := make([]types.Request, 0)
+	for _, id := range ids {
+		var request types.Request
+
+		b := requestStore.Get(types.RequestIDBytes(id))
+
+		if b == nil {
+			// TODO: decide on error content
+			return nil, status.Error(codes.NotFound, "Not found")
```
lubtd, line 40: Can we return the first id not found as discussed?

x/launch/keeper/grpc_query_request_selection.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	if len(ids) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "no request id provided")
+	}
+
+	store := ctx.KVStore(k.storeKey)
+	keyPrefix := append(types.KeyPrefix(types.RequestKeyPrefix), types.RequestPoolKey(req.LaunchID)...)
+	requestStore := prefix.NewStore(store, keyPrefix)
```
lubtd, line 30: We can in fact directly use `k.GetRequest` to get the request

lubtd (changes requested):

lubtd (changes requested):
I realized `github.com/tendermint/starport/starport/pkg/numbers` sort the list of parsed numbers, therefore the returned requests will be in the increasing order from their request ids.

We should rather return the requests in the same order as the provided list of id because this query would be used by `starport network request simulate` and `starport network request approve` and the order of approved requests affects the output state.

ivanovpetr:
> I realized `github.com/tendermint/starport/starport/pkg/numbers` sort the list of parsed numbers, therefore the returned requests will be in the increasing order from their request ids.
> 
> We should rather return the requests in the same order as the provided list of id because this query would be used by `starport network request simulate` and `starport network request approve` and the order of approved requests affects the output state.

`starport network request simulate` and `starport network request approve` also use `github.com/tendermint/starport/starport/pkg/numbers` internally to parse provided request IDs and as a side effect sort them as well. So request will be sent to spn in the sorted order. I can miss something. Can you describe how order can affect output in this case in more details?

lubtd:
> starport network request simulate and starport network request approve also use github.com/tendermint/starport/starport/pkg/numbers internally to parse provided request IDs and as a side effect sort them as well. So request will be sent to spn in the sorted order.

Thanks for having spotted this then, we must change this behavior

>  I can miss something. Can you describe how order can affect output in this case in more details?

The order affects the output because the application of a request depends on an input state.
There isn't much problem now with the request types that we have. An example is add AddGenesisAccount following by a RemoveGenesisAccount of the same address would succeed while the other way would fail
A more important issue is if in the future we have a `ModifyParam` request, if two requests modify the same param, the last one applied would be effective

ivanovpetr:
> > starport network request simulate and starport network request approve also use github.com/tendermint/starport/starport/pkg/numbers internally to parse provided request IDs and as a side effect sort them as well. So request will be sent to spn in the sorted order.
> 
> Thanks for having spotted this then, we must change this behavior
> 
> > I can miss something. Can you describe how order can affect output in this case in more details?
> 
> The order affects the output because the application of a request depends on an input state. There isn't much problem now with the request types that we have. An example is add AddGenesisAccount following by a RemoveGenesisAccount of the same address would succeed while the other way would fail A more important issue is if in the future we have a `ModifyParam` request, if two requests modify the same param, the last one applied would be effective

So then I think the good idea would be to change logic of `github.com/tendermint/starport/starport/pkg/numbers` and use it as input parser here and in starport network commands.

lubtd:
> So then I think the good idea would be to change logic of github.com/tendermint/starport/starport/pkg/numbers and use it as input parser here and in starport network commands.

`numbers` might be used somewhere else, having an option or an additional method to not sort them in the package should be fine

ivanovpetr:
I have created [PR](https://github.com/tendermint/starport/pull/2031) which removes sort side effect. I plan to pull it here once it will be released in a new starport version.

lubtd:
> I have created [PR](https://github.com/tendermint/starport/pull/2031) which removes sort side effect. I plan to pull it here once it will be released in a new starport version.

We will need to upgrade starport and therefore upgrade IBC to v2. I'm checking it here https://github.com/tendermint/spn/issues/442 since we will also need this update for IBC packet. We can keep the PR on hold until then

Pantani (approved):

lubtd (approved):
Approved
We will just need to upgrade Starport. I believe, the current version has no sort yet

Issue #430, opened 2022-01-03
Implement `MsgSetValidatorConsAddress` message


lubtd:
Depends on https://github.com/tendermint/spn/issues/429

`MsgSetValidatorConsAddress` allows a validator to associate a consensus address with its validator address on `spn` in order to receive reward from testnet incentivization

To prove consensus key ownership, a signature must be provided in the message.

 This signature should represent the signature of constant to be defined associated with a Nonce (to allows reassociating the consensus key with a new address). To keep track of the nonce a `map` should be scaffolded `ConsensusKeyNonce`

If the signature is verify, validator profile is updated

PR #451, opened 2022-01-14
feat(profile): add `ConsensusKeyNonce` structure

Pantani:
part of #430

### What does this PR does?

Creates the `ConsensusKeyNonce` map structure into the profile module


lubtd:
Thanks for creating this PR, it seems like I incorporated this structure into https://github.com/tendermint/spn/issues/430 while it should actually be a separate PR

x/profile/types/genesis.go
```diff
@@ -50,12 +61,20 @@ func (gs GenesisState) ValidateValidators() error {
 
 		consAddrIndex := string(ValidatorByConsAddressKey(elem.ConsensusAddress))
 		if _, ok := validatorByConsAddressIndexMap[consAddrIndex]; !ok {
-			return fmt.Errorf("validator consensus address not found for ValidatorByConsAddress: %s", elem.Address)
+			return fmt.Errorf("validator consensus address not found for ValidatorByConsAddress: %s", elem.ConsensusAddress)
+		}
+
+		consNonceIndex := string(ConsensusKeyNonceKey(elem.ConsensusAddress))
+		if _, ok := consensusKeyNonceIndexMap[consNonceIndex]; !ok {
+			return fmt.Errorf("validator consensus address not found for ConsensusKeyNonce: %s", elem.ConsensusAddress)
```
lubtd, line 69: I think it's ok from the code but just to be sure:

A validator consensus address should always be associated to a validator address and a consensus key nonce should always be associated to a validator consensus address

But a validator address can eventually no be associated with any validator consensus address

lubtd (commented):
👍

lubtd (approved):

PR #453, opened 2022-01-18
feat(profile): Implement `MsgSetValidatorConsAddress` message

### LINKED ISSUE #430: Implement `MsgSetValidatorConsAddress` message

Pantani:
Closes #430

### What does this PR does?

Implement `MsgSetValidatorConsAddress` message

### How to test?

- Create the validator key
```shell
echo '{"address":"B4AAC35ED4E14C09E530B10AF4DD604FAAC597C0","pub_key":{"type":"tendermint/PubKeyEd25519","value":"sYTsd7W1+SBtjD3BN/aTEDFvfRbZ9zdfpQH2Lk3MRK4="},"priv_key":{"type":"tendermint/PrivKeyEd25519","value":"j45JhnCflEk3T6FC8LLuJqg9tPfCzJH+UYZY88xn+0exhOx3tbX5IG2MPcE39pMQMW99Ftn3N1+lAfYuTcxErg=="}}' > validator_key.json
```

- Run the command below using the validator signature
```shell
spnd tx profile set-validator-cons-address validator_key.json 0 --signature "jbEFUseQ0xHwbCNNB2uRId1OrsSeDZ6fOyEO8KEfdvNdCCvIcSuuwdPQClyPF5NGgoXTmb8NMV/PVWRy4S3IBA==" --from alice
```

- Run the command passing only the nonce arg with the next nonce
```shell
spnd tx profile set-validator-cons-address validator_key.json 1 --from alice
```


barriebyron (commented):
I love grammar, so always look for how every word is exposed to our users 

nits and editorial suggestions :)

lubtd (changes requested):
An optimization we can have is to add the `nonce` and `chainID` in the `MsgSetValidatorConsAddress ` message. So that we can use `CreateSignMessage` to generate the signMessage in `ValidateMessage` and  verify the signature as a static check

Then, in the message handler we simply check if the nonce used is the current nonce for the consKey and the chainID is `ctx.ChainID()` 

WDYT?

Pantani:
> An optimization we can have is to add the `nonce` and `chainID` in the `MsgSetValidatorConsAddress ` message. So that we can use `CreateSignMessage` to generate the signMessage in `ValidateMessage` and verify the signature as a static check
> 
> Then, in the message handler we simply check if the nonce used is the current nonce for the consKey and the chainID is `ctx.ChainID()`
> 
> WDYT?

Sure!! added. This new implementation also supports different type of keys

proto/profile/tx.proto
```diff
@@ -32,6 +33,16 @@ message MsgDeleteValidator {
 
 message MsgDeleteValidatorResponse {}
 
+message MsgSetValidatorConsAddress {
+  string validatorAddress = 1;
+  bytes validatorPubKey = 2;
```
lubtd, line 38:
```suggestion
  bytes validatorConsPubKey = 2;
```

We should precise this is the consensus key which is different from the validator SDK address

x/profile/client/cli/tx_set_validator_cons_address.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+const (
+	flagSignature = "signature"
+)
+
+func CmdSetValidatorConsAddress() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set-validator-cons-address [validator-key] [nonce]",
+		Short: "Associate a consensus address for a specific SPN address",
```
lubtd, line 22:
```suggestion
		Short: "Associate a Tendermint consensus address to a specific validator address on SPN",
```

x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -0,0 +1,72 @@
...
+
+	valPubKey, err := valtypes.NewValidatorConsPubKey(msg.ValidatorConsPubKey, msg.ValidatorKeyType)
+	if err != nil {
+		return &types.MsgSetValidatorConsAddressResponse{},
+			sdkerrors.Wrap(types.ErrInvalidValidatorKey, string(msg.ValidatorConsPubKey))
+	}
+	consAddress := valPubKey.GetConsAddress().String()
+
+	// cannot set the consensus key if key is used for another validator
+	validatorByConsAddr, found := k.GetValidatorByConsAddress(ctx, consAddress)
```
lubtd, line 26: I think we should allow the validator to set a new validator address for the consensus key since the signature prove they are the owner of it

x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	validator := types.Validator{
+		Address:          valPubKey.Address().String(),
+		ConsensusAddress: consAddress,
+		Description:      types.ValidatorDescription{},
+	}
+
+	// get the current validator to eventually overwrite description
+	validatorStore, found := k.GetValidator(ctx, validatorByConsAddr.ValidatorAddress)
```
lubtd, line 55:
```suggestion
	validatorStore, found := k.GetValidator(ctx, msg.ValidatorAddress)
```
Isn't a mistake here?

x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -0,0 +1,72 @@
...
+	if found {
+		validator.Description = validatorStore.Description
+	}
+
+	// store validator information
+	k.SetValidator(ctx, validator)
+	k.SetValidatorByConsAddress(ctx, types.ValidatorByConsAddress{
```
lubtd, line 62: We should also reset `ValidatorByConsAddress` if there is an existing consensus address associated to the address

lubtd:
> > An optimization we can have is to add the `nonce` and `chainID` in the `MsgSetValidatorConsAddress ` message. So that we can use `CreateSignMessage` to generate the signMessage in `ValidateMessage` and verify the signature as a static check
> > Then, in the message handler we simply check if the nonce used is the current nonce for the consKey and the chainID is `ctx.ChainID()`
> > WDYT?
> 
> Sure!! added. This new implementation also supports different type of keys

I meant doing the signature verification
```
if !valPubKey.VerifySignature(currentNonce, ctx.ChainID(), msg.Signature) {
return &types.MsgSetValidatorConsAddressResponse{},
	sdkerrors.Wrapf(types.ErrInvalidValidatorSignature,
		"invalid signature for consensus address: %s - %s",
		consAddress,
		msg.Signature,
	)
}
```
in the `Validate` of `types/message_set_validator_cons_address.go` instead of having it in `keeper/msg_server_set_validator_cons_address.go` since it's a static check once we have the nonce and the chain ID. So we would add the nonce and the chain ID in the message `MsgSetValidatorConsAddress ` (not the message to sign)

and then we check only the nonce and chain-id in `keeper/msg_server_set_validator_cons_address.go`

Pantani:
@lubtd sorry, I understand the suggestion now. I prefer the user pass only the key, and we can get the chain id and the nonce automatically in the keeper, so the user doesn't need to query the `chainID` and `nonce`, which is more user-friendly. And we can add the nonce and the chain-id as an optional parameter for the CLI.

lubtd:
> so the user doesn't need to query the chainID and nonce

Mmh, the user would need to query the nonce at some point to generate the signature anyway?
This is an internal change mostly so we could maybe discuss it in the next SPN call

Pantani:
@lubtd, I will apply your suggestion, and we can discuss the improvement later in the call. If we make the signature validation into the `ValidateBasic` method, we should remove the `signature` CLI flag because the logic will be hard to handle

lubtd:
Sorry, I didn't precise in my latest message, I wanted to say that we could go the current way for this PR and discuss the eventual change in the next call

> we should remove the signature CLI flag because the logic will be hard to handle

In this case the `nonce` would become a required argument. The flag `--signature` could still be used, the signature would not be generated in the CLI method

Pantani:
But don't have too much sense to have the signature flag if we need to always pass the nonce to validate into the `ValidateBasic` method, the sign will be always the same

lubtd:
> But don't have too much sense to have the signature flag if we need to always pass the nonce to validate into the `ValidateBasic` method, the sign will be always the same

I believe in fact, `--signature` should be an option that allows to not provide the full validator key file (that contain the private key) but only the public key since the signature has been generated externally

Maybe we can ship this PR without `--signature` and see later the advanced interface with signature manually generated

x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -0,0 +1,65 @@
...
+			sdkerrors.Wrap(types.ErrInvalidValidatorChainID, msg.ChainID)
+	}
+
+	validator := types.Validator{
+		Address:          valPubKey.Address().String(),
+		ConsensusAddress: consAddress,
+		Description:      types.ValidatorDescription{},
+	}
+
+	// get the current validator to eventually overwrite description
```
lubtd, line 46:
```suggestion
	// get the current validator to eventually overwrite description and remove existing consensus key
```

x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -0,0 +1,65 @@
...
+func (k msgServer) SetValidatorConsAddress(
+	goCtx context.Context,
+	msg *types.MsgSetValidatorConsAddress,
+) (*types.MsgSetValidatorConsAddressResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	valPubKey, err := valtypes.NewValidatorConsPubKey(msg.ValidatorConsPubKey, msg.ValidatorKeyType)
+	if err != nil {
+		return &types.MsgSetValidatorConsAddressResponse{},
+			sdkerrors.Wrap(types.ErrInvalidValidatorKey, string(msg.ValidatorConsPubKey))
```
lubtd, line 21: I think it can be a critical error since this check is performed in `Validate` message

lubtd (commented):
👍

x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -0,0 +1,65 @@
...
+	if found {
+		validator.Description = validatorStore.Description
+		k.RemoveValidatorByConsAddress(ctx, validatorStore.ConsensusAddress)
+	}
+
+	// store validator information
+	k.SetValidator(ctx, validator)
+	k.SetValidatorByConsAddress(ctx, types.ValidatorByConsAddress{
+		ConsensusAddress: consAddress,
+		ValidatorAddress: valPubKey.Address().String(),
```
lubtd, line 57:
```suggestion
		ValidatorAddress: msg.ValidatorAddress,
```

x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -0,0 +1,65 @@
...
+		return &types.MsgSetValidatorConsAddressResponse{},
+			sdkerrors.Wrapf(types.ErrInvalidValidatorNonce, "%d", msg.Nonce)
+	}
+	if ctx.ChainID() != msg.ChainID {
+		return &types.MsgSetValidatorConsAddressResponse{},
+			sdkerrors.Wrap(types.ErrInvalidValidatorChainID, msg.ChainID)
+	}
+
+	validator := types.Validator{
+		Address:          valPubKey.Address().String(),
```
lubtd, line 41:
```suggestion
		Address:         msg.ValidatorAddress,
```

lubtd (commented):
Great👍

Just one last comment
When registering the consensus key with another address, the previous validator keep the consensus address in its profile:
```
spnd tx profile set-validator-cons-address priv_validator_key.json 0 --from alice
spnd tx profile set-validator-cons-address priv_validator_key.json 1 --from bob

spnd q profile list-validator
validator:
- address: spn1deu7lhyx7aexggv6a4keqycvukzlelkcj2w55h
  consensusAddress: spnvalcons1twduj8s52ccsscu47e0s2asawd58sln4cpfp4e
- address: spn1t6mrjcm5elren67zpqtqesmhed9ujk4cf9hz47
  consensusAddress: spnvalcons1twduj8s52ccsscu47e0s2asawd58sln4cpfp4e
```

we should in the message check if the consensus key is currently registered with `k.GetValidatorByConsAddress` and eventually remove the cons address of the previous validator

Pantani:
> Great👍
> 
> Just one last comment When registering the consensus key with another address, the previous validator keep the consensus address in its profile:
> 
> ```
> spnd tx profile set-validator-cons-address priv_validator_key.json 0 --from alice
> spnd tx profile set-validator-cons-address priv_validator_key.json 1 --from bob
> 
> spnd q profile list-validator
> validator:
> - address: spn1deu7lhyx7aexggv6a4keqycvukzlelkcj2w55h
>   consensusAddress: spnvalcons1twduj8s52ccsscu47e0s2asawd58sln4cpfp4e
> - address: spn1t6mrjcm5elren67zpqtqesmhed9ujk4cf9hz47
>   consensusAddress: spnvalcons1twduj8s52ccsscu47e0s2asawd58sln4cpfp4e
> ```
> 
> we should in the message check if the consensus key is currently registered with `k.GetValidatorByConsAddress` and eventually remove the cons address of the previous validator

nice catch. I added the logic to remove the old consensus key

lubtd (approved):
👍

PR #473, opened 2022-01-28
feat(launch): Implement `CheckValidatorSet` method

### LINKED ISSUE #436: Implement `CheckValidatorSet`
`CheckValidatorSet` verifies from a `tmtypes.ValidatorSet` https://github.com/tendermint/tendermint/blob/master/types/validator_set.go if the validator set is valid from the information 
stored for the launch of the chain

The method must fail if the launch has not been triggered

Pantani:
Closes #436


x/launch/keeper/validator_set.go
```diff
@@ -0,0 +1,64 @@
...
+	"bytes"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/tendermint/spn/x/launch/types"
+	tmtypes "github.com/tendermint/tendermint/types"
+)
+
+// CheckValidatorSet check the information about a validator
+// set (used to create an IBC client) are valid
```
lubtd, line 13:
```suggestion
// CheckValidatorSet checks the information about a validator
// set (used to create an IBC client) is valid
```

x/launch/keeper/validator_set.go
```diff
@@ -0,0 +1,64 @@
...
+		return false, sdkerrors.Wrap(types.ErrInvalidGenesisChainID, chainID)
+	}
+
+	// all validators must be present in the launch module and
+	// the total amount of self-delegation from the provided validators
+	// must reach at least 2/3 of the total self delegation for the chain
+	valSetSelfDelegation := 0.0
+	for _, validator := range validatorSet.Validators {
+		valAddr := sdk.AccAddress(validator.Address.Bytes())
+		launchValidator, found := k.GetGenesisValidator(ctx, launchID, valAddr.String())
```
lubtd, line 38: I realized we omitted in the specs we have the consensus address and not the validator operator's address. Therefore `valAddr` can't retrieve the validator
Since we only index the validator by the operator address, I think we should with all validators on spn retrieved with `GetAllGenesisValidatorByLaunchID` create an internal map that indexes the consensus pub key of the validator with the full validator or the self delegation to make then this check

x/launch/keeper/validator_set.go
```diff
@@ -0,0 +1,64 @@
...
+		if !found {
+			return false, sdkerrors.Wrap(types.ErrValidatorNotFound, valAddr.String())
+		}
+		valConsPubKey := validator.PubKey.Bytes()
+		if bytes.Compare(launchValidator.ConsPubKey, valConsPubKey) != 0 { // nolint
+			return false, sdkerrors.Wrap(types.ErrInvalidConsPubKey, validator.PubKey.Address().String())
+		}
+		valSetSelfDelegation += float64(launchValidator.SelfDelegation.Amount.Int64())
```
lubtd, line 46: I think we should do the decimal conversion at the last check and keep integer here

x/launch/keeper/validator_set.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	// check if 2/3 of total self-delegation is reached from the provided validator set
+	// GetTotalSelfDelegation is the sum of all self delegation
+	totalSelfDelegation := float64(k.GetTotalSelfDelegation(ctx, launchID))
+	reached := valSetSelfDelegation >= (2.0/3.0)*totalSelfDelegation
```
lubtd, line 52: I would rather use `sdk.Dec` type used for all decimal operations in Cosmos SDK

x/launch/keeper/validator_set.go
```diff
@@ -0,0 +1,64 @@
...
+	return reached, nil
```
lubtd, line 53: I think we could return total self delegation not reached as a error as well

x/launch/keeper/genesis_validator.go
```diff
@@ -54,3 +56,40 @@ func (k Keeper) GetAllGenesisValidator(ctx sdk.Context) (list []types.GenesisVal
...
+
+	defer iterator.Close()
+
+	for ; iterator.Valid(); iterator.Next() {
+		var val types.GenesisValidator
+		k.cdc.MustUnmarshal(iterator.Value(), &val)
+		if bytes.Compare(val.ConsPubKey, consPubKey) == 0 {
+			return val, true
+		}
+	}
```
lubtd, line 93: I think we should have a single method that returns a map `map[conspubkey]GenesisValidator` and the total self delegation so we can do a single iteration of the validator set in `CheckValidatorSet`

x/launch/keeper/validator_set_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch/types"
+	"github.com/tendermint/tendermint/crypto"
+	tmtypes "github.com/tendermint/tendermint/types"
+)
+
+func TestKeeper_CheckValidatorSet(t *testing.T) {
+	var (
+		k, _, _, _, _, _, ctx = setupMsgServer(t)
```
lubtd, line 16: I think we can use `k, ctx := testkeeper.Launch(t)`

x/launch/keeper/genesis_validator.go
lubtd: Mmh what do you think about it then? I was meaning doing a single iteration in validators `for ; iterator.Valid(); iterator.Next()` to get the total self delegation and map all validators by their cons pub key only once

Pantani: ohh sorry, I understand wrong. I agree. I also removed the `GetAllGenesisValidatorByLaunchID` so we can avoid duplicated interaction too

Pantani: done on https://github.com/tendermint/spn/pull/473/commits/f1ea06df25fe74578412ce2554ce6e97d34702aa

lubtd (approved):

PR #474, opened 2022-01-28
chore: upgrade IBC to v2 and switch `spm` to `starport`

lubtd:
Upgrade IBC to v2

Can successfully enable connection between SPN and testnets


lubtd:
We're getting more and more `listen tcp 0.0.0.0:40823: bind: address already in use` in the unit tests which seems to be linked to our numerous cli query tests

I wonder if we should keep these tests in fact since the GRPC queries are already tested in `keeper` and therefore these tests test only the CLI layer command parsing that is generated in a generic way🤔

@Pantani

Pantani (approved):

PR #481, opened 2022-02-01
feat(reward): Scaffold the `RewardPool` map

### LINKED ISSUE #475: Scaffold the `RewardPool` map
The `RewardPool` map must be indexed by the launch ID and it specifies the reward tokens and the instruction for rewards distribution

Pantani:
Closes #475

### What does this PR does?

Scaffold the `RewardPool` map

```go
type RewardPool struct {
	LaunchID            uint64
	Provider            string
	Coins               sdk.Coins
	LastRewardHeight    uint
	CurrentRewardHeight uint
}
```


lubtd (commented):
👍

I think we can add in the PR a `Validate` method for a `RewardPool` that checks:
- Provider is a valid bech32 address
- Coins is a valid `sdk.Coins`
- CurrentRewardHeight < LastRewardHeight

And call `Validate` in GenesisValidate

x/reward/types/reward_pool.go
```diff
@@ -0,0 +1,49 @@
...
+		Provider:            provider,
+		Coins:               coins,
+	}
+}
+
+// Validate check the RewardPool object
+func (m RewardPool) Validate() error {
+	if m.LaunchID == 0 {
+		return errors.New("invalid launch id")
+	}
```
lubtd, line 31:
```suggestion
```

I don't think we need this check since by design a launch ID can be equal to 0 if we set the first ID counter for launch ID to `0`

lubtd:
@Pantani seems to be some unit test error somewhere 👀

lubtd (approved):

Issue #463, opened 2022-01-22
List and remove unnecessary queries


lubtd:
There might be in the codebase, different defined queries generated from Starport but that doesn't make sense to keep since the scaffolded model is mainly used for internal logic

Example: `launchIDFromVerifiedClientID` is used by the IBC handshake logic to check the launch ID related to the client for connection

We should list theses queries remove them along with the `grpc` method in `keeper` to simplify the codebase

lubtd:
Considered for removal:
(All queries are from  `monitoring-consumer` module)

- `LaunchIDFromVerifiedClientID`: used internally during handshake to match which channel ID is concerned by the handshake. From a client perspective `launchID` is always retrieve with the client ID with ProviderClientID query
- `LaunchIDFromVerifiedClientIDAll`: same
- `LaunchIDFromChannelIDAll`: `LaunchIDFromChannelID` can be used to see which chain is connected through a specific channel ID, I don''t see usage for All query since all connections can already be queries with `ProviderClientIDAll` 
- `MonitoringHistoryAll`: don't see any usage of getting the monitoring history from all chains, this can in the future become already extensive information for a single chain once we support monitoring period

Any objection from removing these queries @aljo242 @giunatale @Pantani ?

PR #486, opened 2022-02-03
feat(`monitoringp`): implement `BeginBlock` and report signatures

lubtd:
Implement `BeginBlock` for the module where the validator signatures are reported and IBC packet is eventually transmitted

For testing: follow Monitoring Incentivized Testnets guide ;)


x/monitoringp/module.go
```diff
@@ -168,7 +166,14 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
...
 
 // BeginBlock executes all ABCI BeginBlock logic respective to the capability module.
-func (am AppModule) BeginBlock(_ sdk.Context, _ abci.RequestBeginBlock) {}
+func (am AppModule) BeginBlock(ctx sdk.Context, bb abci.RequestBeginBlock) {
+	// reports signatures for the block
+	am.keeper.ReportBlockSignatures(ctx, bb.LastCommitInfo, bb.Header.Height)
+
+	// check and transmit signatures
+	// TODO(): investigate proper error handling
+	_ = am.keeper.TransmitSignatures(ctx, bb.Header.Height)
```
Pantani, line 175:
```suggestion
	err := am.keeper.TransmitSignatures(ctx, bb.Header.Height)
	if err != nil {
		ctx.Logger().Error(fmt.Sprintf("error to transmit the validator signatures %s", err.Error()))
	}
```

x/monitoringp/keeper/begin_block.go
```diff
@@ -0,0 +1,87 @@
...
+	// check condition to transmit packet
+	// IBC connection to consumer must be established
+	// last block height must be reached
+	// monitoring info must exist
+	// signatures must not yet be transmitted
+	cid, cidFound := k.GetConnectionChannelID(ctx)
+	mi, miFound := k.GetMonitoringInfo(ctx)
+	if blockHeight < k.LastBlockHeight(ctx) || !cidFound || !miFound || mi.Transmitted {
+		return nil
+	}
```
Pantani, line 62: only a suggestion, I think this way is more precise and avoid accessing the store if the  block height is not the right one

```suggestion
	if blockHeight < k.LastBlockHeight(ctx) {
		return nil
	}
	cid, cidFound := k.GetConnectionChannelID(ctx)
	if !cidFound {
		return nil
	}
	mi, miFound := k.GetMonitoringInfo(ctx)
	if !miFound || mi.Transmitted {
		return nil
	}
```

x/monitoringp/keeper/begin_block.go
```diff
@@ -0,0 +1,87 @@
...
+	valSetSize := int64(len(lastCommit.Votes))
+	for _, vote := range lastCommit.Votes {
+		if vote.SignedLastBlock {
+			// TODO(490): implement correct address format
+			monitoringInfo.SignatureCounts.AddSignature(vote.Validator.Address, valSetSize)
+		}
+	}
+
+	// increment block count and save the monitoring info
+	monitoringInfo.SignatureCounts.BlockCount++
```
Pantani, line 46: Maybe we should add some logs here?

x/monitoringc/client/cli/query_monitoring_history.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cast"
+	"github.com/spf13/cobra"
+	"github.com/tendermint/spn/x/monitoringc/types"
+)
+
+func CmdListMonitoringHistory() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list-monitoring-history",
+		Short: "list all monitoringHistory",
```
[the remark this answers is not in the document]
lubtd: This is potentially a query not necessary https://github.com/tendermint/spn/issues/463

x/monitoringp/keeper/begin_block.go
lubtd: makes sense actually

x/monitoringp/module.go
lubtd: That's a good suggestion
Also, FYI we have created a task to investigate the best way to handle error in begin block

x/monitoringp/keeper/begin_block.go
lubtd: We could although IMO log output from each block is already a lot verbose with Cosmso-SDK and the current state of the monitoring info can be queried with `q monitoring-provider show-monitoring-info`

x/monitoringc/client/cli/query.go
```diff
@@ -34,6 +34,10 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 		CmdShowLaunchIDFromChannelID(),
 		CmdQueryParams(),
 	)
+	cmd.AddCommand(
```
Pantani, line 37:
```suggestion
```

Pantani (dismissed):

PR #487, opened 2022-02-04
feat(reward): Implement `MsgSetRewards` message

### LINKED ISSUE #477: Implement `MsgSetReward` message
The `MsgSetReward` message allows a coordinator to set a reward pool for a chain from the provided launch ID

Pantani:
Closes #477

### What does this PR does?

Add the `MsgSetReward` implementation to allow a coordinator to set a reward pool for a chain.

### How to test?

- Create the coordinator and chain
```shell
spnd tx launch create-chain orbit-1 orbit.com 0xaaa --from alice
```

- Create the reward
```shell
spnd tx reward set-rewards 1 1000uspn 100000 --from alice
```

- Try to create a reward token with a no-balance account. It should return an error
```shell
spnd tx reward set-rewards 1 1000token 100000 --from alice
```


x/reward/keeper/msg_server_set_reward.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, spnerrors.Criticalf("can't parse provider address %s", err.Error())
+	}
+
+	rewardPool, found := k.GetRewardPool(ctx, msg.LaunchID)
+	if !found {
+		// create the reward pool and transfer tokens if not created yet
+		if err := k.bankKeeper.SendCoinsFromAccountToModule(ctx, provider, types.ModuleName, msg.Coins); err != nil {
+			return nil, sdkerrors.Wrap(types.ErrInsufficientFunds, err.Error())
+		}
+		rewardPool = types.NewRewardPool(msg.LaunchID, ctx.BlockHeight())
```
lubtd, line 46: It can be a bit confusing but `CurrentRewardHeight` is the height on the testnet where the latest rewards have been distributed (maybe the field should be renamed)

The reason for this field is to be future proof where we can directly have the format to support several distribution rounds for the same reward pool

For the first implementation this field should always stay to `0`

x/reward/keeper/msg_server_set_reward.go
```diff
@@ -0,0 +1,89 @@
...
+				return spnerrors.Criticalf("can't send coin from module to account %s", err.Error())
+			}
+		} else if coin.Amount.GT(balance.Amount) { // the balance is lower than the new value
+			if err := bankKeeper.SendCoinsFromAccountToModule(
+				ctx,
+				provider,
+				types.ModuleName,
+				sdk.NewCoins(coin.Sub(balance)),
+			); err != nil {
+				return spnerrors.Criticalf("can't send coin from account to module %s", err.Error())
```
lubtd, line 84: Should not be a critical error since the user may not have the required balance

x/reward/keeper/msg_server_set_reward.go
```diff
@@ -0,0 +1,89 @@
...
+
+	rewardPool, found := k.GetRewardPool(ctx, msg.LaunchID)
+	if !found {
+		// create the reward pool and transfer tokens if not created yet
+		if err := k.bankKeeper.SendCoinsFromAccountToModule(ctx, provider, types.ModuleName, msg.Coins); err != nil {
+			return nil, sdkerrors.Wrap(types.ErrInsufficientFunds, err.Error())
+		}
+		rewardPool = types.NewRewardPool(msg.LaunchID, ctx.BlockHeight())
+	} else if err := SetBalance(ctx, k.accountKeeper, k.bankKeeper, provider, msg.Coins); err != nil {
+		return nil, spnerrors.Criticalf("can't set balance %s", err.Error())
```
lubtd, line 48: We should have a single `spnerrors.Criticalf` at the top of the call stack. Therefore `SetBalance` should not return a critical error. We should parse here the returned error from `SetBalance` and return a critical error here if it is critical

x/reward/keeper/msg_server_set_reward.go
```diff
@@ -0,0 +1,89 @@
...
+// calling the transfer depending on the balance difference
+func SetBalance(
+	ctx sdk.Context,
+	accountKeeper types.AccountKeeper,
+	bankKeeper types.BankKeeper,
+	provider sdk.AccAddress,
+	coins sdk.Coins,
+) error {
+	for _, coin := range coins {
+		balance := bankKeeper.GetBalance(ctx, accountKeeper.GetModuleAddress(types.ModuleName), coin.Denom)
```
lubtd, line 67: We should not check the balance of the module but the coins in the reward pools since the module account will hold all the reward pool balance

If the module balance is lower than the reward pool coins, it should be considered as a critical error

x/reward/keeper/msg_server_set_reward.go
```diff
@@ -0,0 +1,91 @@
...
+// SetBalance set balance to Coins on the module account
+// calling the transfer depending on the balance difference
+func SetBalance(
+	ctx sdk.Context,
+	bankKeeper types.BankKeeper,
+	provider sdk.AccAddress,
+	coins sdk.Coins,
+	poolCoins sdk.Coins,
+) error {
+	if coins.DenomsSubsetOf(poolCoins) && coins.IsEqual(poolCoins) {
```
lubtd, line 67: Isn't `coins.IsEqual(poolCoins)` sufficient condition?

x/reward/keeper/msg_server_set_reward_test.go
```diff
@@ -0,0 +1,210 @@
...
+	err = bk.MintCoins(sdkCtx, types.ModuleName, append(moduleBalance, newBalance...).Sort())
+	require.NoError(t, err)
+	err = bk.SendCoinsFromModuleToAccount(sdkCtx, types.ModuleName, provider, newBalance)
+	require.NoError(t, err)
+
+	tests := []struct {
+		name string
+		msg  types.MsgSetRewards
+		err  error
+	}{
```
lubtd, line 54: We should have a test case where the message execution is successful

lubtd (commented):
CLI works well👍

x/reward/keeper/msg_server_set_reward.go
Pantani: no, unfortunately, if any denom exists in only one coin list, the `IsEqual ` method will throw a panic

lubtd: Oh, I didn't know this, this is a strange behavior

But then shouldn't we do `poolCoins.DenomsSubsetOf(coins)` as well? Currently here `poolCoins` could contains denom not in `coins` and therefore triggering a panic?

lubtd (dismissed):

x/reward/keeper/msg_server_set_reward.go
Pantani: I think we are subtracting the `poolCoins` from `coins`, we only need to check `coins.DenomsSubsetOf(poolCoins)`. I made some tests here, and I think it's right. I didn't get any issue

Pantani: I added more tests to cover this case

lubtd (dismissed):

PR #488, opened 2022-02-05
feat(reward): Implement `DistributeRewards` method

### LINKED ISSUE #476: Implement `DistributeRewards`
Implement `DistributeRewards` method for distributing rewards from a reward pool from a launch ID to entities listed in the provided signature count

Pantani:
Closes #476

### What does this PR does?

Implement `DistributeRewards` method for distributing rewards.


lubtd (commented):
Great work👍

I think we can try to not use `float` at all and fully use `sdk.Dec`
`TruncateInt` method of dec allows to get the floor of the decimal

Pantani:
> Great work👍
> 
> I think we can try to not use `float` at all and fully use `sdk.Dec` `TruncateInt` method of dec allows to get the floor of the decimal

The `sdk.Dec` don't have division and other operations

x/reward/keeper/reward.go
```diff
@@ -0,0 +1,151 @@
...
+	if !found {
+		return sdkerrors.Wrapf(types.ErrRewardPoolNotFound, "%d", launchID)
+	}
+	provider, err := sdk.AccAddressFromBech32(rewardPool.Provider)
+	if err != nil {
+		return spnerrors.Criticalf("can't parse the provider address %s", err.Error())
+	}
+
+	// only the monitored blocks relative to last reward height are rewarded
+	blockRatio := float64(lastBlockHeight-rewardPool.CurrentRewardHeight) / float64(rewardPool.LastRewardHeight-rewardPool.CurrentRewardHeight)
```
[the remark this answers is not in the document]
Pantani: `sdk.Dec` is too verbose for complex operations and does not have division. Any suggestion for we can solve it?

aljo242:
> 

I think you need to use `sdk.Dec` over `float` because floating point arithmetic could lead to consensus errors.  

All of the methods for it can be found [here](https://github.com/cosmos/cosmos-sdk/blob/master/types/decimal.go).  The division method is `Quo()`.

So, `1.0 / 2.0 -> sdk.Dec(1.0).Quo(sdk.Dec(2.0))`

x/reward/keeper/reward.go
```diff
@@ -0,0 +1,151 @@
...
+		types.ModuleName,
+		provider,
+		rewardPool.Coins); err != nil {
+		return spnerrors.Criticalf("send rewards error: %s", err.Error())
+	}
+
+	// update the current reward height for next reward
+	rewardPool.CurrentRewardHeight = lastBlockHeight
+	k.SetRewardPool(ctx, rewardPool)
+	return nil
```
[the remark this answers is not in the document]
Pantani: I made the change, but I don't think it was cleaner, the `RefundRewards` has too many parameters to pass

Pantani:
> > 
> 
> I think you need to use `sdk.Dec` over `float` because floating point arithmetic could lead to consensus errors.
> 
> All of the methods for it can be found [here](https://github.com/cosmos/cosmos-sdk/blob/master/types/decimal.go). The division method is `Quo()`.
> 
> So, `float(1.0) / float(2.0) -> sdk.Dec(1.0).Quo(sdk.Dec(2.0))`

Nice!! Thanks for the help @aljo242. The `Dec` is not well documented. I thought the `Quo` was to get the division coefficient. Weird. I made some tests here, and it works. Thanks for the help.

And in this case, we don't have a problem using a float because we will truncate to integer the float numbers after the calculation. But it's better to keep the `Dec` by design.

aljo242 (approved):
lgtm !

x/reward/keeper/reward.go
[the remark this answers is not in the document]
lubtd: This is ok
We can eventually see for refacto in the future

x/reward/keeper/reward.go
```diff
@@ -0,0 +1,184 @@
...
+		signatureCounts,
+		totalRelativeSignaturesDistributed,
+		blockRatio,
+		launchID,
+		lastBlockHeight,
+		closeRewardPool,
+	)
+}
+
+func (k Keeper) RefundRewards(
```
lubtd, line 118: I'm ok reverting the change if you consider the method is more complex this way
If we let it like this we should document the method

x/reward/keeper/reward.go
```diff
@@ -0,0 +1,184 @@
...
+		return spnerrors.Criticalf("send rewards error: %s", err.Error())
+	}
+
+	// update the current reward height for next reward
+	rewardPool.CurrentRewardHeight = lastBlockHeight
+	k.SetRewardPool(ctx, rewardPool)
+	return nil
+}
+
+func CalculateReward(blockRatio, ratio sdk.Dec, coins sdk.Coins) (sdk.Coins, error) {
```
lubtd, line 173: Same, need documentation

x/reward/keeper/reward.go
lubtd, line 118: I think the method name can start with a lower-case

x/reward/keeper/reward.go
lubtd, line 173: Same comment with lower-case

x/reward/keeper/reward_test.go
```diff
@@ -0,0 +1,233 @@
...
+		ConsensusAddress: validatorConsAddrBar,
+	})
+	pk.SetValidatorByConsAddress(ctx, profiletypes.ValidatorByConsAddress{
+		ValidatorAddress: sample.Address(),
+		ConsensusAddress: notFoundValidatorAddr,
+	})
+	err := bk.MintCoins(ctx, types.ModuleName, moduleSupply)
+	require.NoError(t, err)
+
+	type args struct {
```
lubtd, line 147: We should assert the coin distribution in the tests
We could for example provide in the input an array `[]Distributed{address: string, coins: string}` and check the balance for accounts
We can eventually initialize basic cases for the PR and handle extensive cases in future PRs since this part is specifically critical security speaking
We can also asser the balance of the provider for refund

x/reward/keeper/reward.go
Pantani: Yeah, I agree to revert, is more complex, and the `RefundRewards` is harder to test, too many args

x/reward/keeper/reward_test.go
Pantani: done one https://github.com/tendermint/spn/pull/488/commits/f5d2c7392e282469e0c61c8d857767d181f30dfb
but we can improve more these test cases later

PR #499, opened 2022-02-08
feat(`monitoringc`): add `CheckValidatorSet` in `MsgCreateClient`

lubtd:
Complete the `MsgCreateClient` message by adding the check for the provided validator set the for the IBC client to verify

The message fails if the check fails

Add tests for the message handler

To test the feature:
- Follow the monitoring guide with `orbit`
- When `orbit` is running, add a genesis validator on `spn` with the consensus key of the validator on `orbit` and trigger the launch to simulate a launch with orbit chain
- `MsgCreateClient` should fail if not the right validator set with the proper consensus key is provided


x/monitoringc/keeper/msg_server_test.go
```diff
@@ -1,16 +1,37 @@
 package keeper_test
 
 import (
-	"context"
+	ibckeeper "github.com/cosmos/ibc-go/v2/modules/core/keeper"
+	launchkeeper "github.com/tendermint/spn/x/launch/keeper"
+	launchtypes "github.com/tendermint/spn/x/launch/types"
+	profilekeeper "github.com/tendermint/spn/x/profile/keeper"
+	profiletypes "github.com/tendermint/spn/x/profile/types"
```
Pantani, line 8: wrong import sort

Pantani (dismissed):

Pantani (approved):

PR #500, opened 2022-02-08
feat(`profile`): delete coordinator -> disable coordinator

### LINKED ISSUE #392: Don't remove coordinator info after `MsgDeleteCoordinator`
`MsgDeleteCoordinator` is used to explicitly declare the coordinator as no longer active. However, we should not remove coordinator attached information for documentation purposes (the information can be retrieved through chain history anyway)

aljo242:
Closes #392 

### What does this PR does?

- Adds `Active` field to `Coordinator` messages
- All coordinators have `Active` set to `true` by default, retaining the previous behavior.
- `MsgDeleteCoordinator` -> `MsgDisableCoordinator` where the `Active` field is set to false instead of deleting the coordinator entry.

### Questions

- Should we remove the unused Keeper functions?
- Currently this only has one-way functionality, meaning Coordinators can only go from Active -> Inactive.  Does it make sense to let users set inactive Coordinators to active?  I don't really think it does.  The main point of this PR is to retain the functionality of deleting the entries (also a one-way operation) without actually removing anything from the store.

## Checklist

review them:

- [X] Include tests
- [ ] Update documentation (*.md) (if needed)


lubtd:
Hey, don't hesitate to ping us for the questions :)

> Should we remove the unused Keeper functions?

We have many unused keeper methods coming from Starport generation> I believe, most `RemoveFoo` are unused. This is something we can eventually investigate at the end of the milestone, when being sure a method will never be used, let's remove it for the sake of codebase simplification
But the general idea is to keep PRs atomic and so keep this one for it's own purpose: prevent removing coordinator information

> Currently this only has one-way functionality, meaning Coordinators can only go from Active -> Inactive. Does it make sense to let users set inactive Coordinators to active? I don't really think it does. The main point of this PR is to retain the functionality of deleting the entries (also a one-way operation) without actually removing anything from the store.

Same as the preceding answer, let's keep this PR atomic. This is a feature we can eventually discuss but I may not think it will be necessary indeed

x/campaign/keeper/msg_server_add_shares.go
```diff
@@ -18,12 +18,19 @@ func (k msgServer) AddShares(goCtx context.Context, msg *types.MsgAddShares) (*t
...
-	// Get the coordinator ID
-	coordinatorID, found := k.profileKeeper.CoordinatorIDFromAddress(ctx, msg.Coordinator)
+	// Get the coordinator ID associated to the sender address
+	coordByAddress, found := k.profileKeeper.GetCoordinatorByAddress(ctx, msg.Coordinator)
 	if !found {
 		return nil, sdkerrors.Wrap(profiletypes.ErrCoordAddressNotFound, msg.Coordinator)
 	}
-	if campaign.CoordinatorID != coordinatorID {
+
+	coord, _ := k.profileKeeper.GetCoordinator(ctx, coordByAddress.CoordinatorID)
```
lubtd, line 27: We must handle the case where the coordinator is not found (while the `CoordinatorByAddress` exists) with a critical error

x/campaign/keeper/msg_server_add_vesting_options.go
```diff
@@ -19,12 +19,19 @@ func (k msgServer) AddVestingOptions(goCtx context.Context, msg *types.MsgAddVes
...
 	if !found {
 		return nil, sdkerrors.Wrap(profiletypes.ErrCoordAddressNotFound, msg.Coordinator)
 	}
-	if campaign.CoordinatorID != coordinatorID {
+
+	coord, _ := k.profileKeeper.GetCoordinator(ctx, coordByAddress.CoordinatorID)
+	if !coord.Active {
+		return nil, sdkerrors.Wrapf(profiletypes.ErrCoordInactive,
+			"coordinator %d inactive", coord.CoordinatorID)
+	}
```
lubtd, line 32: We should have this logic in many place since many operation must be done by the coordinator

I think it could be relevant to implement a `GetActiveCoordinatorByAddress` method for simplicity

Then we can add test cases to this method and we don't need to update current tests for messages since the retrieve validator must be active

lubtd: Or maybe even a simpler implement would be to keep `GetCoordinatorByAddress` and return a critical error inside this method if the retrieve a inactive validator since we should unmatch the address in `CoordinatorByAddress` once the coordinator is disabled

And then we could skip
```
coord, _ := k.profileKeeper.GetCoordinator(ctx, coordByAddress.CoordinatorID)
if !coord.Active {
return nil, sdkerrors.Wrapf(profiletypes.ErrCoordInactive,
	"coordinator %d inactive", coord.CoordinatorID)
}
```

x/profile/keeper/msg_server_disable_coordinator.go
```diff
@@ -2,36 +2,37 @@ package keeper
...
 
 	coord, found := k.GetCoordinator(ctx, coordByAddress.CoordinatorID)
 	if !found {
-		return &types.MsgDeleteCoordinatorResponse{},
+		return &types.MsgDisableCoordinatorResponse{},
 			spnerrors.Criticalf("a coordinator address is associated to a non-existent coordinator ID: %d",
 				coordByAddress.CoordinatorID)
 	}
 
-	k.RemoveCoordinatorByAddress(ctx, msg.Address)
```
lubtd, line 32: We can keep removing this particular structure

The address will remain documented since this is part of `Coordinator` structure
This would allow the user to create a new coordinator

lubtd (commented):
🚀

x/campaign/keeper/msg_server_add_shares.go
aljo242: Addressed with `GetActiveCoordinatorByAddress()`

x/launch/keeper/msg_server_settle_request_test.go
```diff
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	profiletypes "github.com/tendermint/spn/x/profile/types"
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
```
Pantani, line 7:
```suggestion
	"testing"

	sdk "github.com/cosmos/cosmos-sdk/types"
	profiletypes "github.com/tendermint/spn/x/profile/types"
```

x/campaign/keeper/msg_server_add_shares.go
aljo242: This PR keeprs `GetCoordinator` and also adds `GetActiveCoordinatorByAddress` which returns `ErrCoordInactive` if the coordinator is found inactive.

x/launch/keeper/msg_server_settle_request_test.go
aljo242: resolved with format

lubtd:
Hey @aljo242 is it ready for review?

Don't hesitate to ping us with the `Re-request review` button if it's the case

aljo242:
> Hey @aljo242 is it ready for review?
> 
> Don't hesitate to ping us with the `Re-request review` button if it's the case

Got it! Thanks y'all :)

x/profile/keeper/msg_server_update_coordinator_address_test.go
```diff
@@ -62,6 +72,13 @@ func TestMsgUpdateCoordinatorAddress(t *testing.T) {
 				NewAddress: coord1.Address,
 			},
 			err: types.ErrCoordAlreadyExist,
+		}, {
+			name: "inactive coordinator",
+			msg: types.MsgUpdateCoordinatorAddress{
+				Address:    disableCoord.Address,
+				NewAddress: addr2,
+			},
+			err: types.ErrCoordAddressNotFound,
```
lubtd, line 81: It should be `types.ErrCoordInactive`, the fact that the test is passing with a different error means the code is invalid
In this case, it is because `GetCoordinatorByAddress` the inactive validator
So it makes more sense I think to use `GetActiveCoordinatorByAddress `

x/profile/keeper/msg_server_update_coordinator_description.go
```diff
@@ -30,6 +30,13 @@ func (k msgServer) UpdateCoordinatorDescription(
 				coordByAddress.CoordinatorID)
 	}
 
+	// Check if the coordinator is inactive
+	if !coord.Active {
```
lubtd, line 34: Same as above, we can use `GetActiveCoordinatorByAddress`

lubtd (changes requested):
i finally think that `GetActiveCoordinatorByAddress` can be kept as `GetCoordinatorByAddress`
I haven't realise at first by `CoordinatorByAddress` indexation is removed on coordinator disabling. So `GetCoordinatorByAddress` would implicitly means the coordinator is active. And if it is active this is handled with a critical error as you actually implemented

Also, messages currently using `GetActiveCoordinatorByAddress` doesn't need a new test case for inactive coordinator since `GetActiveCoordinatorByAddress` is tested itself as not being able to retrieve inactive coordinator

x/profile/keeper/msg_server_update_coordinator_address_test.go
aljo242: So, as things are, only the `Coordinator` message has the `Active` field.  If we run `GetCoordinatorByAddress`, and we don't find the `CoordinatorByAddress`, then there is no way to then find the `Coordinator` and check if it is active.  

So if this is the case, the only error we can return is `ErrCoordAddressNotFound`.  The Coordinator could have never been created OR could have been disabled and removed.  

In other words, we can't definitively say if the Coordinator is Active if we are trying to access it by the Address (instead of ID).  

If this kind of behavior is not desired, and we want to always be able to say if the Coordinator is Active, Inactive, or Not Created, we will have to no longer remove the `CoordinatorByAddress` entry when disabling.  We can definitely do this, but I want to see your thoughts.  Initially @lubtd suggested we continue removing the `CoordinatorByAddress` entry on removing. 

Thoughts?

x/profile/keeper/msg_server_update_coordinator_address.go
```diff
@@ -37,6 +37,14 @@ func (k msgServer) UpdateCoordinatorAddress(
 			spnerrors.Criticalf("a coordinator address is associated to a non-existent coordinator ID: %d",
 				coordByAddress.CoordinatorID)
 	}
+
+	// Check if the coordinator is inactive
```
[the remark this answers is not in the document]
aljo242: addressed below.  Doing this would introduce extra KVstore accesses

x/profile/keeper/msg_server_update_coordinator_address_test.go
lubtd: Yes, we must keep removing `CoordinatorByAddress` on disable because it enable the user to create a new coordinator, this is the right behavior.

As mentioned in the general comment in my review, we can remove test case that assert solely on inactive coordinator for methods using `GetCoordinatorByAddress`/`GetActiveCoordinatorByAddress` because `GetCoordinatorByAddress`/`GetActiveCoordinatorByAddress` is already tested for not retrieving inactive coordinator so a single test case when the coordinator is not found is sufficient

x/profile/keeper/msg_server_update_coordinator_address_test.go
```diff
@@ -12,18 +12,28 @@ import (
...
 	}
 	if _, err := srv.CreateCoordinator(wCtx, &coord2); err != nil {
 		t.Fatal(err)
 	}
+	if _, err := srv.CreateCoordinator(wCtx, &disableCoord); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := srv.DisableCoordinator(wCtx, &disableMsg); err != nil {
+		t.Fatal(err)
+	}
```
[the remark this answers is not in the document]
lubtd: Mmh any concern related to this comment @aljo242 ?

x/profile/keeper/msg_server_update_coordinator_description.go
lubtd: I got some confusion between `getCoordinatorByAddress` and `GetCoordinatorByAddress`
Yes, this is a good point. Although, I believe having an extra read in the store is not a big concern if it's related to a single operation, as long as it doesn't become more than O(n) complexity

So we have `GetCoordinatorByAddress` used by external module where the invariant is checked and internal method using `getCoordinatorByAddress` where the invariant is checked internally because `Coordinator` mst be fetched as well

lubtd: We can keep this behaviour although in any case, we must return a Critical error if an inactive coordinator has been fetched with `getCoordinatorByAddress`, so in this particular condition, a critical error must be returned

x/profile/keeper/msg_server_disable_coordinator_test.go
```diff
@@ -22,32 +22,34 @@ func TestMsgDeleteCoordinator(t *testing.T) {
 	}
 	tests := []struct {
 		name string
-		msg  types.MsgDeleteCoordinator
+		msg  types.MsgDisableCoordinator
 		err  error
 	}{
 		{
```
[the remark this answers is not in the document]
aljo242: The behavior will be the same as if the coordinatorByAddr does not exist (`ErrCoordAddrNotFound`), since we perform this Tx by Address.

lubtd (approved):

PR #512, opened 2022-02-11
feat(profile): Turn `consensusAddress` field of validator profile into an array

### LINKED ISSUE #480: Allowing associating several consensus addr to the same validator profile
Currently, validator address on `spn` and consensus address of validator for launched testnet is a 1-1 association

A consensusaddress must be associated to a single validator profile for reward distribution, but a validator profile could have several consensus address to use the same validator on `spn` for different testnets

Changes:

Turn `consensusAddress` field of validator profile into an array
The message `MsgSetValidatorConsAddress` append the new consensus address in the array and eventually remove it from another existing validator profile
The signature function with on-chain nonce would remain the same

Pantani:
Closes #480

### What does this PR does?

Allow associating several consensus address to the same validator profile


x/profile/keeper/msg_server_set_validator_cons_address.go
```diff
@@ -39,9 +39,9 @@ func (k msgServer) SetValidatorConsAddress(
...
 	validator := types.Validator{
-		Address:          msg.ValidatorAddress,
-		ConsensusAddress: consAddress,
-		Description:      types.ValidatorDescription{},
+		Address:            msg.ValidatorAddress,
+		ConsensusAddresses: [][]byte{consAddress},
+		Description:        types.ValidatorDescription{},
 	}
 
 	if valByConsAddr, found := k.GetValidatorByConsAddress(ctx, consAddress); found {
```
lubtd, line 47: Can we add a comment: remove the consensus key from previous address

lubtd (commented):
The CLI works perfectly👍

In the genesis validate, there is something we need to add:

We currently check for each `validatorByConsAddress` if an entry in `validator` exists: https://github.com/tendermint/spn/blob/de22e6fa6c0facea27ecc95818c899baec3df696/x/profile/types/genesis.go#L52

We should also check that this validator contains the consensus address in its consAddr array because if it not the case we may end up with an inconsistent state

lubtd (approved):

aljo242 (approved):

Issue #502, opened 2022-02-09
Analyse and improve `DistributeRewards` tests


lumtis:
Depends on https://github.com/tendermint/spn/pull/488

As a critical part of the protocol, the reward distribution logic must be tested with extensive test cases
We should analyse the current set of test, define every possible edge case for the system and eventually improve the tests

lubtd:
Depends on https://github.com/tendermint/spn/issues/552

PR #520, opened 2022-02-14
fix: reward distribution not balanced

lubtd:
- Perform some fix in the reward distribution algorithm
- Simplify and improve tests. Partially but not fully solve https://github.com/tendermint/spn/issues/502 some assertions and some test cases must be added


lubtd:
@Pantani `reward_test.go` is showed as deleted but it has been actually renamed into `reward_distribution_test.go` and the tests have been a bit changed. Tell me if you have any comments on those changes.
I think for the test cases it's easier to use multiples of 10 for more readable assertion 50% of 10 -> 5

x/reward/keeper/reward_distribution.go
```diff
@@ -120,25 +119,30 @@ func (k Keeper) DistributeRewards(
 	// Otherwise, the refund is relative to the block ratio and the reward pool is updated
 	// refundRation is blockCount.
```
aljo242, line 120:
```suggestion
	// refundRatio is blockCount.
```

a bit more clear

aljo242 (changes requested):

x/reward/keeper/msg_server_test.go
```diff
@@ -36,3 +38,33 @@ func setupMsgServer(t testing.TB) (
 		launchkeeper.NewMsgServerImpl(*launchKeeper),
 		ctx
 }
+
+// coinsFromString returns a sdk.Coins from a string
+func coinsFromString(t testing.TB, str string) sdk.Coins {
```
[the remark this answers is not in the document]
lubtd: This makes sense, we already have similar constructor for signature counts type tests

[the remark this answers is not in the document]
lubtd: I think we can have a `constructor` package in `testutil` for simple constructor for test purpose

lubtd:
@giunatale please check this one when you have time

testutil/constructor/constructor.go
```diff
@@ -0,0 +1,41 @@
...
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+
+	spntypes "github.com/tendermint/spn/pkg/types"
+)
+
+// Coins returns a sdk.Coins from a string
+func Coins(t testing.TB, str string) sdk.Coins {
```
aljo242, line 14: Is there any reason these (and some other of our testing functions) receive a `testing.TB` rather than a `*testing.T`?

My understanding is that the `TB` interface is for both testing and benchmarking, but I'm not sure how that functionality is being leveraged here.

lubtd: Well spotted, I added `TB` below as well so that the methods can be eventually used in Benchmarking test as well

aljo242 (approved):

giunatale (approved):
lgtm

PR #523, opened 2022-02-16
feat(`launch`): revert delay param

### LINKED ISSUE #497: Create param `RevertDelay`
Create the param `RevertDelay` that defines the time after the actual launch of a chain when a chain launch can be reverted if something went wrong

This time should be by default sufficient to let the time to enable the monitoring IBC connection

aljo242:
Closes #497

### What does this PR does?

Adds param `RevertDelay` to the `x/launch` module.  Uses this new param in `x/launch/msg_server_revert_launch.go` instead of the previous hard-coded value.

### How to test?

New tests are added for param validation.  `x/launch/msg_server_revert_launch.go` functionality is covered by its existing tests.

## Checklist

review them:

- [X] Include tests


proto/launch/params.proto
```diff
@@ -11,4 +11,5 @@ message Params {
 
   uint64 minLaunchTime = 1 [(gogoproto.moretags) = "yaml:\"min_launch_time\""];
   uint64 maxLaunchTime = 2 [(gogoproto.moretags) = "yaml:\"max_launch_time\""];
+  uint64 revertDelay = 3   [(gogoproto.moretags) = "yaml:\"revert_delay\""];
```
lubtd, line 14: I think we can use `int64` instead since the type is manipulated with UNIX timestamp, we can easily ensure the value is positive from the validate method while `int64` conversion comes with some vulnerabilities

If we go this way, we will do the change as well for `minLaunchTime` and `maxLaunchTime`

lubtd (commented):
👍
Some small comments

proto/launch/params.proto
aljo242: Ok, I'll add a check that values are greater than or equal to zero as well.

x/launch/keeper/msg_server_revert_launch_test.go
```diff
@@ -41,7 +41,7 @@ func TestMsgRevertLaunch(t *testing.T) {
 	chain, found := k.GetChain(sdkCtx, delayNotReached)
 	require.True(t, found)
 	chain.LaunchTriggered = true
-	chain.LaunchTimestamp = testkeeper.ExampleTimestamp.Unix() - types.RevertDelay + 1
+	chain.LaunchTimestamp = testkeeper.ExampleTimestamp.Unix() - int64(k.RevertDelay(sdkCtx)) + 1
```
Pantani, line 44:
```suggestion
	chain.LaunchTimestamp = testkeeper.ExampleTimestamp.Unix() - k.RevertDelay(sdkCtx) + 1
```

x/launch/keeper/msg_server_revert_launch_test.go
```diff
@@ -50,7 +50,7 @@ func TestMsgRevertLaunch(t *testing.T) {
 	chain, found = k.GetChain(sdkCtx, delayReached)
 	require.True(t, found)
 	chain.LaunchTriggered = true
-	chain.LaunchTimestamp = testkeeper.ExampleTimestamp.Unix() - types.RevertDelay
+	chain.LaunchTimestamp = testkeeper.ExampleTimestamp.Unix() - int64(k.RevertDelay(sdkCtx))
```
Pantani, line 53:
```suggestion
	chain.LaunchTimestamp = testkeeper.ExampleTimestamp.Unix() - k.RevertDelay(sdkCtx)
```

lubtd (approved):
Perfect 👍

I would also suggest to set a small value in `config.yml` param so chain launches can be reverted for test purposes:

```
    launch:
      params:
        revertDelay: "5"
        minLaunchTime: "5"
```

lubtd (approved):

Pantani (approved):

PR #524, opened 2022-02-16
feat(`campaign`): create min and max supply params

### LINKED ISSUE #493: Create params `MinSupply` and `MaxSupply`
We currently prevent when send `MsgUpdateTotalSupply` removing an existing denomination from the `TotalSupply` of the campaign otherwise the coordinator has the ability to remove value from already minted vouchers. The existing denom can only be set to a new value or new denoms can be added in the total supply

However, the value of the supply for a denom has no range and can be set to `1`. We should create params to set of reasonable range for any use case for the minimum value and the maximum value of a total supply

giunatale:
Closes #493. 

### What does this PR does?

Adds a `TotalSupplyRange` param to the `campaign` module. This param is a struct comprised of two `sdk.Int` values `MinTotalSupply` and `MaxTotalSupply` that define the range for a valid total supply.

## Checklist

review them:

- [x] Include tests

### DIFF
--- a/testutil/constructor/constructor.go
+++ b/testutil/constructor/constructor.go
@@ -4,6 +4,8 @@ package constructor
 import (
 	"testing"
 
+	campaigntypes "github.com/tendermint/spn/x/campaign/types"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/stretchr/testify/require"
 
@@ -38,4 +40,21 @@ func SignatureCounts(blockCount uint64, sc ...spntypes.SignatureCount) spntypes.
 		BlockCount: blockCount,
 		Counts:     sc,
 	}
+}
+
+// Shares returns a Shares object from a string of coin inputs
+func Shares(t testing.TB, coinStr string) campaigntypes.Shares {
+	shares := campaigntypes.NewSharesFromCoins(Coins(t, coinStr))
+	return shares
+}
+
+// Vouchers returns a Vouchers object from a string of coin inputs
+func Vouchers(t testing.TB, coinStr string, campaignID uint64) sdk.Coins {
+	coins := Coins(t, coinStr)
+	vouchers := make(sdk.Coins, len(coins))
+	for i, coin := range coins {
+		coin.Denom = campaigntypes.VoucherDenom(campaignID, coin.Denom)
+		vouchers[i] = coin
+	}
+	return vouchers
 }
--- a/x/campaign/types/message_update_total_shares_test.go
+++ b/x/campaign/types/message_update_total_shares_test.go
@@ -3,7 +3,8 @@ package types_test
 import (
 	"testing"
 
-	sdk "github.com/cosmos/cosmos-sdk/types"
+	tc "github.com/tendermint/spn/testutil/constructor"
+
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/stretchr/testify/require"
 
@@ -57,9 +58,7 @@ func TestMsgUpdateTotalShares_ValidateBasic(t *testing.T) {
 			msg: types.MsgUpdateTotalShares{
 				Coordinator: sample.Address(),
 				CampaignID:  0,
-				TotalShares: types.Shares(sdk.NewCoins(
-					sdk.NewCoin("foo", sdk.NewInt(100)),
-				)),
+				TotalShares: types.Shares(tc.Coins(t, "100foo")),
 			},
 			err: types.ErrInvalidShares,
 		},
--- a/x/campaign/types/share_vesting_options_test.go
+++ b/x/campaign/types/share_vesting_options_test.go
@@ -4,6 +4,8 @@ import (
 	"testing"
 	"time"
 
+	tc "github.com/tendermint/spn/testutil/constructor"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/stretchr/testify/require"
 
@@ -11,15 +13,9 @@ import (
 	"github.com/tendermint/spn/x/campaign/types"
 )
 
-func sharesStr(t *testing.T, str string) types.Shares {
-	shares, err := types.NewShares(str)
-	require.NoError(t, err)
-	return shares
-}
-
 func TestNewDelayedVesting(t *testing.T) {
-	totalShares := sharesStr(t, "1000foo,1000bar,500toto")
-	vesting := sharesStr(t, "1000foo,500bar")
+	totalShares := tc.Shares(t, "1000foo,1000bar,500toto")
+	vesting := tc.Shares(t, "1000foo,500bar")
 	endTime := time.Now().Unix()
 
 	vestingOptions := types.NewShareDelayedVesting(totalShares, vesting, endTime)
@@ -31,8 +27,8 @@ func TestNewDelayedVesting(t *testing.T) {
 }
 
 func TestDelayedVesting_Validate(t *testing.T) {
-	totalShares := sharesStr(t, "1000foo,1000bar,500toto")
-	vesting := sharesStr(t, "1000foo,500bar")
+	totalShares := tc.Shares(t, "1000foo,1000bar,500toto")
+	vesting := tc.Shares(t, "1000foo,500bar")
 
 	tests := []struct {
 		name   string
@@ -87,8 +83,8 @@ func TestDelayedVesting_Validate(t *testing.T) {
 		{
 			name: "total shares smaller than vesting",
 			option: *types.NewShareDelayedVesting(
-				sharesStr(t, "1000foo,500bar,2000toto"),
-				sharesStr(t, "1000foo,501bar,2000toto"),
+				tc.Shares(t, "1000foo,500bar,2000toto"),
+				tc.Shares(t, "1000foo,501bar,2000toto"),
 				time.Now().Unix(),
 			),
 			valid: false,
@@ -96,8 +92,8 @@ func TestDelayedVesting_Validate(t *testing.T) {
 		{
 			name: "vesting denoms is not a subset of total shares",
 			option: *types.NewShareDelayedVesting(
-				sharesStr(t, "1000foo,500bar"),
-				sharesStr(t, "1000foo,500bar,2000toto"),
+				tc.Shares(t, "1000foo,500bar"),
+				tc.Shares(t, "1000foo,500bar,2000toto"),
 				time.Now().Unix(),
 			),
 			valid: false,
--- a/x/campaign/types/shares_test.go
+++ b/x/campaign/types/shares_test.go
@@ -1,7 +1,10 @@
 package types_test
 
 import (
+	"fmt"
 	"testing"
+
+	tc "github.com/tendermint/spn/testutil/constructor"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/stretchr/testify/require"
@@ -26,10 +29,7 @@ func TestNewShares(t *testing.T) {
 
 	shares, err := campaign.NewShares("100foo,200bar")
 	require.NoError(t, err)
-	require.Equal(t, shares, campaign.Shares(sdk.NewCoins(
-		sdk.NewCoin(prefixedShareFoo, sdk.NewInt(100)),
-		sdk.NewCoin(prefixedShareBar, sdk.NewInt(200)),
-	)))
+	require.Equal(t, shares, tc.Shares(t, "100foo,200bar"))
 }
 
 func TestNewSharesFromCoins(t *testing.T) {
@@ -37,20 +37,14 @@ func TestNewSharesFromCoins(t *testing.T) {
 		sdk.NewCoin("foo", sdk.NewInt(100)),
 		sdk.NewCoin("bar", sdk.NewInt(200)),
 	))
-	require.Equal(t, shares, campaign.Shares(sdk.NewCoins(
-		sdk.NewCoin(prefixedShareFoo, sdk.NewInt(100)),
-		sdk.NewCoin(prefixedShareBar, sdk.NewInt(200)),
-	)))
+	require.Equal(t, shares, tc.Shares(t, "100foo,200bar"))
 }
 
 func TestCheckShares(t *testing.T) {
-	require.NoError(t, campaign.CheckShares(campaign.Shares(sdk.NewCoins(
-		sdk.NewCoin(prefixedShareFoo, sdk.NewInt(100)),
-		sdk.NewCoin(prefixedShareBar, sdk.NewInt(200)),
-	))))
+	require.NoError(t, campaign.CheckShares(tc.Shares(t, "100foo,200bar")))
 	require.Error(t, campaign.CheckShares(campaign.Shares(sdk.NewCoins(
 		sdk.NewCoin("foo", sdk.NewInt(100)),
-		sdk.NewCoin(prefixedShareBar, sdk.NewInt(200)),
+		sdk.NewCoin("s/bar", sdk.NewInt(200)),
 	))))
 }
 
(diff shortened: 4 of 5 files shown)


x/campaign/types/params_test.go
```diff
@@ -0,0 +1,87 @@
...
+		{
+			name:   "invalid range",
+			params: NewParams(DefaultMaxTotalSupply, DefaultMinTotalSupply),
+			err:    errors.New("maximum total supply should be greater than greater or equal than minimum total supply"),
+		},
+		{
+			name:   "valid range",
+			params: NewParams(DefaultMinTotalSupply, DefaultMaxTotalSupply),
+		},
+	}
```
aljo242, line 27:
```suggestion
		{
			name:   "invalid min total supply",
			params: NewParams(sdk.ZeroInt(), DefaultMaxTotalSupply),
		},
	
	}
```

x/campaign/types/total_supply.go
```diff
@@ -21,3 +25,34 @@ func UpdateTotalSupply(coins, updatedCoins sdk.Coins) sdk.Coins {
 
 	return append(updatedCoins, notUpdated...).Sort()
 }
+
+// ValidateTotalSupply checks whether the total supply for each denom is within the provided total supply range
+func ValidateTotalSupply(coins sdk.Coins, supplyRange TotalSupplyRange) error {
```
aljo242, line 30:
```suggestion
func ValidateCoinSupplies(coins sdk.Coins, supplyRange TotalSupplyRange) error {
```

I think this name is a bit clearer

aljo242 (changes requested):

x/campaign/types/total_supply.go
lubtd: `TotalSupply` is our current term in the official terminology of SPN to designate the supply of tokens of a campaign's mainnet. If you have idea for renaming this should be further discussed in a call because there are many places where it would change
I believe Total Supply is a good term because it is very used in the ecosystem

proto/campaign/params.proto
```diff
@@ -0,0 +1,25 @@
...
+    (gogoproto.nullable)   = false
+  ];
+}
+
+// Params defines the set of params for the campaign module.
+message Params {
+  option (gogoproto.goproto_stringer) = false;
+
+  TotalSupplyRange totalSupplyRange = 1 [(gogoproto.nullable) = false];
+}
```
lubtd, line 25: We should put `Params` definition first in the file

testutil/sample/campaign.go
```diff
@@ -77,10 +77,25 @@ func MainnetAccount(campaignID uint64, address string) campaign.MainnetAccount {
 
 // MsgCreateCampaign returns a sample MsgCreateCampaign
 func MsgCreateCampaign(coordAddr string) campaign.MsgCreateCampaign {
+	coins := Coins()
+	for i := range coins {
+		coins[i].Amount = coins[i].Amount.Add(campaign.DefaultMinTotalSupply)
+	}
```
lubtd, line 83: I think we can introduce a `TotalSupply()` method in `sample` that initializes a `Coins` with token in the total supply range

x/campaign/keeper/keeper.go
```diff
@@ -3,6 +3,8 @@ package keeper
 import (
 	"fmt"
 
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
```
lubtd, line 6: We should normalize the import order
```
"fmt"

"github.com/cosmos/cosmos-sdk/codec"
paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
"github.com/tendermint/tendermint/libs/log"

```

x/campaign/types/genesis_test.go
```diff
@@ -306,3 +309,43 @@ func TestGenesisState_Validate(t *testing.T) {
 		})
 	}
 }
+
+func TestGenesisState_ValidateParams(t *testing.T) {
```
lubtd, line 313: Params validaty test cases should be complete in the `types/params_test.go`. For genesis state we assume `params.ValdiateBasics()` is fully tested and therefore we should just test once that the genesis state is invalid if the params are invalid

x/campaign/types/params.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// ValidateBasic performs basic validation on campaign parameters.
+func (p Params) ValidateBasic() error {
+	if p.TotalSupplyRange.MinTotalSupply.LT(sdk.OneInt()) {
+		return errors.New("minimum total supply should be greater than one")
+	}
+	if p.TotalSupplyRange.MaxTotalSupply.LT(p.TotalSupplyRange.MinTotalSupply) {
+		return errors.New("maximum total supply should be greater than greater or equal than minimum total supply")
+	}
```
lubtd, line 65: Mmh I think we can simply use
```
return validateTotalSupplyRange(p.TotalSupplyRange)
```

x/campaign/types/params_test.go
```diff
@@ -0,0 +1,87 @@
...
+			},
+			err: errors.New("parameter minTotalSupply cannot be less than one"),
+		},
+		{
+			name: "invalid range",
+			supplyRange: TotalSupplyRange{
+				MinTotalSupply: DefaultMaxTotalSupply,
+				MaxTotalSupply: DefaultMinTotalSupply,
+			},
+			err: errors.New("parameter maxTotalSupply cannot be less than minTotalSupply"),
```
lubtd, line 66: They're actually equal in this case, and I don't think we should prevent this situation, even though it shouldn't be set like this in practice, this means there is a single total supply value that can be set of all coin

x/campaign/types/total_supply.go
lubtd, line 30:
```suggestion
func ValidateTotalSupply(totalSupply sdk.Coins, supplyRange TotalSupplyRange) error {
```

x/campaign/types/total_supply.go
```diff
@@ -21,3 +25,34 @@ func UpdateTotalSupply(coins, updatedCoins sdk.Coins) sdk.Coins {
...
+			supplyRange.MinTotalSupply.String(),
+			supplyRange.MaxTotalSupply.String(),
+		)
+	}
+
+	for _, coin := range coins {
+		totalSupply := coin.Amount
+		if totalSupply.LT(supplyRange.MinTotalSupply) {
+			return fmt.Errorf(
+				"total supply for %s cannot be less than minTotalSupply %s: %s",
```
lubtd, line 44:
```suggestion
				"total supply for %s cannot be lower than minTotalSupply %s: %s",
```

x/campaign/types/total_supply_test.go
```diff
@@ -60,3 +60,47 @@ func TestUpdateTotalSupply(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateTotalSupply(t *testing.T) {
+	newCoins := func(coinsStr string) sdk.Coins {
```
lubtd, line 65: FYI, https://github.com/tendermint/spn/pull/520 will introduce a package `testutil/constructor` to centralized methods for this type of initialization

We can see depending on the first PR merged

x/campaign/types/total_supply.go
```diff
@@ -21,3 +25,34 @@ func UpdateTotalSupply(coins, updatedCoins sdk.Coins) sdk.Coins {
 
 	return append(updatedCoins, notUpdated...).Sort()
 }
+
+// ValidateTotalSupply checks whether the total supply for each denom is within the provided total supply range
+func ValidateTotalSupply(coins sdk.Coins, supplyRange TotalSupplyRange) error {
+	// extra safety check, this should never happen
+	if supplyRange.MaxTotalSupply.LT(supplyRange.MinTotalSupply) {
```
lubtd, line 32: I think it would make sense to introduce a `func (TotalSupplyRange) Validate() error` method so the check is always performed from a single entrypoint
Then the param validate calls this method

x/campaign/types/total_supply.go
```diff
@@ -21,3 +25,34 @@ func UpdateTotalSupply(coins, updatedCoins sdk.Coins) sdk.Coins {
 
 	return append(updatedCoins, notUpdated...).Sort()
 }
+
+// ValidateTotalSupply checks whether the total supply for each denom is within the provided total supply range
+func ValidateTotalSupply(coins sdk.Coins, supplyRange TotalSupplyRange) error {
+	// extra safety check, this should never happen
```
lubtd, line 31: I think it's actually a good practive to include these checks
In this case since it is an inconsistent state error, this should be handled by a `CriticalError`

Since the package `types` is stateless and unopinionated from `CriticalError`, this method should return a specific error type at this stage `ErrInvalidSupplyrange` then the calling methods in `keeper` should return themselves a `CriticalError` if `ErrInvalidSupplyrange` occurs

x/campaign/keeper/msg_server_create_campaign.go
```diff
@@ -19,6 +19,12 @@ func (k msgServer) CreateCampaign(goCtx context.Context, msg *types.MsgCreateCam
 		return nil, sdkerrors.Wrap(profiletypes.ErrCoordAddressNotFound, msg.Coordinator)
 	}
 
+	// Validate provided totalSupply
+	totalSupplyRange := k.TotalSupplyRange(ctx)
+	if err := types.ValidateTotalSupply(msg.TotalSupply, totalSupplyRange); err != nil {
+		return nil, sdkerrors.Wrap(types.ErrInvalidTotalSupply, err.Error())
+	}
```
lubtd, line 26: We must introduce a test case in message test for this condition

x/campaign/keeper/msg_server_update_total_supply.go
```diff
@@ -35,6 +35,12 @@ func (k msgServer) UpdateTotalSupply(goCtx context.Context, msg *types.MsgUpdate
 		return nil, sdkerrors.Wrapf(types.ErrMainnetInitialized, "%d", msg.CampaignID)
 	}
 
+	// Validate provided totalSupply
+	totalSupplyRange := k.TotalSupplyRange(ctx)
+	if err := types.ValidateTotalSupply(msg.TotalSupplyUpdate, totalSupplyRange); err != nil {
+		return nil, sdkerrors.Wrap(types.ErrInvalidTotalSupply, err.Error())
+	}
```
lubtd, line 42: We must introduce a test case in message test for this condition

lubtd (changes requested):

x/campaign/types/total_supply.go
giunatale: Agreed with @lubtd, that was my intention when writing the function. I'd keep it as it is tbh

x/campaign/types/params_test.go
giunatale: They aren't equal (`DefaultMinTotalSupply` used for max and `DefaultMaxTotalSupply` used for min hence max < min (?) ), and the situation in which the two range boundaries are equal is allowed.

x/campaign/types/genesis_test.go
giunatale: I see `lanch/types/genesis_test.go` and I should just remove the last test case (the valid one) to make it consistent with `TestGenesisState_ValidateParams`. Is this what you are asking?

x/campaign/types/total_supply.go
giunatale: This is arbitrary, it's american vs british english imho (but I don't think we have chosen one or the other for style afaik), but fine I'll change it

x/campaign/keeper/msg_server_update_total_supply.go
giunatale: right, I was thinking about it yesterday when I logged off, completely forgot to add test cases for this check in both messages. Will do

testutil/sample/campaign.go
giunatale: I' also adding a `CoinWithRange(min, max)` and `CoinsWithRange(min, max)` to `sample` since I need it for a test and perhaps having sample functions with boundaries for amounts might be useful for other things

x/campaign/types/params_test.go
giunatale: added

x/campaign/keeper/msg_server_update_total_supply.go
```diff
@@ -35,6 +37,15 @@ func (k msgServer) UpdateTotalSupply(goCtx context.Context, msg *types.MsgUpdate
 		return nil, sdkerrors.Wrapf(types.ErrMainnetInitialized, "%d", msg.CampaignID)
 	}
 
+	// Validate provided totalSupply
+	totalSupplyRange := k.TotalSupplyRange(ctx)
+	if err := types.ValidateTotalSupply(msg.TotalSupplyUpdate, totalSupplyRange); err != nil {
+		if errors.Is(err, types.ErrInvalidSupplyRange) {
```
giunatale, line 43: This branch here is not covered from tests (same in `msg_create_campaign`) since it would require to create a state where the range params are invalid. I assumed this would be fine

lubtd: Yes, critical errors are generally (except some exception) not tested simply because can't create a state where those are triggered unless we do some hacks in the store which is not in the idea of message tests

lubtd:
@giunatale don't hesitate to resolve the conversions when you consider them as resolved

x/campaign/types/genesis_test.go
lubtd: I'm thinking about two test cases:
- Genesis state is valid
- Genesis state is invalid because params are invalid (whatever the reason is)

x/campaign/keeper/msg_server_update_total_supply_test.go
```diff
@@ -99,6 +99,15 @@ func TestMsgUpdateTotalSupply(t *testing.T) {
 			},
 			err: types.ErrMainnetInitialized,
 		},
+		{
```
lubtd, line 102: Same comment as above

Maybe it can be worth for check all places where `TotalSupply` is assigned for test purposes and use `sample.TotalSupply()`

x/campaign/types/total_supply.go
```diff
@@ -21,3 +26,42 @@ func UpdateTotalSupply(coins, updatedCoins sdk.Coins) sdk.Coins {
...
+				coin.Denom,
+				supplyRange.MaxTotalSupply.String(),
+				coin.Amount.String())
+		}
+	}
+	return nil
+}
+
+// Validate performs basic validation on an instance of TotalSupplyRange
+func (sr TotalSupplyRange) Validate() error {
```
lubtd, line 57: We should also have `MinTotalSupply` being positive in the check

x/campaign/types/params_test.go
lubtd: Sorry I misread

x/campaign/types/total_supply.go
giunatale: Since it became identical to `validateTotalSupplyRange()` I replaced it with a call to that function

lubtd: There is maybe the interface conversion in addition in `validateTotalSupplyRange`

giunatale: perhaps we could do the other way around? (use this function in `validateTotalSupplyRange`?)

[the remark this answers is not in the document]
giunatale: To avoid duplication I ended up calling `TotalSupplyRange.ValidateBasic()` inside the param `validateTotalSupplyRange`). Thus, I've added test cases for  `TotalSupplyRange.ValidateBasic()` that cover all cases while in `validateTotalSupplyRange` I'm just testing: wrong interface, wrong param, valid param.

x/campaign/types/total_supply_test.go
lubtd: Can be used now :)

lubtd:
Hey @giunatale, is it ready for review? Don't hesitate to push `Re-request review` button when it's the case ;)

x/campaign/types/genesis_test.go
lubtd: Yes good point, we added more test case than necessary. I believe we can change this in https://github.com/tendermint/spn/issues/525 if we can consider this issue as a general refactoring of the `launch param`

x/campaign/types/total_supply_test.go
giunatale: rebased on `develop` and used `tc.Coins()` also here :)

lubtd (changes requested):
Everything looks LGTM to me. We should just add a `Params` query as we have for other modules

giunatale:
I missed that. makes sense

x/campaign/types/params_test.go
```diff
@@ -0,0 +1,84 @@
...
+	tests := []struct {
+		name        string
+		supplyRange interface{}
+		err         error
+	}{
+		{
+			name:        "invalid interface",
+			supplyRange: "test",
+			err:         fmt.Errorf("invalid parameter type: string"),
+		},
```
Pantani, line 56: we should test the `MinTotalSupply` lower than one too

giunatale: we have decided to not fully test validation except where the core functionality is implemented. In all methods using that, we just have a single valid and invalid tests. @lubtd can confirm

x/campaign/types/genesis_test.go
```diff
@@ -306,3 +309,36 @@ func TestGenesisState_Validate(t *testing.T) {
...
+		genState      types.GenesisState
+		shouldBeValid bool
+	}{
+		{
+			desc: "max total supply below min total supply",
+			genState: types.GenesisState{
+				Params: types.NewParams(types.DefaultMinTotalSupply, types.DefaultMinTotalSupply.Sub(sdk.OneInt())),
+			},
+			shouldBeValid: false,
+		},
```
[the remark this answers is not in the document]
giunatale: same as above

Pantani (approved):

lubtd (approved):

PR #527, opened 2022-02-17
chore: set up environment for local testing with several nodes

lubtd:
Set up an environment with 3 validators under `localnet` for local testing.

For usage, read: https://github.com/tendermint/spn/blob/chore/localnet/localnet/readme.md


lubtd:
Can you try building the chain?
```
starport chain build
```

We will have to update `genesis_template.json` but let wait https://github.com/tendermint/spn/pull/523 to be merged

aljo242:
Yep, this was because I was using `spnd` from the wrong branch - sorry about that!

aljo242:
If I run `python3 undelegate.py 10000000 0 0`, this will remove all tokens from node1 and the localnet will no longer work.  Do we want add some kind of check or do we think this is fine.  If people fully undelegate from the main node, perhaps that's just on them.

aljo242:
> Can you try building the chain?
> 
> ```
> starport chain build
> ```
> 
> We will have to update `genesis_template.json` but let wait #523 to be merged

Now that it is merged, a param entry will need to be added

lubtd:
> > Can you try building the chain?
> > ```
> > starport chain build
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We will have to update `genesis_template.json` but let wait #523 to be merged
> 
> Now that it is merged, a param entry will need to be added

Done [23919c8](https://github.com/tendermint/spn/pull/527/commits/23919c86836676750d4e2eb4dcbdfa8ff366e6ad)

lubtd:
> If I run `python3 undelegate.py 10000000 0 0`, this will remove all tokens from node1 and the localnet will no longer work. Do we want add some kind of check or do we think this is fine. If people fully undelegate from the main node, perhaps that's just on them.

If a single validator is removed from the validator set, the network should continue to run.
The above issue seems to occurs when a validator undelegates and is unbonded within the same block (which happen because we set here `unbonding_time` to 1 second for test purpose)
I set the `unbonding_time` to 10 seconds and the issue seems to no longer occur

lubtd:
> Do we want add some kind of check or do we think this is fine. If people fully undelegate from the main node, perhaps that's just on them.

The localnet will be mainly used internally for experimenting with IBC client and validator set changes. This is ok to have no check

aljo242 (approved):

Pantani (approved):

PR #534, opened 2022-02-17
feat(`campaign`):  metadata

### LINKED ISSUE #516: Add `metadata` field to `Campaign` object
Add a generic []byte `metadata` field to the campaign definition

This field should have a maximum length

The metadata can be provided from `MsgCreateCampaign`

For simplicity `MsgUpdateCampaignName` should be renamed to `MsgEditCampaign` and add the ability to edit the metadata attached to the campaign

The metatada is fully generic on-chain. Client can provide and interpret a JSON object

aljo242:
Closes #516

### What does this PR does?

Changes the `UpdateCampaignName` transaction to `EditCampaign` where both the name and metadata fields can be modified.

Adds `metadata []byte` field to the `Campaign` data structure.  This can be set using the `--metadata` field on the "create-campaign" and "edit-campaign" transactions.

NOTE:

The current maximum length for metadata fields is 100, which is most likely too short.  We should open another issue for deciding a proper value for this field and whether having this as an on-chain param makes sense.

### How to test?

Test the new CLI commands with flags.  Run the new tests.

## Checklist

review them:

- [X] Include tests


giunatale:
Wouldn't it be preferable to have `UpdateCampaign` to be more consistent with naming I've seen for other functionalities?

giunatale:
I think the length of `metadata` field should be a governance-ruled parameter, since it will impact the blockchain size. Are we planning on dealing with making this a param in a separate PR?
It's worthwhile considering doing it here imho.

giunatale (changes requested):

lubtd:
> I think the length of `metadata` field should be a governance-ruled parameter, since it will impact the blockchain size. Are we planning on dealing with making this a param in a separate PR? It's worthwhile considering doing it here imho.

@aljo242 if I remember it correctly, we decided to hardcoded it in the PR for simplicity and see if we want to have a param later

x/campaign/keeper/msg_server_edit_campaign.go
```diff
@@ -31,8 +31,14 @@ func (k msgServer) UpdateCampaignName(goCtx context.Context, msg *types.MsgUpdat
 			campaign.CoordinatorID,
 		))
 	}
+
 	campaign.CampaignName = msg.Name
```
lubtd, line 35: IMO we should also have the option to not update the name
```suggestion
if len(msg.Name) > 0 {
	campaign.CampaignName = msg.Name
}
```

x/campaign/keeper/msg_server_edit_campaign_test.go
```diff
@@ -35,49 +35,62 @@ func TestMsgUpdateCampaignName(t *testing.T) {
...
+		msg  types.MsgEditCampaign
+		err  error
 	}{
 		{
 			name: "invalid campaign id",
-			msg: types.MsgUpdateCampaignName{
+			msg: types.MsgEditCampaign{
 				Coordinator: coordAddr,
 				CampaignID:  100,
 				Name:        "new_name",
```
lubtd, line 47: Was apparently missing. We can use `sample.CampaignName()`

x/campaign/types/message_edit_campaign.go
```diff
@@ -45,5 +47,12 @@ func (msg *MsgUpdateCampaignName) ValidateBasic() error {
 	if err := CheckCampaignName(msg.Name); err != nil {
 		return sdkerrors.Wrap(ErrInvalidCampaignName, err.Error())
 	}
+
+	// TODO parameterize
+	if len(msg.Metadata) > spntypes.MaxMetadataLength {
```
lubtd, line 52: We should allows providing empty campaign name for not updating it
Since I believe `CheckCampaignName` fail for empty name, `CheckCampaignName` is only checked if the name is non-null
The message ValidateBasic should fail if none of the two value are provided

lubtd (changes requested):
Resending review, so you can push `Re-request review`

giunatale:
> Resending review, so you can push `Re-request review`

my bad, I triggered the review request here as a mistake.

x/campaign/simulation/simulation.go
```diff
@@ -264,12 +264,15 @@ func SimulateMsgCreateCampaign(ak types.AccountKeeper, bk types.BankKeeper, pk t
...
 			sample.CampaignName(),
 			sample.Coins(),
+			sample.Metadata(20),
 		)
 
 		return deliverSimTx(r, app, ctx, ak, bk, simAccount, msg, sdk.NewCoins())
 	}
 }
 
+// TODO add SimulateMsgEditCampaign
```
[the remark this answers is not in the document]
aljo242: I think this makes more sense as a separate PR/Issue, so I opened this [issue](https://github.com/tendermint/spn/issues/539) for now.

x/campaign/client/cli/tx_edit_campaign.go
```diff
@@ -10,11 +12,13 @@ import (
...
 
-func CmdUpdateCampaignName() *cobra.Command {
+const flagName = "flagName"
+
+func CmdEditCampaign() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "update-campaign-name [campaign-id] [name]",
-		Short: "Update the campaign name",
-		Args:  cobra.ExactArgs(2),
+		Use:   "edit-campaign [campaign-id]",
```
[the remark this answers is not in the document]
aljo242: We still require the campaign-id to be provided for any valid transaction, so I think it makes sense to keep this and the `cobra.ExactArgs(1)` line. 

What would the reasoning for removing this be?

aljo242:
> > I think the length of `metadata` field should be a governance-ruled parameter, since it will impact the blockchain size. Are we planning on dealing with making this a param in a separate PR? It's worthwhile considering doing it here imho.
> 
> @aljo242 if I remember it correctly, we decided to hardcoded it in the PR for simplicity and see if we want to have a param later

Issue created [here](https://github.com/tendermint/spn/issues/541).

x/campaign/client/cli/tx_edit_campaign.go
```diff
@@ -10,11 +10,13 @@ import (
 	"github.com/tendermint/spn/x/campaign/types"
 )
 
-func CmdUpdateCampaignName() *cobra.Command {
+const flagName = "flagCampaignName"
```
lubtd, line 13:
```suggestion
const flagName = "flag-campaign-name"
```

We should use kebab case for flag

lubtd (approved):
Perfect👍 Just a single suggestion for using kebab case convention for flag. I take the initiative to merge this change so the PR is approved form my side

lubtd (approved):

giunatale (approved):

PR #540, opened 2022-02-18
feat(reward): enable empty reward coins for `MsgSetRewards`

Pantani:
### What does this PR does?

Enable empty coins reward in `MsgSetRewards` so we can remove all rewards for a campaign

### How to test?

- Create the chain and set reward
```shell
spnd tx launch create-chain orbit-1 orbit.com 0xaaa --from alice && \
spnd tx reward set-rewards 1 1000uspn 100000 --from alice
```

- Remove the reward
```shell
spnd tx reward set-rewards 1 0token 0 --from alice
```


aljo242 (dismissed):
lgtm!

lubtd (commented):
I think we can totally remove the reward pool from the store if rewards are set to 0

I think we can also have the same behavior when the height is set to 0

x/reward/keeper/msg_server_set_reward_test.go
```diff
@@ -44,7 +50,22 @@ func TestMsgSetRewards(t *testing.T) {
...
+	})
+	zeroRewardHeightLaunchID := lk.AppendChain(sdkCtx, sample.Chain(5, res.CoordinatorID))
+	k.SetRewardPool(sdkCtx, types.RewardPool{
+		LaunchID:            zeroRewardHeightLaunchID,
+		Coins:               zeroRewarHeightBalance,
+		LastRewardHeight:    100,
+		CurrentRewardHeight: 30,
+	})
+
+	err = bk.MintCoins(sdkCtx, types.ModuleName, moduleBalance)
```
lubtd, line 68: I understand from the comment above that we must initialize coins object to initialize reward pools and mint here.

My suggestion would be to add in the test case structure a `InitialRewardPool`. When defined, the logic of initializing it and minting tokens would be performed in the body of the TDT.
It would become in my mind more in the idea of TDT where the each test case define (initialState, input, expectedOutput)

lubtd: Maybe we could also think in the future of even refactoring the coordinator part above where you provide the coordinator and associated balance in test cases

lubtd (commented):
CLI works well 👍

x/reward/keeper/msg_server_set_reward_test.go
Pantani: I agree!! I think this applies to all message test methods. We can define the initial keeper state and the expected output state. This can be a new issue, for we investigate the better way to do it.

x/reward/keeper/msg_server_set_reward_test.go
```diff
@@ -20,11 +20,17 @@ func TestMsgSetRewards(t *testing.T) {
...
 
 		ctx            = sdk.WrapSDKContext(sdkCtx)
-		moduleBalance  = sample.Coins()
-		newBalance     = sample.Coins()
 		provider       = sample.AccAddress()
 		invalidCoord   = sample.Address()
 		noBalanceCoord = sample.Address()
+
+		emptyCoinsBalance      = sample.Coins()
+		zeroRewarHeightBalance = sample.Coins()
```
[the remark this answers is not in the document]
Pantani: we need the coins to mind the balance before the test, but the current implementation is more clear

x/reward/keeper/msg_server_set_reward_test.go
lubtd: https://github.com/tendermint/spn/issues/556

lubtd (approved):

aljo242 (approved):

PR #542, opened 2022-02-18
refactor(rewards): add parameters to `MsgSetRewardsResponse`

Pantani:
### What does this PR does?

Add response parameters to`MsgSetRewardsResponse` so we can have better log messages into Starport: 
- `PreviousCoins`: previous reward pool coins value
- `PreviousLastRewardHeight`: previous reward pool last reward height value
- `NewCoins`: new reward pool coins value
- `NewLastRewardHeight`: new reward pool last pool height value


aljo242:
Once we move forward with emitting events on all transactions, I think it will make sense to migrate this information to events.  I believe that in some broadcast modes events can be collected.

aljo242 (approved):

lubtd (approved):

lubtd:
> Once we move forward with emitting events on all transactions, I think it will make sense to migrate this information to events. I believe that in some broadcast modes events can be collected.

I think it's fine. Since the response doesn't bring performance overhead, we can use it to describe the occurred state transition depending on our need in the client

PR #543, opened 2022-02-20
feat(`launch`): allow association of campaign with chain

### LINKED ISSUE #432: Add the ability for a chain with no campaign to be associated to a campaign
Add a `campaignID` field into the `MsgEditChain` message to add the ability to associate a chain to a campaign when the chain is not yet associated with a campaign

The message fails if the chain is already associated to a campaign
The message fails if the chain coordinator and the campaign coordinator doesn't match

aljo242:
Closes #432

### What does this PR does?

Adds the `setCampaignID` and `campaignID` fields to the `MsgEditChain` proto.  

### How to test?

- Create a chain without campaign:

```shell
spnd tx launch create-chain orbit-1 orbit.com 0xaaa --from alice
```

- Create a campaign and associate it with chain:

```shell 
spnd tx campaign create-campaign "orbit-test" "10000orbit" --from alice && \
spnd tx launch edit-chain 1 --set-campaign-id 1 --from alice
```

## Checklist

review them:

- [X] Include tests


proto/launch/tx.proto
```diff
@@ -48,7 +48,9 @@ message MsgEditChain {
   string sourceURL = 4;
   string sourceHash = 5;
   InitialGenesis initialGenesis = 6;
-  bytes metadata = 7;
+  bool setCampaignID = 7;
```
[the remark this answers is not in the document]
Pantani: the campaign id cannot be `0`. Maybe we don't need the `setCampaignID` flag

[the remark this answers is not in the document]
lubtd: > the campaign id cannot be 0. Maybe we don't need the setCampaignID flag

It can by design if we set the initial campaign counter to `0`, I would prefer to not rely on this assumption

aljo242:
See @lubtd comment above.  Since we don't want to rely on the assumption that the 0 value of campaignID will always be restricted, using this `setCampaignID` bool seems ok.

giunatale (approved):
@lubtd is test coverage complete in your opinion? are we missing something? asking you since you know better than me for sure.

lgtm though, so I'll approve for now.

x/launch/client/cli/tx_edit_chain.go
```diff
@@ -56,6 +57,8 @@ func CmdEditChain() *cobra.Command {
 				initialGenesis = &genesisURL
 			}
 
+			modifyCampaignID := cmd.Flags().Changed(flagCampaignID)
```
lubtd, line 60:
```suggestion
			setCampaignID := cmd.Flags().Changed(flagCampaignID)
```

x/launch/keeper/msg_server_edit_chain.go
```diff
@@ -50,6 +52,34 @@ func (k msgServer) EditChain(goCtx context.Context, msg *types.MsgEditChain) (*t
...
+			return nil, sdkerrors.Wrapf(profiletypes.ErrCoordInvalid,
+				"coordinator of the campaign is %d, chain coordinator is %d",
+				campaign.CoordinatorID,
+				chain.CoordinatorID,
+			)
+		}
+
+		chain.CampaignID = msg.CampaignID
+		chain.HasCampaign = true
+	}
```
lubtd, line 81: We must also call `AddChainToCampaign` of `campaignkeeper`

x/launch/keeper/msg_server_edit_chain_test.go
```diff
@@ -129,6 +188,47 @@ func TestMsgEditChain(t *testing.T) {
...
+				false,
+				false,
+				true,
+				999,
+				false,
+			),
+			err: campaigntypes.ErrCampaignNotFound,
+		},
+		{
+			name: "campaign does not exist",
```
lubtd, line 223:
```suggestion
			name: "campaign has a different coordinator",
```

x/launch/keeper/msg_server_edit_chain_test.go
```diff
@@ -129,6 +188,47 @@ func TestMsgEditChain(t *testing.T) {
...
+			msg: sample.MsgEditChain(coordAddress, launchID2,
+				false,
+				false,
+				false,
+				false,
+				true,
+				invalidCampaignID,
```
lubtd, line 230:
```suggestion
				campaignDifferentCoordinator,
```
Would make more explicit

lubtd (changes requested):
👍

For `x/launch/keeper/msg_server_edit_chain_test.go`
We need to add the new assertions based on the modification on the message logic:
- If `SetCampaignID == true` then we have assertion than the stored chain contain the campaign ID as `CampaignID` and the launch ID must be added in the `CampaignChains` structure in the `campaign` module

x/launch/keeper/msg_server_edit_chain_test.go
```diff
@@ -97,6 +138,20 @@ func TestMsgEditChain(t *testing.T) {
 				false,
 				false,
 				true,
+				validCampaignID,
+				false,
+			),
+		},
+		{
+			name: "edit metadata",
```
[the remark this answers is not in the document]
aljo242: No, this is to set the metadata. 

The `ModifyMetadata` argument is set to true, and all other args are false

x/launch/types/message_edit_chain.go
```diff
@@ -19,6 +19,8 @@ func NewMsgEditChain(
 	sourceURL,
 	sourceHash string,
 	initialGenesis *InitialGenesis,
+	hasCampaign bool,
```
lubtd, line 22: I think `setCampaign` was more self explanatory

giunatale (approved):

PR #551, opened 2022-02-21
fix(`profile`): check `coordByAddress` is associated to an active coordinator in genesis valdiate

### LINKED ISSUE #547: Genesis `Validate`: add check for `CoordinatorByAddress` indexing an inactive coordinator
https://github.com/tendermint/spn/pull/500 brought a new invariant: an entry in the `CoordinatorByAddress` should not index a coordinator that is inactive (because disabling the coordinator removes the associated `CoordinatorByAddress` entry).

We should add the check in the `ValidateCoordinators` method in `types/genesis` that all `CoordinatorByAddress` index a active `Coordinator`

aljo242:
Closes #547

### DIFF
--- a/x/profile/types/genesis.go
+++ b/x/profile/types/genesis.go
@@ -98,11 +98,11 @@ func (gs GenesisState) ValidateCoordinators() error {
 		}
 		index := string(CoordinatorByAddressKey(elem.Address))
 		_, found := coordinatorByAddressIndexMap[index]
-		if !found {
-			return errors.New("coordinator address not found for CoordinatorByAddress")
-		}
 
-		if !elem.Active {
+		switch {
+		case !found && elem.Active:
+			return errors.New("coordinator address not found for CoordinatorByAddress")
+		case found && !elem.Active:
 			return errors.New("coordinator found by CoordinatorByAddress should not be inactive")
 		}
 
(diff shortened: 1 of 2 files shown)


x/profile/types/genesis.go
```diff
@@ -97,9 +97,15 @@ func (gs GenesisState) ValidateCoordinators() error {
 			return errors.New("coordinator id should be lower or equal than the last id")
 		}
 		index := string(CoordinatorByAddressKey(elem.Address))
-		if _, ok := coordinatorByAddressIndexMap[index]; !ok {
+		_, found := coordinatorByAddressIndexMap[index]
+		if !found {
 			return errors.New("coordinator address not found for CoordinatorByAddress")
 		}
+
+		if !elem.Active {
```
lubtd, line 105: In the previous state, there was a 1-1 relationship between `Coordinator` and `CoordinatorByAddress` therefore we checked that all `Coordinator` had an associated `CoordinatorByAddress`

Now a `Coordinator` can eventually have no matched `CoordinatorByAddress` but if this is the case  the coordinator must be inactive.

Therefore we should rather have condition that allows inactive coordinators in the initial genesis

```
if !found && elem.Active {
	return errors.New("....")
}

if found && !elem.Active {
	return errors.New("....")
}
```

lubtd (changes requested):

Pantani (approved):

lubtd (approved):
👍

One comment: we should try to put PR name as explicit as possible

PR #561, opened 2022-02-23
feat(`launch`): collapse `minLaunchTime` and `maxLaunchTime` params into atomic `LaunchTimeRange`

### LINKED ISSUE #525: Collapse `MinLaunchTime` and `MaxLaunchTime` params into single struct
As of now these two parameters are handled as separate. This also means any of the two can be updated independently.
Although in principle this is a good thing, it also prevents from making sure the two values that define the range are always consistent, *i.e.* it could be that `minLaunchTime` > `maxLaunchTime` and this would not be caught, leading to potential dangerous consequences.

We should collapse the two values into a single `range` struct that is updated as a single object. This would make possible to enforce consistency of the range, similarly to what was done for [`campaign`](https://github.com/tendermint/spn/pull/524) with `MinTotalSupply` and `MaxTotalSupply`.

giunatale:
Closes #525 

### What does this PR does?

The two `launch` params in `minLaunchTime` and `maxLaunchTime` are now collapsed into a single struct `LaunchTimeRange`. Genesis validation tests and localnet template genesis have also been updated accordingly.

## Checklist

review them:

- [x] Include tests


giunatale:
wondering what caused the simulation tests to fail after merging with `develop`, any idea?

x/launch/types/genesis_test.go
```diff
@@ -441,20 +441,6 @@ func TestGenesisState_ValidateParams(t *testing.T) {
 			},
 			shouldBeValid: false,
 		},
-		{
-			desc: "min launch time above max launch time",
```
aljo242, line 445: Why are these tests removed?

aljo242 (changes requested):

proto/launch/params.proto
```diff
@@ -9,7 +9,11 @@ option go_package = "github.com/tendermint/spn/x/launch/types";
...
-  int64 revertDelay = 3   [(gogoproto.moretags) = "yaml:\"revert_delay\""];
-}
\ No newline at end of file
+  LaunchTimeRange launchTimeRange = 1 [(gogoproto.nullable) = false];
+  int64 revertDelay = 2;
+}
+
+message LaunchTimeRange {
+  uint64 minLaunchTime = 1;
+  uint64 maxLaunchTime = 2;
```
lubtd, line 18: What do you think about setting the type for both fields to `int64` since they represent Unix timestamp?
We should then add check for `minLaunchTime>0`

Then we can use `int64` for `remainingTime` in `MsgTriggerLaunch` and prevent from doing the `int64` conversion that can bring security problematics

lubtd (commented):
👍

proto/launch/params.proto
giunatale: Do we want also to allow that `MaxLaunchTime == MinLaunchTime`?

giunatale:
> Do we want also to allow that `MaxLaunchTime == MinLaunchTime`?

I haven't allowed this in the recent changes to not change the current behavior. Holding off until I have your opinion.

Simtests fails are unrelated to this PR, will update once we have #562 merged in `develop`.

lubtd:
>Do we want also to allow that MaxLaunchTime == MinLaunchTime?

I see no issue related to having an equal value, a single launch time would be allowed, but at the same time and don't think it's really important it's just one-second difference, in practice I expected that we will always have a range of few days of difference to give some flexibility. The choice of the minimal and maximal one matter

giunatale:
> > Do we want also to allow that MaxLaunchTime == MinLaunchTime?
> 
> I see no issue related to having an equal value, a single launch time would be allowed, but at the same time and don't think it's really important it's just one-second difference, in practice I expected that we will always have a range of few days of difference to give some flexibility. The choice of the minimal and maximal one matter

I'll leave as it is 👍

lubtd:
Seems like I can't serve the chain with Starport 🤔
```
starport chain serve -r
Error: failed to validate genesis state: failed to unmarshal launch genesis state: unknown field "minLaunchTime" in types.Params
```

x/launch/types/message_trigger_launch.go
```diff
@@ -9,7 +9,7 @@ const TypeMsgTriggerLaunch = "trigger_launch"
 
 var _ sdk.Msg = &MsgTriggerLaunch{}
 
-func NewMsgTriggerLaunch(coordinator string, launchID, remainingTime uint64) *MsgTriggerLaunch {
+func NewMsgTriggerLaunch(coordinator string, launchID uint64, remainingTime int64) *MsgTriggerLaunch {
```
lubtd, line 12: We should add the check `remainingTime>0` in `ValidateBasic` with a test

x/launch/types/params.go
```diff
@@ -38,12 +37,19 @@ func ParamKeyTable() paramtypes.KeyTable {
...
+func NewLaunchTimeRange(minLaunchTime, maxLaunchTime int64) LaunchTimeRange {
+	return LaunchTimeRange{
 		MinLaunchTime: minLaunchTime,
 		MaxLaunchTime: maxLaunchTime,
-		RevertDelay:   revertDelay,
+	}
+}
+
+// NewParams creates a new Params instance
+func NewParams(minLaunchTime, maxLaunchTime int64, revertDelay int64) Params {
```
lubtd, line 49:
```suggestion
func NewParams(minLaunchTime, maxLaunchTime, revertDelay int64) Params {
```

x/launch/types/message_trigger_launch.go
giunatale: true! thanks for the suggestion

giunatale: true! thanks for the suggestion

giunatale: I've added a test case for `remainingTime<=0` only in the test for `ValidateBasic` following the testing conventions we have discussed 👍

giunatale:
> Seems like I can't serve the chain with Starport 🤔
> 
> ```
> starport chain serve -r
> Error: failed to validate genesis state: failed to unmarshal launch genesis state: unknown field "minLaunchTime" in types.Params
> ```

The `config.yml` was outdated. My fault. Fixed

lubtd:
From the `serve` issue stated above, the genesis generated from `starport serve` contains
```
"launch": {
      "chainCounter": "1",
      "chainList": [],
      "genesisAccountList": [],
      "genesisValidatorList": [],
      "params": {
        "launchTimeRange": {
          "maxLaunchTime": "604800",
          "minLaunchTime": "5"
        },
        "minLaunchTime": "5",
        "revertDelay": "5"
      },
      "requestCounterList": [],
      "requestList": [],
      "vestingAccountList": []
    },
```
`minLaunchTime` is still present in the params

lubtd (approved):

aljo242:
Will approve after merge conflicts are resolved!

giunatale:
@aljo242 can you re-review this PR?

aljo242 (approved):

giunatale:
@lubtd @Pantani this needs one more review before it can be merged. Can any of you take this on? 🙏

lubtd (approved):

PR #562, opened 2022-02-23
fix(`launch`): add check for MsgSettleRequest simulation

### LINKED ISSUE #405: Simulation test: `MsgSettleRequest` verify that the request can be apply before approving it
https://github.com/tendermint/spn/blob/3d0d523cd21296202d9dcea66e43f623cc7f40f9/x/launch/simulation/simulation.go#L362

There is no verification if the request can be applied before sending the message (eg: the genesis account already exists)
If the request can't be applied, the message fails and the simulation stops
We should check if the request can be applied and send a `NoOpMsg ` if not (or reject the request)

aljo242:
Closes #405

Perform a check using the logic of `ApplyRequest()` without modifying state.  Implemented in a new function in the simulation package called `checkRequest()`.


giunatale:
Is this fixing the current simulation error on `develop`?

aljo242:
> Is this fixing the current simulation error on `develop`?

Yessir :)

giunatale (changes requested):
small lint fix, otherwise LGTM

x/launch/simulation/store.go
```diff
@@ -2,6 +2,8 @@ package simulation
 
 import (
 	"fmt"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
```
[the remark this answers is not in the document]
giunatale: we should keep this format here, where imports are grouped with spn at the bottom. That's what I've seen done everywhere.

lubtd:
Can you try the simulation with some different seed to ensure if we no longer get this error?

```
starport chain simulate --seed [nb]
```

aljo242:
> 

Ran with seeds [1, 15] and everything is ok.

x/launch/simulation/store.go
```diff
@@ -163,3 +165,76 @@ func FindRandomValidator(
 	}
 	return simAccount, valAcc, found
 }
+
+func checkRequest(
```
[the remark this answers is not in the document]
aljo242: Ok, I removed the check part and created `CheckRequest` in the keeper package. This means we perform the switch-case on the request contents twice (once to check, once to apply), but the code is a bit cleaner now i think.

aljo242:
Ok, I removed the check part and created CheckRequest in the keeper package. This means we perform the switch-case on the request contents twice (once to check, once to apply), but the code is a bit cleaner now i think.

lubtd:
> Ok, I removed the check part and created CheckRequest() in the keeper package. This means we perform the switch-case on the request contents twice (once to check, once to apply), but the code is a bit cleaner now i think.

This is fine

x/launch/keeper/request.go
```diff
@@ -133,6 +133,40 @@ func ApplyRequest(
...
+		vr := requestContent.ValidatorRemoval
+		k.RemoveGenesisValidator(ctx, launchID, vr.ValAddress)
+	default:
+		return spnerrors.Critical("unknown request content type")
+	}
+	return nil
+}
+
+// CheckRequest verifies that a request can be applied
+func CheckRequest(
```
lubtd, line 165: We should create the `TestCheckRequest` test to test request checking

`TestApplyRequest` would keep the state transition assertion and should verify that it fails if a request is invalid (without testing all invalid requests)

giunatale (dismissed):

x/launch/keeper/request_test.go
```diff
@@ -183,3 +184,85 @@ func TestApplyRequest(t *testing.T) {
...
+	)
+	tests := []struct {
+		name    string
+		request types.Request
+		err     error
+	}{
+		{
+			name:    "test GenesisAccount content",
+			request: sample.RequestWithContent(launchID, contents[0]),
+			err:     nil,
```
lubtd, line 206:
```suggestion
```

x/launch/keeper/request_test.go
```diff
@@ -183,3 +184,85 @@ func TestApplyRequest(t *testing.T) {
...
+		}, {
+			name:    "test duplicated GenesisAccount content",
+			request: sample.RequestWithContent(launchID, contents[0]),
+			err:     types.ErrAccountAlreadyExist,
+		}, {
+			name:    "test genesis AccountRemoval content",
+			request: sample.RequestWithContent(launchID, contents[1]),
+			err:     nil,
```
lubtd, line 214:
```suggestion
```

x/launch/keeper/request_test.go
```diff
@@ -183,3 +184,85 @@ func TestApplyRequest(t *testing.T) {
...
+		}, {
+			name:    "test not found genesis AccountRemoval content",
+			request: sample.RequestWithContent(launchID, contents[1]),
+			err:     types.ErrAccountNotFound,
+		}, {
+			name:    "test VestingAccount content",
+			request: sample.RequestWithContent(launchID, contents[2]),
+			err:     nil,
```
lubtd, line 222:
```suggestion
```

x/launch/keeper/request_test.go
```diff
@@ -183,3 +184,85 @@ func TestApplyRequest(t *testing.T) {
...
+		}, {
+			name:    "test duplicated VestingAccount content",
+			request: sample.RequestWithContent(launchID, contents[2]),
+			err:     types.ErrAccountAlreadyExist,
+		}, {
+			name:    "test vesting AccountRemoval content",
+			request: sample.RequestWithContent(launchID, contents[3]),
+			err:     nil,
```
lubtd, line 230:
```suggestion
```

x/launch/keeper/request_test.go
```diff
@@ -183,3 +184,85 @@ func TestApplyRequest(t *testing.T) {
...
+		}, {
+			name:    "test not found vesting AccountRemoval content",
+			request: sample.RequestWithContent(launchID, contents[3]),
+			err:     types.ErrAccountNotFound,
+		}, {
+			name:    "test GenesisValidator content",
+			request: sample.RequestWithContent(launchID, contents[4]),
+			err:     nil,
```
lubtd, line 238:
```suggestion
```

x/launch/keeper/request_test.go
```diff
@@ -183,3 +184,85 @@ func TestApplyRequest(t *testing.T) {
...
+		}, {
+			name:    "test duplicated GenesisValidator content",
+			request: sample.RequestWithContent(launchID, contents[4]),
+			err:     types.ErrValidatorAlreadyExist,
+		}, {
+			name:    "test ValidatorRemoval content",
+			request: sample.RequestWithContent(launchID, contents[5]),
+			err:     nil,
```
lubtd, line 246:
```suggestion
```

lubtd (approved):

Issue #369, opened 2021-10-27
Simulation test: `out of gas` error for some simulation


lubtd:
**Describe the bug**
When setting values for `FlagNumBlocksValue` and `FlagBlockSizeValue` flags, the simulation test can fail with the error `out of gas in location: WritePerByte; gasWanted: 1000000, gasUsed: 1000373: out of gas`
While it particularly happen when setting high values, the error can sometimes occurs for random low values. Therefore, there might be a flaw in our configuration for simulation tests 

**To Reproduce**
Set following values in `app/simulation_test.go`
```
simapp.FlagNumBlocksValue = 500
simapp.FlagBlockSizeValue = 100
```
```
Simulating... block 50/500, operation 0/273.    simulate.go:295: error on block  50/500, operation (14/80) from x/staking:
        out of gas in location: WritePerByte; gasWanted: 1000000, gasUsed: 1000373: out of gas
```

giunatale:
I think this error is due to this value https://github.com/cosmos/cosmos-sdk/blob/master/simapp/helpers/test_helpers.go#L17
being used here https://github.com/cosmos/cosmos-sdk/blob/master/x/simulation/util.go#L107 and other places.

We can't override it being a constant, so the best option is to use a custom `GenAndDeliverTxWithRandFees` that internally uses a `GenAndDeliverTx` with a different value for gas.

However, I've seen this error arise also for SDK internal modules (e.g. staking) where we can't really do much.

Thoughts?

giunatale:
https://github.com/cosmos/cosmos-sdk/issues/9463 
JayB may know more about this

kogisin:
@giunatale @lubtd That issue was created a while ago, so if I had to trace it down for a bit. If I remember it correctly, it occurs due to the usage of denomination other than `sdk.DefaultBondDenom` when spending fees for a transaction (I don't know if this is the case for the `spn` module). A quick fix was to increase gas, but since it is dependent on Cosmos SDK we had to resolve the issue by having a custom function that returns random amount of fee with only default bond denom. See [this function](https://github.com/tendermint/liquidity/blob/develop/x/liquidity/simulation/operations_utils.go#L94) and where it is being used in operations.

From our team's experience of developing other simulation tests, we decide to have our own custom `GenAndDeliverTx` and `GenAndDeliverTxWithFees` functions so that we can pass whatever gas and fees we desire to be spent for a transaction. 

Reference the following code lines:

[...]

PR #565, opened 2022-02-24
test: Simulation and Makefile enhancements

### LINKED ISSUE #367: Simulation test: split current test into simple simulation test and benchmark test
We might need to fix https://github.com/tendermint/spn/issues/369 first

Split the current test in `app/simulation_test.go` into:

- Simple simulation test to test the application as a whole
- Benchmark test with high flag values for blocks to benchmark the application

Add the simple test into the CI

aljo242:
Closes #367 

### What does this PR does?

Adds new `Makefile` targets for more simulation options.  Add a new non-determinism test which runs multiple sims to verify the state is the same multiple times.

Modifies the `simulation.yml` workflow so that it runs a `Makefile` target, but in effect the test is the same simulation with the same parameters.

### How to test?

Run the new targets:
- `make test-race`
- `make test-cover`
- `make bench`
- `make test-sim-nondeterminism`
- `make test-sim-ci`
- `make test-sim-benchmark`
- `make test-sim-profile`

Note that the larger sims (nondeterminism, benchmark, profile) run into the out of gas issue described in #369. This will need to be addressed in a new PR.


Makefile
```diff
@@ -18,19 +24,68 @@ lint:
 	@echo Running gocilint...
 	@golangci-lint run --out-format=tab --issues-exit-code=0
 
+help: Makefile
```
lubtd, line 27: Show
```
 Choose a command run in , or just run 'make' for install

  govet       Run go vet.
  format      Run gofmt and goimports.
  lint        Run Golang CI Lint.
  test-unit   Run the unit tests.
  test-unit   Run the unit tests checking for race conditions
  test-unit   Run the unit tests and create a coverate html report
  test
```
`test-unit` instead of `test-race` and `test-cover`

app/simulation_test.go
```diff
@@ -86,28 +98,107 @@ func BenchmarkSimulation(b *testing.B) {
...
+	err = simapp.CheckExportSimulation(app, config, simParams)
 	require.NoError(b, err)
 	require.NoError(b, simErr)
 
 	if config.Commit {
 		simapp.PrintStats(db)
 	}
 }
+
+func TestAppStateDeterminism(t *testing.T) {
```
lubtd, line 127: This is very interesting, I thought incorrect `apphash` was raised in the classic simulation testing we have.

If this is relevant, we can maybe raise an issue on `starport` to include this into the app scaffolding since we already scaffold simulation testing

lubtd (commented):
I think we can add `cover.html` to `.gitignore`

app/simulation_test.go
aljo242: You need to run the same set of transactions (basically just using the same seed) multiple times and save the app hash of the previous run to check it against the current state.

There are a lot more simulation options used in the [Cosmos SDK repo](https://github.com/cosmos/cosmos-sdk/blob/master/simapp/sim_test.go) that gave me inspiration.  

I think the non-determinism testing is one of the most important parts of having simulation testing, so it's probably relevant to raise an issue on `starport`, yes.

giunatale (approved):
looking good! really handy QoL improvement

PR #566, opened 2022-02-25
feat(`monitoringp`): skip block height 1 in begin-blocker

### LINKED ISSUE #564: Begin-blocker: skip block `1` for counting signatures
We currently in the begin-blocker starting counting signature at height `1` for monitoring packet.

However, the first block is the genesis block and not signed by any validator in the validator set. This give a monitoring packet containing value like:
```
    blockCount: "60"
    counts:
    - RelativeSignatures: "19.666666666666666647"
      consAddress: iYH/hrUeZvFXtBpNe2UUBxiCdjw=
    - RelativeSignatures: "19.666666666666666647"
      consAddress: 4aI+0x4FyCcJ176uXMUBXhN7GIU=
    - RelativeSignatures: "19.666666666666666647"
      consAddress: 7hA1g2S4ulkRUmWYAoa2ukIYZUA=
```

If validators signed all blocks we should have
```
    blockCount: "60"
    counts:
    - RelativeSignatures: "20"
      consAddress: iYH/hrUeZvFXtBpNe2UUBxiCdjw=
    - RelativeSignatures: "20"
      consAddress: 4aI+0x4FyCcJ176uXMUBXhN7GIU=
    - RelativeSignatures: "20"
      consAddress: 7hA1g2S4ulkRUmWYAoa2ukIYZUA=
```

If should skip the first height in the Begin-Blocker

aljo242:
Closes #564

### What does this PR does?

Skip the `BeginBlock()` function in `x/monitoringp/module.go` when BlockHeight == 1.  I'm not sure how to test this other than recreating the scenario described in the issue.  Any thoughts are welcome.

### DIFF
--- a/x/monitoringc/keeper/msg_server_create_client_test.go
+++ b/x/monitoringc/keeper/msg_server_create_client_test.go
@@ -2,9 +2,10 @@ package keeper_test
 
 import (
 	"encoding/base64"
-	ibctmtypes "github.com/cosmos/ibc-go/v2/modules/light-clients/07-tendermint/types"
 	"testing"
 	"time"
+
+	ibctmtypes "github.com/cosmos/ibc-go/v2/modules/light-clients/07-tendermint/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/stretchr/testify/require"
--- a/x/monitoringp/keeper/begin_block.go
+++ b/x/monitoringp/keeper/begin_block.go
@@ -20,6 +20,11 @@ const (
 
 // ReportBlockSignatures gets signatures from blocks and update monitoring info
 func (k Keeper) ReportBlockSignatures(ctx sdk.Context, lastCommit abci.LastCommitInfo, blockHeight int64) {
+	// skip first block because it is not signed
+	if blockHeight == 1 {
+		return
+	}
+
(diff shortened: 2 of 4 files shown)


giunatale:
> Skip the `BeginBlock()` function in `x/monitoringp/module.go` when BlockHeight == 1. I'm not sure how to test this other than recreating the scenario described in the issue. Any thoughts are welcome.

I would indeed just test that. I think at least 2 test cases (BH == 1 and != 1) would be needed for regression purposes. Hopefully testing this is not too convoluted

lubtd:
I think we can add the logic in `ReportBlockSignatures` and then test the change here. We assert that no change on the monitoring info are performed on the first block

lubtd (changes requested):
See my suggestion above

lubtd (approved):
👍
Great, there is no almost no token refunded to the coordinator when validators sign all the blocks. Here's just 2 from decimal floor conversion
```
spnd q bank balances spn1ezptsm3npn54qx9vvpah4nymre59ykr9exx2ul

balances:
- amount: "200000000"
  denom: uspn
- amount: "2"
  denom: v/1/orbit
pagination:
  next_key: null
  total: "0"
```

giunatale (approved):

PR #567, opened 2022-02-28
feat(`campaign`): add `CampaignCreationFee` param

### LINKED ISSUE #538: Create `CampaignCreationFee` param
Add a `CampaignCreationFee` as param in the`campaign` module

This param would be `EmptyCoin`s by default

If a value is defined, a fee is transferred from the coordinator account and set to the... community pool?

giunatale:
Closes #538 

### What does this PR does?

Adds a `CampaignCreationFee` param (an `sdk.Coins`) to `campaign`. This fee is deducted at campaign creation if set (which means creation is prevented if account balance is not sufficient). Default value is an empty `sdk.Coins`, so no fee.

Collected fees are used to fund the `CommunityPool` hence why the need for including the `distribution` keeper to `campaign`.

### How to test?

Set a meaningful value for `CampaignCreationFee` param. Try creating a campaign with or without the sufficient balance to cover for the fee set. The latter case should fail.

## Checklist

review them:

- [x] Include tests


x/campaign/keeper/msg_server_create_campaign_test.go
```diff
@@ -78,9 +115,20 @@ func TestMsgCreateCampaign(t *testing.T) {
...
 				TotalSupply:  sample.CoinsWithRange(10, 20),
+				Metadata:     sample.Metadata(20),
 			},
 			err: types.ErrInvalidTotalSupply,
 		},
+		{
+			name: "insufficient balance to cover creation fee",
+			msg: types.MsgCreateCampaign{
+				CampaignName: sample.CampaignName(),
+				Coordinator:  coordAddr1,
```
lubtd, line 126: We reuse `coordAddr1` here used above. Does it fails because `coordAddr1` has only enough funds for 1 message?
I would prefer creating a coordinator without fund specifically for this test case to make it more maintainable
For example, the test above `"create a campaign with an invalid token supply"` fails because `TotalSupply` is invalid but it would also fail because of insufficient funds. I would consider giving to `coordAddr1` enough funds for all other test cases

x/campaign/types/params.go
```diff
@@ -11,10 +11,12 @@ import (
...
 var (
-	DefaultMinTotalSupply = sdk.NewInt(100)                   // One hundred
-	DefaultMaxTotalSupply = sdk.NewInt(1_000_000_000_000_000) // One Quadrillion
+	DefaultMinTotalSupply      = sdk.NewInt(100)                   // One hundred
+	DefaultMaxTotalSupply      = sdk.NewInt(1_000_000_000_000_000) // One Quadrillion
+	DefaultCampaignCreationFee = sdk.Coins(nil)                    // EmptyCoins
 
-	ParamStoreKeyTotalSupplyRange = []byte("totalsupplyrange")
+	ParamStoreKeyTotalSupplyRange    = []byte("totalsupplyrange")
+	ParamStoreKeyCampaignCreationFee = []byte("campaigncreationfee")
```
lubtd, line 19:
```suggestion
	KeyTotalSupplyRange    = []byte("TotalSupplyRange")
	KeyCampaignCreationFee = []byte("CampaignCreationFee")
```

To be consistent with other module

x/campaign/types/params_test.go
```diff
@@ -82,3 +87,37 @@ func TestValidateTotalSupplyRange(t *testing.T) {
...
+			name:        "invalid interface",
+			creationFee: "test",
+			err:         fmt.Errorf("invalid parameter type: string"),
+		},
+		{
+			name:        "invalid coin",
+			creationFee: sdk.Coins{sdk.Coin{Denom: "foo", Amount: sdk.NewInt(-1)}},
+			err:         errors.New("coin -1foo amount is not positive"),
+		},
+		{
```
lubtd, line 107: If `DefaultCampaignCreationFee` is empty coins, we should also have a test case with `sample.Coins()`

lubtd (changes requested):
We should assert in `MsgCreateCampaign` handler tests that the signer balance has been decreased by the fee in the Table Tests body and, if possible, that the Community Pool has been increased by the fee

lubtd (approved):
👍

giunatale:
*note*: testing the community pool balance would have introduced too much further complexity (and some invasive changes) for a minimal check, so I deemed it not worth it.

aljo242 (approved):
Nice! Thanks for showing how the CommunityPool works!  I didn't even know that function in the DistKeeper existed

PR #568, opened 2022-02-28
feat(`reward`): close reward pool instead of removing in DistributeRewards

### LINKED ISSUE #545: Not removing the Reward Pool from store once rewards are distributed
For documentation purposes, we should keep the Reward Pool information for previous chains.

- Introduce a boolean `closed` set to `true` once all rewards for a chain are distributed
- Introduce a `InitialCoins` value that contains the initial tokens set for reward pool and that never change

- `DistributeRewards` should fail if the reward pool is closed

aljo242:
Closes #545

### What does this PR does?

Adds the fields `initialCoins` and `closed` to the `RewardPool` proto message.
- `initalCoins` is set when a reward pool is set.  It is unmodified during `DistributeRewards()`.  This serves as a record of the reward pool even when it is closed
- changed `coins` field to `currentCoins`. 

RewardPools are closed in `DistributeRewards()` instead of being removed from the KVStore.  
- `DistributeRewards()` returns `types.ErrRewardPoolClosed` if reward pool closed

## Checklist

review them:

- [X] Include tests


x/monitoringc/keeper/msg_server_create_client_test.go
```diff
@@ -2,10 +2,11 @@ package keeper_test
 
 import (
 	"encoding/base64"
-	ibctmtypes "github.com/cosmos/ibc-go/v2/modules/light-clients/07-tendermint/types"
 	"testing"
 	"time"
 
+	ibctmtypes "github.com/cosmos/ibc-go/v2/modules/light-clients/07-tendermint/types"
+
```
lubtd, line 9:
```suggestion
```

x/reward/keeper/reward_distribution_test.go
```diff
@@ -322,9 +354,9 @@ func TestKeeper_DistributeRewards(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			// set test reward pool if contains coins
-			if tt.rewardPool.Coins != nil {
+			if tt.rewardPool.CurrentCoins != nil {
```
lubtd, line 357: We should add in the assertion that `InitialCoin` is left unchanged in all case
And I realize we should also check that `Coins`/`CurrentCoins`/`RemainingCoins` is decreased by the sum of `tt.WantBalances`

It can be also part of https://github.com/tendermint/spn/issues/502 but I may think it makes sense to add the assertions in this PR

x/reward/types/reward_pool.go
```diff
@@ -12,17 +12,29 @@ func NewRewardPool(launchID uint64, currentRewardHeight int64) RewardPool {
...
 		return fmt.Errorf("invalid reward pool coins: %s", err)
 	}
+	if m.CurrentCoins.Empty() {
+		return errors.New("empty reward pool coins")
+	}
+	if err := m.CurrentCoins.Validate(); err != nil {
+		return fmt.Errorf("invalid reward pool coins: %s", err)
+	}
+
+	if m.CurrentCoins.IsAnyGTE(m.InitialCoins) {
```
lubtd, line 34: We should ensure that you can't have a denom in `CurrentCoins` not present in `InitialCoins`, does this check verify this? We should add a unit test for this case

x/reward/keeper/reward_distribution_test.go
aljo242: I added the `InitialCoin` check.

The content of #502 was giving me some trouble.  I added a check (which is now commented out) but it was causing some of the the tests to fail. Thoughts?

lubtd:
```
// TODO: https://github.com/tendermint/spn/issues/502
// assert the module account has no coin left
// accAddr, err := sdk.AccAddressFromBech32(tt.rewardPool.Provider)
// require.NoError(t, err)
// err = bk.SendCoinsFromModuleToAccount(ctx, types.ModuleName, accAddr,
//	sdk.NewCoins(sdk.NewCoin(
//		rewardPool.InitialCoins.GetDenomByIndex(0),
//		sdk.OneInt(),
//	)))
// require.ErrorIs(t, err, sdkerrors.ErrInsufficientFunds)
```
Mmh I don't actually get what we are trying to test here
At the end of the table test body we are burning back the values for each account specified in `WantBalances` to make each test case independent and stateless. We should not from this design send coins to account balances

The idea mentioned above was that we fetch at the beginning of the body `previousInitialCoins` and `previousRemainingCoins` and we assert after `DistributeRewards`
```
previousInitialCoins == newInitialCoins
previousRemainingCoins == newRemainingCoins + sum(wantBalances)
```
I don't think we should transfer balances

aljo242: Got it.  I was trying to find a way to actually check the amount stored in the module account is 0.  As far as I am aware, the only way to check that is to try to send from the module account.  If we send 1 coin from it and get `ErrInsufficientFunds`, this means that the balance has to be 0.

x/reward/keeper/reward_distribution_test.go
```diff
@@ -321,11 +353,26 @@ func TestKeeper_DistributeRewards(t *testing.T) {
...
+			if tt.rewardPool.RemainingCoins != nil {
 				k.SetRewardPool(ctx, tt.rewardPool)
-				err := bk.MintCoins(ctx, types.ModuleName, tt.rewardPool.Coins)
+				err := bk.MintCoins(ctx, types.ModuleName, tt.rewardPool.RemainingCoins)
 				require.NoError(t, err)
+
+				previousRewardPool, found = k.GetRewardPool(ctx, tt.args.launchID)
+				require.True(t, found)
+				previousInitialCoins = previousRewardPool.InitialCoins
+				previousRemainingCoins = previousRewardPool.RemainingCoins
```
lubtd, line 372: I think directly assigning the values from `tt.rewardPool` is sufficient since `SetRewardPool` is a method already tested

lubtd (approved):
Nice👌
```
spnd q reward list-reward-pool
pagination:
  next_key: null
  total: "0"
rewardPool:
- closed: true
  currentRewardHeight: "0"
  initialCoins:
  - amount: "50000"
    denom: v/1/orbit
  lastRewardHeight: "30"
  launchID: "1"
  provider: spn1ezptsm3npn54qx9vvpah4nymre59ykr9exx2ul
  remainingCoins: []
```

lubtd (approved):

lubtd (approved):

Pantani (approved):

PR #577, opened 2022-03-01
refactor(`monitoring-provider`): count signatures for operator address

lubtd:
### NOTE: Don't mind the `reward` unit tests error, we need to implement the second part of the refactoring to fix these tests. This PR is not merged on `develop`

Part of https://github.com/tendermint/spn/issues/552

Add `staking` as dependency for monitoring provider and count signature for the operator address instead of directly the consensus address

To test: https://github.com/lubtd/orbit use the branch
Operator address can be visualized in the monitoring info (cons address is still here to keep some comptabilities)
```
MonitoringInfo:
  signatureCounts:
    blockCount: "5"
    counts:
    - RelativeSignatures: "1.333333333333333332"
      consAddress: null
      opAddress: cosmosvaloper17qqqsmyx43efr5ywp33h35l4dlmacfxvtkyed5
    - RelativeSignatures: "1.333333333333333332"
      consAddress: null
      opAddress: cosmosvaloper1kgjf5nyuk2288tqe540ujt8l2zrd83pygumegm
    - RelativeSignatures: "1.333333333333333332"
      consAddress: null
      opAddress: cosmosvaloper1v5a0gg2aeflfkzvjwqucu20arlawj0aywquu22
  transmitted: false
```

Issue #552, opened 2022-02-22
Replacing `consensusAddresses` of Validator profile with `operatorAddresses`


lubtd:
Thanks @ilgooz for some of the investigation on this

We currently associate to a validator profile a list of Tendermint Consensus Addresses to dispatch the rewards distribution to the correct address on `spn`. The monitoring packet then associates the signature count with Cons Addresses.
To prove Consensus Address ownership `MsgSetConsAddress` verify a signature from the Tendermint key to prove ownership of the key and set the cons address to the validator profile.

Under some practical circumstances, we can't use the Tendermint key to sign any messages outside of the blocks of the related blockchain. This can also be a bad practice in real use cases.

To let the Tendermint Key isolated, the alternative can be to instead associate the operator addresses (Cosmos level) to the validator profile.
This change would simplify the implementation on `profile` because we would no longer need to track the nonce in the module but `auth` would directly verify message signature
This would add some complexity in the `monitoringp` module because we need to add `staking` dependency to the module to fetch the operator address from the voter in begin-blocker

Another downside: the new dependency may make more difficult to support PoA in the future, although the PoA module should also contain a Operator validator address.

Full changes:
Profile:
- Remove `ConsensusKeyNonce` structure
- Remove `MsgSetConsAddress`
- Replace `ConsensusAddresses` field of `Validator` into a string array `[]OperatorAddresses`
- Create `MsgSetOperatorAddress` or `MsgAddOperatorAddress` the message simply add the SDK operator address (`sdk.ValAddress`) in the the `OperatorAddresses` field of the validator profile. (Ideally the message should be signed by the validator address itself and the operator address (multiple signer)). The operator address is stored with the address prefix of `spn`
- Replace `ValidatorByConsAddress` structure by `ValidatorByOperatorAddress`. The idea remains the same 

Monitoring Provider:
- Change `SignatureCounts` structure to include `operatorAddress` instead of `consAddress`
- Add `staking` (or a interface with a module able to return a validator set) as dependency of the `monitoringp` module
- In `begin_block.go`, retrieve the operator address from the consensus address from the current validator set to update the signature counts

PR #579, opened 2022-03-01
refactor: simplify test keeper functions

aljo242:
Closes ##498

### What does this PR does?

Changes the `testutil/keeper` function `AllKeepers()` to: 
```golang
func NewTestSetup(t testing.TB) (sdk.Context, TestKeepers, TestMsgServers)
```

The `TestKeepers` struct just contains all of the keepers:
```golang
type TestKeepers struct {
	CampaignKeeper           *campaignkeeper.Keeper
	LaunchKeeper             *launchkeeper.Keeper
	ProfileKeeper            *profilekeeper.Keeper
	RewardKeeper             *rewardkeeper.Keeper
	MonitoringConsumerKeeper *monitoringcmodulekeeper.Keeper
	BankKeeper               bankkeeper.Keeper
	IBCKeeper                *ibckeeper.Keeper
}
```

The `TestMsgServers` struct just contains all of the message servers:
```golang
type TestMsgServers struct {
	ProfileSrv     profiletypes.MsgServer
	LaunchSrv      launchtypes.MsgServer
	CampaignSrv    campaigntypes.MsgServer
	RewardSrv      rewardtypes.MsgServer
	MonitoringcSrv monitoringcmoduletypes.MsgServer
}
```

The remaining changes apply these changes throughout our tests.

PR #588, opened 2022-03-02
refactor(`profile`): change consensus address list into operator address list

lubtd:
Part of https://github.com/tendermint/spn/issues/552

Make changes in `profile` module to support operator address instead of consensus address
`MsgSetConsAddress` is renamed into `MsgAddOperatorAddress`
Since we can now use operator bech32 address for signature verification, we no longer need on-chain nonce tracking, therefore `ConsensusNonce` has been removed
We currently only check Operator Address for the signature. Later we should require both signature from Operator Address and Validator Address. This requires to investigate on double signatures, which should be handled in a further PR

NOTES: the changes in `reward` are just temporary workaround. https://github.com/tendermint/spn/pull/577 will have to be merged to fix the reward errors

USAGE
```
spnd tx profile add-validator-operator-address [validator-address] --from [operator-address-owner]
```
Operator addresses are stored with normal account format prefixed by `spn` the resolution from the monitoring packet is perform when distributing rewards


x/profile/keeper/validator_by_operator_address_test.go
```diff
@@ -0,0 +1,60 @@
...
+		)
+		require.True(t, found)
+		require.Equal(t,
+			nullify.Fill(&item),
+			nullify.Fill(&rst),
+		)
+	}
+}
+func TestValidatorByOperatorAddressRemove(t *testing.T) {
+	k, ctx := keepertest.Profile(t)
```
aljo242, line 40:
```suggestion
	ctx, tk, _ := testkeeper.NewTestSetup(t)
```

These lines will need to use the new signature introduced in [PR #579](https://github.com/tendermint/spn/pull/579)

lubtd:
Seems like I can't add you as Reviewer @aljo242

lubtd:
Merging now, full PR will be on `refactor/consensus-to-operator-address`

Issue #587, opened 2022-03-02
Remove `MsgDeleteValidator` message


lubtd:
We should not remove validator information for documentation purpose. We should remove this message

PR #592, opened 2022-03-03
refactor(`profile`): remove `MsgDeleteValidator`

lubtd:
Solve https://github.com/tendermint/spn/issues/587


x/profile/types/errors.go
```diff
@@ -12,7 +12,6 @@ var (
 	ErrCoordAddressNotFound      = sdkerrors.Register(ModuleName, 3, "coordinator address not found")
 	ErrCoordInvalid              = sdkerrors.Register(ModuleName, 4, "invalid coordinator")
 	ErrEmptyDescription          = sdkerrors.Register(ModuleName, 5, "you must provide at least one description parameter")
-	ErrValidatorNotFound         = sdkerrors.Register(ModuleName, 6, "validator address not found")
 	ErrInvalidValidatorSignature = sdkerrors.Register(ModuleName, 7, "invalid validator signature")
```
aljo242, line 15: We should ensure the codespace numbers are going in ascending order.  Now that `ErrValidatorNotFound` is removed, we jump from 5 to 7

aljo242 (changes requested):

Pantani (dismissed):

x/profile/types/errors.go
lubtd: Thanks. I don't think it is a problem technically speaking but that's better for consistency

go.mod
```diff
@@ -18,8 +18,9 @@ require (
 	github.com/tendermint/starport v0.19.3
 	github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db v0.6.4
-	google.golang.org/genproto v0.0.0-20220118154757-00ab72f36ad5
-	google.golang.org/grpc v1.43.0
+	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
```
giunatale, line 21:
```suggestion

```

giunatale (approved):

PR #593, opened 2022-03-03
refactor: remove uint64 to int64 conversions

### LINKED ISSUE #580: Search for all `uint64` to `int64` conversions and remove them
`uint64` to `int64` is dangerous
We may possibly have some because of Unix timestamp conversion
We must remove them

aljo242:
Closes #580

### What does this PR does?

Removes the `uint64` -> `int64` conversions that I could find in our codebase.  Also changes all `BlockHeight` variables to `int64`.  This was the source of some of the conversions and makes sense because the Cosmos SDK itself stores blockheight as int64.


Pantani (dismissed):
LGMT

proto/reward/reward_pool.proto
```diff
@@ -19,7 +19,7 @@ message RewardPool {
     (gogoproto.casttype) = "github.com/cosmos/cosmos-sdk/types.Coin",
     (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
   ];
-  uint64 lastRewardHeight = 5;
-  uint64 currentRewardHeight = 6;
+  int64 lastRewardHeight = 5;
+  int64 currentRewardHeight = 6;
```
lubtd, line 23: We have a `Validate` method in `RewardPool`, we should check the values are non-negative

lubtd (changes requested):

go.mod
```diff
@@ -18,8 +18,9 @@ require (
 	github.com/tendermint/starport v0.19.3
 	github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db v0.6.4
-	google.golang.org/genproto v0.0.0-20220118154757-00ab72f36ad5
-	google.golang.org/grpc v1.43.0
+	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
```
giunatale, line 21:
```suggestion

```

giunatale (changes requested):
agreed with @lubtd on the non-negativity checks

x/reward/types/reward_pool.go
```diff
@@ -3,7 +3,6 @@ package types
 import (
 	"errors"
 	"fmt"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
```
giunatale, line 6: goimports

giunatale (changes requested):

lubtd (approved):

PR #598, opened 2022-03-04
feat(`launch`): Monitoring connected check to prevent MsgRevertLaunch

### LINKED ISSUE #491: Prevent `MsgRevertLaunch` when the monitoring IBC connection has been established
The monitoring IBC connection being established proves the chain has been launched, therefore `MsgRevertLaunch`  must fails and the launch can never be reverted

Two approaches can possibly be considered:
- Adding `monitoringc` has a dependency to `launch` so the `launch` module can detect a chain being connected to `spn` from the value `providerClientId` (official client ID from a launch ID for the IBC connection once established)
- Adding a boolean `isLaunch` or `isConnected` in `launch` in `Chain` struct that is set by `monitoringc` when the connection is established. `isLaunch` can be confusing since not all chains connect to `spn`

aljo242:
Closes #491

### What does this PR does?

Adds an `IBCConnected` bool to the `Chain` in `Launch`.  This is set to true in `monitoringc/keeper/msg_server_create_client.go` and is checked in `launch/keeper/msg_server_revert_launch.go`.

Currently I do not have a check in the `CreateClient` logic that throws an error if the chain is already IBC connected (essentially preventing a client from being created twice in a row).  Does this make sense to add?  Are there any other regions in the code where adding this check would make sense?


x/launch/keeper/validator_set.go
```diff
@@ -25,6 +25,9 @@ func (k Keeper) CheckValidatorSet(
 	}
 	if !chain.LaunchTriggered {
 		return sdkerrors.Wrapf(types.ErrNotTriggeredLaunch, "%d", launchID)
+	}
+	if chain.IbcConnected {
+
```
giunatale, line 30:
```suggestion
	if chain.IbcConnected {
                    return sdkerrors.Wrapf(types.ErrChainIBCConnected, "%d", launchID)
```

giunatale (changes requested):

giunatale (changes requested):
should do a fmt pass on the file you changed

also @lubtd what do you think of his last question?

x/launch/types/errors.go
```diff
@@ -38,4 +38,5 @@ var (
 	ErrChainHasCampaign            = sdkerrors.Register(ModuleName, 29, "chain already is associated with a campaign")
 	ErrAddChainToCampaign          = sdkerrors.Register(ModuleName, 30, "unable to add chain to campaign")
 	ErrRemainingTimeNotPositive    = sdkerrors.Register(ModuleName, 31, "invalid remaining time")
+	ErrChainIBCConnected           = sdkerrors.Register(ModuleName, 32, "chain is already IBC connected")
```
lubtd, line 41: Same comment as above for the error name. Should specify `monitoring` is concerned

x/monitoringc/keeper/msg_server_create_client.go
```diff
@@ -72,6 +72,10 @@ func (k msgServer) CreateClient(goCtx context.Context, msg *types.MsgCreateClien
 		ClientID: clientID,
 	})
 
+	// set chain to be IBC connected
+	chain.IbcConnected = true
+	k.launchKeeper.SetChain(ctx, chain)
```
lubtd, line 77: I would prefer having `k.launchKeeper.EnableMonitoringConnection(ctx, chainID)` or `MonitoringConnected(...)` to prevent the module to directly write into the store to keep some kind of encapsulation

There would be a duplicated read in the store for the `chainID` but it's in my opinion not significant since establishing a connection to a testnet is already an extensive operation (with relayers etc...)

lubtd (dismissed):
`MsgCreateClient` is a message to create a verified IBC client for the testnet on `spn` but at this point the IBC connection is not yet established

This verified client is then used for the IBC connection (or channel creation) from the relayer.
Thefore the logic of setting the chain as connection to `true` should be in the `RegisterProviderClientIDFromChannelID` method of `handshake.go`. This method is called in `OnChanOpenAck` IBC phase when the handshake is successful

lubtd:
> Currently I do not have a check in the CreateClient logic that throws an error if the chain is already IBC connected (essentially preventing a client from being created twice in a row). Does this make sense to add? Are there any other regions in the code where adding this check would make sense?

The connection can't be established twice because once established `ProviderClientID` is set and once this value is set all further handshakes will fail
When a client with `MsgCreateClient` is created, the client can be created with incorrect values for the IBC client state (unbonding period, etc...) so we don't prevent creating several because only when the handshake is perform the client is fully verified.
We can eventually make `MsgCreateClient` fails once `ProviderClientID` is set, but in any case all clients created at this point will not be usable to establish a new connection

Pantani:
> Currently I do not have a check in the `CreateClient` logic that throws an error if the chain is already IBC connected (essentially preventing a client from being created twice in a row). Does this make sense to add? Are there any other regions in the code where adding this check would make sense?

I think it makes total sense to add.

Pantani (approved):

x/launch/keeper/chain.go
```diff
@@ -125,6 +125,18 @@ func (k Keeper) SetChain(ctx sdk.Context, chain types.Chain) {
...
 }
 
+// EnableMonitoringConnection sets a chain with MonitoringConnected set to true
+func (k Keeper) EnableMonitoringConnection(ctx sdk.Context, launchID uint64) error {
+	chain, found := k.GetChain(ctx, launchID)
+	if !found {
+		return types.ErrChainNotFound
+	}
+
+	chain.MonitoringConnected = true
```
lubtd, line 135: I think we can return an error as well when already connected

x/launch/keeper/chain_test.go
```diff
@@ -237,6 +237,24 @@ func TestGetChain(t *testing.T) {
...
+
+	items := createNChain(tk.LaunchKeeper, ctx, 10)
+	for _, item := range items {
+		err = tk.LaunchKeeper.EnableMonitoringConnection(ctx, item.LaunchID)
+		require.NoError(t, err)
+		rst, found := tk.LaunchKeeper.GetChain(ctx, item.LaunchID)
+		require.True(t, found)
+		item.MonitoringConnected = true
+		require.Equal(t, item, rst)
+	}
```
lubtd, line 255: I think creating a single chain is enough

giunatale (approved):

lubtd (approved):
👍

One thing: we can currently actually establish the IBC connection after the launch is reverted if the verified client has been created. The handshake doesn't verify if `LaunchTriggered` is set
But the issue is a little bit more complex, I put all details here: https://github.com/tendermint/spn/issues/607
Approving since this should be handled in a separate PR

PR #611, opened 2022-03-09
feat(`participation`): Add `AllocationPrice` and `ParticipationTierList` params

### LINKED ISSUE #572: Add `AllocationPrice` and `ParticipationTierList` params
Add `AllocationPrice` and `ParticipationTierList` params with validatation methods

`AllocationPrice` can be discussed, we need to only determine the condition to get allocations from staking for now

Depends on https://github.com/tendermint/spn/issues/569

giunatale:
Closes #572


x/participation/types/params.go
```diff
@@ -1,12 +1,14 @@
 package types
 
 import (
+	"errors"
 	"fmt"
 
+	"gopkg.in/yaml.v2"
+
```
lubtd, line 8:
```suggestion
```

lubtd (commented):
Just a tiny comment
I will do a more through review tomorrow

x/participation/types/params.go
```diff
@@ -107,8 +132,41 @@ func validateParticipationTierList(v interface{}) error {
 		return fmt.Errorf("invalid parameter type: %T", v)
 	}
 
-	// TODO implement validation
-	_ = participationTierList
+	tiersIndexMap := make(map[uint64]struct{})
+	lastTierID := uint64(0)
+	lastRequiredAllocations := uint64(0)
+	lastTierBenefits := TierBenefits{MaxBidAmount: sdk.ZeroInt()}
+	for _, tier := range participationTierList {
```
aljo242, line 139: So if a TierList is a []Tier and must be sorted, do we really need the TierID?  Couldn't we just use the index in the slice as the ID?

lubtd: In my opinion, it gives more flexibility and make it a bit more reliable to explicitly define IDs
We can add a new tier in between without modifying the ID of the other tiers. In the front-end the tiers would be sorted by the required staking amount

testutil/sample/participation.go
```diff
@@ -1,11 +1,42 @@
...
+			},
+		}
+		tiers = append(tiers, tier)
+
+		// increment values for next tier
+		allocCnt += uint64(rand.Int63n(5) + 1)
+		maxBidCnt = maxBidCnt.AddRaw(rand.Int63n(10000) + 1)
+	}
+
+	// TODO: randomize param generation (?)
```
lubtd, line 36: Randomization would be good for the simulation test since we need entropy

`sample` is used for unit test, we use randomness for some method like `String` but this is not an important requirement. Unit tests we test deterministic `input -> expected outpout`

x/participation/module_simulation.go
```diff
@@ -47,7 +47,7 @@ func (AppModule) ProposalContents(_ module.SimulationState) []simtypes.WeightedP
 
 // RandomizedParams creates randomized  param changes for the simulator
 func (am AppModule) RandomizedParams(_ *rand.Rand) []simtypes.ParamChange {
-	participationParams := types.DefaultParams()
+	participationParams := sample.ParticipationParams()
```
lubtd, line 50: See my comment above, randomness would be good for this part but not blocking for this PR

x/participation/types/params.go
```diff
@@ -94,8 +118,9 @@ func validateAllocationPrice(v interface{}) error {
 		return fmt.Errorf("invalid parameter type: %T", v)
 	}
 
-	// TODO implement validation
-	_ = allocationPrice
+	if allocationPrice.Bonded.LTE(sdk.ZeroInt()) {
```
lubtd, line 121: We should also have `!allocationPrice.Bonded.IsNil()` since `Int` is a pointer

lubtd, line 121:
```suggestion
	if !allocationPrice.Bonded.IsPositive() {
```

x/participation/types/params.go
```diff
@@ -107,8 +132,41 @@ func validateParticipationTierList(v interface{}) error {
...
+
+		if tier.RequiredAllocations <= lastRequiredAllocations {
+			return errors.New("required allocations must be strictly increasing and greater than zero")
+		}
+		lastRequiredAllocations = tier.RequiredAllocations
+
+		if err := validateNextTierBenefits(tier.Benefits, lastTierBenefits); err != nil {
+			return err
+		}
+		lastTierBenefits = tier.Benefits
```
lubtd, line 160: In my opinion, this shouldn't be a requirement. We can manage the sorting in the frontend, using `TierID` to represent the benefits make the structure more flexible

x/participation/types/params.go
giunatale: fair question. this is up for debate, as well as the validation logic I implemented. happy to discuss possible refinements.

giunatale: agreed with @lubtd

testutil/sample/participation.go
giunatale: yeah, but for `withdrawalAllocationDelay` we have yet to investigate it (and is left out of this PR) hence why the todo comment. the other params I've made sure they are sufficiently randomized.

x/participation/types/params.go
giunatale: I am not sure about that. if we don't do this here we could end up with a lot of bloat tiers that would hurt performance of the dynamic computation of allocations. enforcing this mitigates the problem (since tiers are forced to be somewhat "meaningful")

x/participation/module_simulation.go
giunatale: I think there's a misunderstanding. params addressed in this PR *are* randomized

testutil/sample/participation.go
giunatale: marking as resolved since I think this point was actually addressed

x/participation/module_simulation.go
giunatale: same as above

x/participation/types/params.go
lubtd: > that would hurt performance of the dynamic computation of allocations

I don't get this part. Computing the available allocation would be fully performed in my mind before checking the tier list. You computed the total available allocation and then you check if the value is higher than the necessary allocation for the tier you want to benefit from

x/participation/types/params_test.go
```diff
@@ -62,7 +62,12 @@ func TestValidateAllocationPrice(t *testing.T) {
 		},
 		{
 			name:            "bonded amount lower or equal than zero",
```
lubtd, line 64: "uninintialized bonded amount"

x/participation/types/params_test.go
```diff
@@ -205,6 +210,11 @@ func TestValidateNextTierBenefits(t *testing.T) {
 		tierBenefits TierBenefits
 		err          error
 	}{
+		{
+			name:         "bonded amount lower or equal than compared benefits",
```
lubtd, line 214: I make think the test case name should be renamed for this one and the other one below. We mention "bonded" while `MaxBidAmount` values are asserted

giunatale: thanks for spotting this, was a copy-paste mistake

x/participation/types/params_test.go
giunatale: same as below, copy paste error. will fix

x/participation/types/params.go
giunatale: fair, I think I haven't fully thought it through when I replied. The tier is provided in `MsgParticipate` so what I said is plain wrong.
I still think not enforcing any constraint on the tier list is less elegant, but quite frankly I have no strong opinion. 
So I'm open to removing these checks.

@aljo242 @Pantani what's your opinion?

aljo242: I agree with @lubtd as long as the front end sorts these things.  My main reasoning is that you get a bit more flexibility to add "in between" tiers

x/participation/types/params.go
```diff
@@ -107,8 +135,34 @@ func validateParticipationTierList(v interface{}) error {
...
+
+		if err := validateNextTierBenefits(tier.Benefits); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func validateNextTierBenefits(b TierBenefits) error {
```
lubtd, line 158:
```suggestion
func validateTierBenefits(b TierBenefits) error {
```

Mmh why "next"?

lubtd (approved):

PR #614, opened 2022-03-10
feat(`participation`): add invariant to check mismatch between `AuctionUsedAllocations` and `UsedAllocations`

### LINKED ISSUE #613: Add invariant for consistency between `AuctionUsedAllocations` and `UsedAllocations`
We should add an invariant that checks whether the sum of all used per-auction allocations for an address in `AuctionUsedAllocations` is equal to the value in `UsedAllocations`

giunatale:
Closes #613

### DIFF
--- a/x/participation/keeper/invariants_test.go
+++ b/x/participation/keeper/invariants_test.go
@@ -1,0 +1,57 @@
+package keeper_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	testkeeper "github.com/tendermint/spn/testutil/keeper"
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/participation/keeper"
+	"github.com/tendermint/spn/x/participation/types"
+)
+
+func TestMismatchUsedAllocationsInvariant(t *testing.T) {
+	var (
+		ctx, tk, _        = testkeeper.NewTestSetup(t)
+		addr              = sample.Address()
+		auctionUsedAllocs = []types.AuctionUsedAllocations{
+			{
+				Address:        addr,
+				AuctionID:      1,
+				NumAllocations: 1,
+			},
+			{
+				Address:        addr,
+				AuctionID:      2,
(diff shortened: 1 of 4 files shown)


x/participation/keeper/invariants_test.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	)
+
+	t.Run("valid case", func(t *testing.T) {
+		tk.ParticipationKeeper.SetUsedAllocations(ctx, validUsedAllocs)
+		for _, auction := range auctionUsedAllocs {
+			tk.ParticipationKeeper.SetAuctionUsedAllocations(ctx, auction)
+		}
+		_, isValid := keeper.MismatchUsedAllocationsInvariant(*tk.ParticipationKeeper)(ctx)
+		require.Equal(t, false, isValid)
```
aljo242, line 46:
```suggestion
		require.False(t, isValid)
```

x/participation/keeper/invariants_test.go
```diff
@@ -0,0 +1,57 @@
...
+	})
+
+	t.Run("invalid case", func(t *testing.T) {
+		tk.ParticipationKeeper.SetUsedAllocations(ctx, invalidUsedAllocs)
+		for _, auction := range auctionUsedAllocs {
+			tk.ParticipationKeeper.SetAuctionUsedAllocations(ctx, auction)
+		}
+		_, isValid := keeper.MismatchUsedAllocationsInvariant(*tk.ParticipationKeeper)(ctx)
+		require.Equal(t, true, isValid)
```
aljo242, line 55:
```suggestion
		require.True(t, isValid)
```

aljo242 (changes requested):

aljo242:
Looks good! Just tiny nitpicks

x/participation/keeper/invariants_test.go
giunatale: fyi, this is the same as per other invariant tests in other modules, but worth perhaps fixing it here

PR #615, opened 2022-03-10
refactor(`MsgAddValidatorOperatorAddress`): requires two signatures

lubtd:
`MsgAddValidatorOperatorAddress`, when `ValidatorAddress` and `OperatorAddress` are different, the message must be signed with both addresses

To test:

`spnd tx profile add-validator-operator-address` CLI command has been modified.
Now the comment has no positional argument. By default, `--from` is used as the Operator Address.
The flag `--operator-address` allows to specify a different address for the operator. If the address is not found in the keyring the message can't be signed and will fail
```
spnd tx profile add-validator-operator-address --from bob --sign-mode amino-json
spnd tx profile add-validator-operator-address --operator-address spn1hmx8eakt2948szjgmksvpv9ha0q9s6w08ppedf --from alice --sign-mode amino-json
```

Providing ` --sign-mode amino-json` is mandatory for double signing


aljo242 (approved):
Works well !

x/profile/client/cli/tx_add_validator_operator_address.go
```diff
@@ -9,29 +11,96 @@ import (
...
+			if err != nil {
+				return err
+			}
+
+			// broadcast tx
+			resp, err := clientCtx.BroadcastTxSync(encodedTx)
+			if err != nil {
+				return err
+			}
+			fmt.Println(resp.String())
```
Pantani, line 73:
```suggestion
			fmt.Print(resp.String())
```

Pantani (dismissed):

PR #616, opened 2022-03-10
feat(`launch`): remove verified client IDs in `monitoringc`

### LINKED ISSUE #607: Reverting launch: remove all verified client IDs in `monitoring-consumer` for the launch ID
`revert-launch` allows to roll back the launch triggering of the chain if the launch time has been reached and the network failed to start for a misc reason.

`revert-launch` can be called by the coordinator once the revert delay has been reached and the monitoring connection (for a testnet) is not yet established (with https://github.com/tendermint/spn/pull/598)
Monitoring connection being established is the only on-chain event that allows detecting the launch of the chain on `spn`. In practice if the connection is not established the launch can be reverted even if the network started.
There is an issue related to this: a IBC verified client can be created with `MsgCreateClient` when the network started and before the launch is reverted
Then:
- The IBC connection can be established with this client after the launch is reverted, meaning that `spn` will be connected to a chain specified as "not launched" (IBC handshake currently doesn't check if launch is triggered)
- If the chain restart after validator set changes through `spn`, the previous client can be reused for the IBC connection. Meaning that `spn` can connect with a client that has not been verified

To handle these two issues, we should remove the currently verified client IDs in `monitoring-consumer` for the launch when the launch is reverted, if any exist

- We should implement a method `ClearVerifiedClientIDs(ctx, launchID)` that deletes the `VerifiedClientID` entry for the launch ID in `monitoring-consumer` keeper
- Call this method in `launch` during `MsgRevertLaunch`

This is an additional protection, in practice `RevertDelay` should let way enough time to create the verified client ID and perform the monitoring connection when the network starts and the coordinator is not deincentivized to revert the launch if the network started

aljo242:
Closes #607

### What does this PR does?

- Add `ClearVerifiedClientIDs()` method
- Add `monitoringc` dependency to `launch` keeper
- Add tests

## Checklist

review them:

- [X] Include tests


lubtd:
Looks good, testing right now

lubtd:
I can create the connection with a non-verified IBC client. I don't think this is related to this PR but I need to investigate more

x/monitoringc/keeper/verified_client_id.go
```diff
@@ -14,6 +14,12 @@ func (k Keeper) SetVerifiedClientID(ctx sdk.Context, verifiedClientID types.Veri
 	store.Set(types.VerifiedClientIDKey(verifiedClientID.LaunchID), b)
 }
 
+// ClearVerifiedClientIDs removes a set of verifiedClientID in the store from its launch ID
+func (k Keeper) ClearVerifiedClientIDs(ctx sdk.Context, launchID uint64) {
```
lubtd, line 18: We must also remove all entries of `LaunchIDFromVerifiedClientID` for each of the client IDs in the array. This is the actual entry read when doing the handshake, hence the error I mentioned above

lubtd (changes requested):

lubtd (dismissed):
👍

x/launch/keeper/msg_revert_launch.go
```diff
@@ -55,5 +55,8 @@ func (k msgServer) RevertLaunch(goCtx context.Context, msg *types.MsgRevertLaunc
 	chain.LaunchTimestamp = 0
 	k.SetChain(ctx, chain)
 
+	// clear associated client IDs from monitoring
+	k.monitoringcKeeper.ClearVerifiedClientIDs(ctx, msg.LaunchID)
```
giunatale, line 59: linting says error is not checked. is it intentional?

giunatale (changes requested):

giunatale (dismissed):

lubtd:
> @giunatale I returned an error from that function. Error conditions from ClearVerifiedClientIDs() are already tested in the monitoringc package so I did not add an extra test.

We may not have created a verified client before reverting (it will even always be the case in practice) therefore we should either only call `ClearVerifiedClientIDs ` if verified client IDs exist or filter out the not found error

(or maybe simply remove this error, the clear has no effect if no verified client IDs)

aljo242:
I think removing the error makes the most sense as that's how the other keeper functions work.

Pantani (approved):

PR #624, opened 2022-03-15
feat(`launch`): split `MsgEditChain`

### LINKED ISSUE #623: Split `MsgEditChain` into two messages
We currently have in `MsgEditChain`, fields to be edited of two types:
- Information not involved in the logic of genesis generation (campaign ID, metadata)
- Information used for the genesis generation (sources, initial genesis, chain ID)

`MsgEditChain` should be splitted.
The former type of information can be editable at any time.
We should not be able to modify the later after the chain launch has been triggered

Proposed solution:
- Keep `MsgEditChain` name and current behavior for metadata and campaign ID, it keeps some consistency with `MsgEditCampaign` that has the same idea with campaign name and metadata
- Add a new message: ex `MsgEditSourceInformation` that modify source hash, URL, initial genesis and chain ID. Prevent updating after trigger launch

Any idea is welcome for message names ideas

aljo242:
Closes #623

### What does this PR does?

- Add new tx `MsgEditChainSourceInformation` which edits source info
- modifies `MsgEditChain` to only modify campaignID and metadata


lubtd:
Mmh we should address the issue with simulation tests

aljo242:
> Mmh we should address the issue with simulation tests

ah, adding the check for `launchTriggered` broke the sims.  Fixed now :)

x/launch/keeper/msg_edit_chain_source_information.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, sdkerrors.Wrapf(types.ErrChainNotFound, "%d", msg.LaunchID)
+	}
+
+	if chain.LaunchTriggered {
+		return nil, sdkerrors.Wrapf(types.ErrTriggeredLaunch, "%d", msg.LaunchID)
+	}
+
+	if chain.MonitoringConnected {
+		return nil, sdkerrors.Wrapf(types.ErrChainMonitoringConnected, "%d", msg.LaunchID)
+	}
```
lubtd, line 32: `MonitoringConnected` should imply `LaunchTriggered`. This can be removed

lubtd (commented):
Looks legit, doing some experimentation

There is the naming that remains tricky.
Having `MsgEditChain` and `MsgEditChainSourceInformation` can be confusing.
I would rename `MsgEditChainSourceInformation` into `MsgUpdateSourceInformation` but we also have `chain-id` value that can be updated that is not really a source information, we should have it in the message because modification should be prevented after launch triggering.

Since information that can't be changed after trigger are related to launch. I would suggest a message name like `MsgUpdateLaunchInformation`
`MsgEditChain` edits miscellaneous static data related to the chain be not influence the launch
`MsgUpdateLaunchInformation` updates information that are actively fetched for the launch with genesis

x/launch/client/cli/tx_update_launch_information.go
```diff
@@ -0,0 +1,86 @@
...
+	flagGenesisChainID = "genesis-chain-id"
+	flagSourceURL      = "source-url"
+	flagSourceHash     = "source-hash"
+	flagDefaultGenesis = "default-genesis"
+)
+
+func CmdUpdateLaunchInformation() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update-launch-information [launch-id]",
+		Short: "Update chain genesis information",
```
lubtd, line 25:
```suggestion
		Short: "Update initial information for the launch of the chain",
```

lubtd (dismissed):

x/launch/keeper/msg_update_launch_information_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"testing"
+
+	testkeeper "github.com/tendermint/spn/testutil/keeper"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+
+	"github.com/tendermint/spn/testutil/sample"
+	"github.com/tendermint/spn/x/launch/types"
+	profiletypes "github.com/tendermint/spn/x/profile/types"
```
Pantani, line 13:
```suggestion
	sdk "github.com/cosmos/cosmos-sdk/types"
	"github.com/stretchr/testify/require"

	testkeeper "github.com/tendermint/spn/testutil/keeper"
	"github.com/tendermint/spn/testutil/sample"
	"github.com/tendermint/spn/x/launch/types"
	profiletypes "github.com/tendermint/spn/x/profile/types"
```

x/launch/client/cli/tx.go
```diff
@@ -29,6 +29,7 @@ func GetTxCmd() *cobra.Command {
 	cmd.AddCommand(CmdSettleRequest())
 	cmd.AddCommand(CmdTriggerLaunch())
 	cmd.AddCommand(CmdRevertLaunch())
+	cmd.AddCommand(CmdUpdateLaunchInformation())
```
Pantani, line 32:
```suggestion
	cmd.AddCommand(
		CmdCreateChain(),
		CmdEditChain(),
		CmdRequestAddAccount(),
		CmdRequestAddVestingAccount(),
		CmdRequestRemoveAccount(),
		CmdRequestAddValidator(),
		CmdRequestRemoveValidator(),
		CmdSettleRequest(),
		CmdTriggerLaunch(),
		CmdRevertLaunch(),
		CmdUpdateLaunchInformation(),
	)
```

lubtd:
Hey @Pantani, do you approve the PR outside of the two suggestions you made?
If yes, we could apply the suggestions to then merge the PR since @aljo242 is OOO end of this week

Pantani (approved):

x/launch/client/cli/tx.go
lubtd: Fine, I just changed the order of txs

giunatale (approved):

PR #630, opened 2022-03-16
feat(`participation`): full `MsgParticipate` functionality 

### LINKED ISSUE #575: Implement full message `MsgParticipate`
Depends on https://github.com/tendermint/spn/pull/619

Implement `MsgParticipate` that allows an account to add itself as a bidder for an auction and update the used allocation

### LINKED ISSUE #621: Create a `AvailableAllocation` query
Depends on https://github.com/tendermint/spn/pull/620

A query `AvaiblableAllocation` allows to directly fetch from an address, the available allocation to be used (total - used).
This is convenient to choose a tier for an auction

aljo242:
Closes #621
Closes #575


x/participation/client/cli/query_available_allocations.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cobra"
+
+	"github.com/tendermint/spn/x/participation/types"
+)
+
+var _ = strconv.Itoa(0)
+
```
lubtd, line 14:
```suggestion
```

x/participation/keeper/available_allocations.go
```diff
@@ -0,0 +1,22 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	usedAlloc, found := k.GetUsedAllocations(ctx, address)
+	if !found {
+		return numTotalAlloc, nil
+	}
+
+	availableAlloc := numTotalAlloc - usedAlloc.NumAllocations
```
lubtd, line 19: `usedAlloc.NumAllocations` can eventually be bigger than `numTotalAlloc` in the case the user unbonding tokens after having used allocations.

We should return `0` if negative

x/participation/keeper/available_allocations_test.go
```diff
@@ -0,0 +1,76 @@
...
+		AllocationPrice: allocationPrice,
+	})
+
+	validAddress := sample.Address()
+	validAddressNoUse := sample.Address()
+	addressNegativeDelegations := sample.Address()
+
+	tk.DelegateN(sdkCtx, validAddress, 100, 10)
+	tk.DelegateN(sdkCtx, validAddressNoUse, 100, 10)
+	tk.DelegateN(sdkCtx, addressNegativeDelegations, -100, 10)
```
lubtd, line 31: Is a negative delegation possible by design?

x/participation/keeper/available_allocations_test.go
```diff
@@ -0,0 +1,76 @@
...
+		NumAllocations: 2,
+	})
+
+	for _, tc := range []struct {
+		desc       string
+		address    string
+		allocation uint64
+		wantError  bool
+	}{
+		{
```
lubtd, line 44: Based on the comment above, we should have a test case where `totalAllocation < usedAllocation` and it returns 0

x/participation/keeper/msg_participate.go
```diff
@@ -13,16 +13,41 @@ import (
...
+		return nil, err
+	}
 
+	// check if auction exists
 	_, found := k.fundraisingKeeper.GetAuction(ctx, msg.AuctionID)
 	if !found {
 		return nil, sdkerrors.Wrapf(types.ErrAuctionNotFound, "auction %d not found", msg.AuctionID)
 	}
 
+	// check if a bid already exists
```
lubtd, line 27:
```suggestion
	// check if the user is already added as an allowed bidder for the auction 
```

x/participation/keeper/msg_participate.go
```diff
@@ -31,5 +56,21 @@ func (k msgServer) Participate(goCtx context.Context, msg *types.MsgParticipate)
 		return nil, sdkerrors.Wrap(types.ErrInvalidBidder, err.Error())
 	}
 
+	// set used allocations
+	usedAllocations, _ := k.GetUsedAllocations(ctx, msg.Participant)
```
lubtd, line 60: Does it returns an empty `usedAllocations` when not found?

I would prefer to explicitly initialize the `usedAllocations` structure if `GetUsedAllocations` returns `found==false` even if it makes the code a bit more complex, we are supposed to ignore `usedAllocations` value if `found==false`

x/participation/keeper/msg_participate_test.go
```diff
@@ -34,23 +41,72 @@ func Test_msgServer_Participate(t *testing.T) {
...
+			msg: &types.MsgParticipate{
+				Participant: addrWithDels,
+				AuctionID:   auctionID,
+				TierID:      1,
+			},
+			desiredUsedAlloc:      1,
+			currentAvailableAlloc: availableAlloc,
+		},
+		{
+			name: "should prevent if double bid from same address",
```
lubtd, line 68:
```suggestion
			name: "should prevent participating twice in the same auction",
```
There is no bid before auction start

x/participation/keeper/msg_participate_test.go
```diff
@@ -34,23 +41,72 @@ func Test_msgServer_Participate(t *testing.T) {
...
+		},
+		{
+			name: "should prevent if address is invalid",
+			msg: &types.MsgParticipate{
+				Participant: "invalid",
+				AuctionID:   auctionID,
+				TierID:      1,
+			},
+			untypedErr: true,
+			err:        errors.New("decoding bech32 failed: invalid bech32 string length 7"),
```
lubtd, line 84: We don't need a testcase for invalid address, invalid address must always be handled in the message static `Validate` method

Also I would prefer not asserting exact error because it can be hard to maintain even more when the error is not returned from our codebase in this case

x/participation/keeper/msg_participate.go
```diff
@@ -13,16 +13,41 @@ import (
...
+	// check if auction exists
 	_, found := k.fundraisingKeeper.GetAuction(ctx, msg.AuctionID)
 	if !found {
 		return nil, sdkerrors.Wrapf(types.ErrAuctionNotFound, "auction %d not found", msg.AuctionID)
 	}
 
+	// check if a bid already exists
+	_, found = k.GetAuctionUsedAllocations(ctx, msg.Participant, msg.AuctionID)
+	if found {
+		return nil, sdkerrors.Wrapf(types.ErrInvalidBidder,
```
lubtd, line 30: `types.ErrInvalidBidder` we should not consider the notion of `Bidder` at this point

Also the error would be in my opinion more `ErrAlreadyParticipating`

x/participation/keeper/msg_participate_test.go
```diff
@@ -73,6 +132,22 @@ func Test_msgServer_Participate(t *testing.T) {
 				Bidder:       tt.msg.Participant,
 				MaxBidAmount: sdk.NewIntFromUint64(1000),
```
lubtd, line 133: We must change this with the tier `MaxBidAmount`, this is not normal if the tests pass at this point

x/participation/keeper/msg_participate_test.go
```diff
@@ -73,6 +132,22 @@ func Test_msgServer_Participate(t *testing.T) {
...
 			})
+
+			// check used allocations entry for bidder
+			usedAllocations, found := tk.ParticipationKeeper.GetUsedAllocations(sdkCtx, tt.msg.Participant)
+			require.True(t, found)
+			require.EqualValues(t, tt.desiredUsedAlloc, usedAllocations.NumAllocations)
+
+			// check auction used allocations entry for bidder exists
+			_, found = tk.ParticipationKeeper.GetAuctionUsedAllocations(sdkCtx, tt.msg.Participant, tt.msg.AuctionID)
+			require.True(t, found)
```
lubtd, line 143: We can also check, this is the required allocation of the tier

lubtd (dismissed):

x/participation/keeper/available_allocations_test.go
aljo242: No, not directly through the standard mechanisms of the `staking` module.  This would only be a protection against another module that uses the `staking` keeper to create some kind of incorrect state.

x/participation/keeper/msg_participate.go
aljo242: Yeah it returns an empty struct and since we are setting both fields, the effect is the same as initializing the empty struct.  I'll refactor to make it more readable though

x/participation/keeper/msg_participate.go
```diff
@@ -13,16 +13,41 @@ import (
...
 	_, found := k.fundraisingKeeper.GetAuction(ctx, msg.AuctionID)
 	if !found {
 		return nil, sdkerrors.Wrapf(types.ErrAuctionNotFound, "auction %d not found", msg.AuctionID)
 	}
 
+	// check if the user is already added as an allowed bidder for the auction
+	_, found = k.GetAuctionUsedAllocations(ctx, msg.Participant, msg.AuctionID)
+	if found {
+		return nil, sdkerrors.Wrapf(types.ErrAlreadyParticipating,
+			"participant %s already has already bid in auction %d",
```
lubtd, line 31:
```suggestion
			"participant %s is already a participant for the auction %d",
```

lubtd (commented):
CLI works well👍

x/participation/keeper/msg_participate.go
giunatale:
```suggestion
			"address %s is already a participant for auction %d",
```

lubtd:
Yes, removed the last suggestion👍

giunatale (approved):

lubtd (approved):

PR #641, opened 2022-03-21
refactor(`monitoring`): remove debugMode params

### LINKED ISSUE #633: Removing `DebugMode` params
We have a `DebugMode` parameter for `monitoringp` and `monitoringc` modules.
This param when set to true on both chains, allows to perform an IBC connection without the need to setup the consumer IBC client in provider params and create a verified IBC client on the consumer through `MsgCreateClient` message.

We initially had this param for initial experimentation for testing reward distribution without initializing validators, fetching consensus state, etc....
Now we have local environment for automatizing steps above for experimentation.

I suggest to remove this param on both module. This is not really a parameter since it is not aimed to be modified by governance, `DebugMode` must always be set to `false` in practice

aljo242:
Closes #633

### What does this PR does?

Remove the `DebugMode` param from `monitoringc` and `monitoringp` modules and the logic using it.  Removes the associated tests that used it.


x/monitoringc/keeper/handshake.go
```diff
@@ -22,23 +22,20 @@ func (k Keeper) VerifyClientIDFromChannelID(ctx sdk.Context, channelID string) e
...
-		// check if the connection with the provider for this launch ID is already established
-		pCid, found := k.GetProviderClientID(ctx, lidFromCid.LaunchID)
-		if found {
-			return sdkerrors.Wrapf(
-				types.ErrConnectionAlreadyEstablished,
-				"provider client ID for launch ID %d is: %s",
-				pCid.LaunchID, pCid.ClientID,
-			)
-		}
+	// check if the client ID is verified
```
lubtd, line 25: We should also remove `DebugModeLaunchID` defined above

lubtd (commented):
LGTM, just one comment

We can also remove all mentions of `debugMode` in `localnet`. Removing the `debugMode` in `genesis_template`, in the config, etc...

giunatale (approved):

lubtd (dismissed):
We should remove `debugMode` setting up in `localnet/start.py`

lubtd:
I think the new compiled proto are not committed. After `starport c serve`, I get
```
	modified:   ../x/monitoringc/types/params.pb.go
	modified:   ../x/monitoringp/types/params.pb.go
```

lubtd:
We should remove this line https://github.com/tendermint/spn/blob/52055b93220c64c90959a25bce4b5dab81f41172/localnet/start.py#L39 @aljo242

lubtd:
I still get different compiled proto after build
```
	modified:   x/monitoringc/types/params.pb.go
	modified:   x/monitoringp/types/params.pb.go
```

aljo242:
> I still get different compiled proto after build
> 
> ```
> 	modified:   x/monitoringc/types/params.pb.go
> 	modified:   x/monitoringp/types/params.pb.go
> ```

Ok, I was super confused by what was happening but figured it out.  `make format` was modifying those `.pb.go` files (just the order of imports).  I was running it before pushing all of my commits.  I will make another PR to stop this from modifying `.pb.go` files as we shouldn't have any of our tools touch them.

lubtd:
Ok thanks for the investigation I think we can merge this one now

lubtd (approved):

Issue #636, opened 2022-03-18
Implement full `MsgWithdrawAllocations` message


giunatale:
With #576 we address the scaffolding of `msgWithdrawAllocations` without however accounting for the additional time check as discussed in #573.

We should add the time check and prevent withdrawal if the withdrawal delay after the auction start has not yet passed.

Depends on #576 and #573

lubtd:
`MsgParticipate` only during the Registration Period is not yet tracked, right?

giunatale:
> `MsgParticipate` only during the Registration Period is not yet tracked, right?

tracked in #635

Issue #638, opened 2022-03-18
Simulate `MsgParticpate`


aljo242:
Add simulation for this message

PR #643, opened 2022-03-22
feat(`participation`): add time-related checks in `msgParticipate`

### LINKED ISSUE #635: Add check for `registrationPeriod` param in `msgParticipate`
With #575 we address the full implementation of `msgParticipate` without however accounting for the additional time check as discussed in #573.

We should add the time check and prevent participation if not within the allowed registration period.

Depends on #575 and #573

giunatale:
Closes #635

### What does this PR does?
Add checks in `msgParticipate` to prevent participation in case:

1. auction has started
2. registration period not yet started

I've also switched to `time.Duration` for `registrationPeriod`, following the same as per `unbondingTime` in `staking`.
I think it makes it cleaner and prevents from error-prone conversions.

Will do the same for `withdrawalDelay` in #636


giunatale:
I think this might be relevant for #638

lubtd:
> I think this might be relevant for https://github.com/tendermint/spn/issues/638

It may gonna be tricky with the Block Time for simulation test. We would need to see if we can increment the block time automatically between each blocks during simulation if this is not the case

testutil/sample/participation.go
```diff
@@ -35,9 +35,10 @@ func ParticipationParams() participation.Params {
 	}
 
 	// generate a random time frame between an hour and four weeks for both params
+	// TODO clean up after switching withdrawamDelay to time.Duration
 	fourWeeks := int64(time.Hour.Seconds() * 24 * 7 * 4)
 	oneHour := int64(time.Hour.Seconds())
-	registrationPeriod := rand.Int63n(fourWeeks-oneHour) + oneHour
+	registrationPeriod := time.Duration(rand.Int63n(fourWeeks-oneHour)+oneHour) * time.Second
```
lubtd, line 41: What about a `sample.Duration()`?

x/participation/keeper/msg_participate.go
```diff
@@ -12,18 +12,31 @@ import (
...
 
+	// check if auction is already started
+	if auction.IsAuctionStarted(blockTime) {
+		return nil, sdkerrors.Wrapf(types.ErrParticipationNotAllowed, "auction %d is already started", msg.AuctionID)
+	}
+
+	// check if auction allows participation at this time
+	registrationPeriod := k.RegistrationPeriod(ctx)
+	// as commented in `Time.Sub()`: To compute t-d for a duration d, use t.Add(-d).
+	if !blockTime.After(auction.GetStartTime().Add(-registrationPeriod)) {
```
lubtd, line 36:
```suggestion
	// as commented in `Time.Sub()`: To compute t-d for a duration d, use t.Add(-d).
	registrationStart := auction.GetStartTime().Add(-registrationPeriod)
	if !blockTime.After(registrationStart) {
```

lubtd, line 36: I may think `StartTime` is arbitrarily chosen by the auction creator, we need to ensure there are no edge case where `Add(-registrationPeriod)` is sensitive

Example:  `StartTime` == Unix 0 time (January 1th 1970), does it give a valid time if subtracted by the registration period

We can eventually discuss if `StartTime` should be > current time for auction creation

giunatale: the check `auction.IsAuctionStarted(blockTime)` done before would make this practically impossible, since `blockTime` should be a valid Unix time much greater than 1970 (which basically means we ensure that `startTime` is also greater than that). 

Moreover, `time.Duration` can be negative (see `(t Time) Sub()` and `minDuration`) so I think we are safe here.

We have two ways of dealing with this:

1. truncate negative values so that they are set to 0 (meaning that blockTime will always be after that)
2. return a critical error

I argue for the first. since the registration period is a param, we assume it will be correctly set by the community. And if the value is too big, that means registration is always possible.

thoughts?

lubtd: Yes `1` sounds good to me
This is a edge case that can't happen in practice but it is reachable in theory by design based on the possible values for the param we should still think about them
We can eventually think back about setting a large max range for the registration period like one years

giunatale: Let me know if how I addressed it seems fair.
I am not sure what would be the best approach.

Perhaps adding back a max value could actually be the safest solution

x/participation/keeper/msg_participate.go
```diff
@@ -12,18 +13,38 @@ import (
...
+	}
+
+	// check if auction allows participation at this time
+	registrationPeriod := k.RegistrationPeriod(ctx)
+	auctionStartTime := auction.GetStartTime()
+	if auctionStartTime.Unix() < int64(registrationPeriod.Seconds()) {
+		// subtraction would result in negative value, clamp the result to ~0
+		// by making registrationPeriod ~= auctionStartTime
+		registrationPeriod = time.Duration(auctionStartTime.Unix()) * time.Second
+	}
```
lubtd, line 41: Do you think we can have a test case for this part?

x/participation/keeper/msg_participate.go
lubtd: The way it is addressed is fine IMO
This would happen in practice but since `block.Time()` and `StartTime` represent any timestamp we must handle edge cases allowed by design

x/participation/keeper/msg_participate_test.go
```diff
@@ -17,31 +17,37 @@ func Test_msgServer_Participate(t *testing.T) {
...
 		sdkCtx, tk, ts        = testkeeper.NewTestSetup(t)
 		auctioneer            = sample.Address()
-		sellingCoin           = sample.Coin()
-		startTime             = sdkCtx.BlockTime().Add(time.Hour * 10) // 10 hours from now
-		registrationPeriod    = time.Hour * 5                          // 5 hours before start
+		sellingCoin1          = sample.Coin()
+		sellingCoin2          = sample.Coin()
+		registrationPeriod    = time.Hour * 5 // 5 hours before start
+		startTime1            = sdkCtx.BlockTime().Add(time.Hour * 10)
+		startTime2            = time.Unix(int64((registrationPeriod - time.Hour).Seconds()), 0)
```
lubtd, line 24: If `startTime2` is only used for the test case where startTime < time(0) + Registration Period
, I would suggest avoiding numbered variable and make the name more descriptive

`startTimeLowerRegistrationPeriod`

x/participation/keeper/msg_participate_test.go
```diff
@@ -54,30 +60,41 @@ func Test_msgServer_Participate(t *testing.T) {
...
+				AuctionID:   auctionID1,
 				TierID:      2,
 			},
 			desiredUsedAlloc:      2,
-			currentAvailableAlloc: availableAllocTier2,
+			currentAvailableAlloc: availableAllocsTier[1],
 			blockTime:             validRegistrationTime,
 		},
+		{
+			name: "should allow participation when blockTime < registrationPeriod",
```
lubtd, line 83: Mmh not sure about the name of the test case, in this case we try to participate when [unix(0) - startTime] is shorter than registration period?

If this is the case, I would suggest `should allows participating when registration period is longer than range between Unix time 0 and auction's start time`

Which is a bit long but I thing it is fine for test maintainability

lubtd (approved):
👍

aljo242 (approved):

PR #644, opened 2022-03-23
chore: remove `.pre-commit-config.yaml`

aljo242:
I added `pre-commit` to give developers the option of running additional hooks right before they made commits (mainly for formatting).  It turns out that this tool does not let us ignore the `.pb.go` files which should not be modified by anything other than the `protoc` compiler.

I suggest we remove this tool because I'm not sure if anyone was using it other than me, and it seems it might cause more trouble than the utility it provides.


lubtd (approved):
Sounds good to me

giunatale (approved):

PR #645, opened 2022-03-23
refactor(`reward`): improve and refactor reward distribution tests

lubtd:
- Add more thorough test cases for the method
- Add check lastBlockHeight > currentRewardHeight since argument for `DistributeRewards` are set by the monitoring packet and therefore controlled by the testnet that could potentially send crafted values


x/reward/keeper/reward_distribution_test.go
```diff
@@ -82,6 +82,15 @@ func TestCalculateRewards(t *testing.T) {
 			},
 			want: sdk.NewCoins(),
 		},
+		{
+			name: "full block and signature ration should give all rewards",
```
aljo242, line 86:
```suggestion
			name: "full block and signature ratios should give all rewards",
```

aljo242 (changes requested):
lgtm !

aljo242 (approved):

giunatale (approved):

PR #646, opened 2022-03-23
feat(`participation`): simulate MsgParticipate

### LINKED ISSUE #638: Simulate `MsgParticpate`

aljo242:
Closes #638

### What does this PR does?

- Creates simulation for `MsgParticipate`.
- Adds simulation for `CreateFixedAuction` in `x/fundraising`. This can be removed once proper simulation is added to the fundraising module

Tested with `starport chain simulate` with seeds [1 - 10]

Tests pass, but I was able to see some simulation issues (in other modules) when running the larger simulation `make` targets.  I will open some issues for these.


x/participation/simulation/simulation.go
```diff
@@ -0,0 +1,249 @@
...
+		// auction must not be started
+		if !a.IsAuctionStarted(ctx.BlockTime()) {
+			return a, true
+		}
+	}
+
+	return auction, false
+}
+
+func RandomAccWithAllocations(ctx sdk.Context, r *rand.Rand,
```
giunatale, line 65:
```suggestion
func RandomAccWithAvailableAllocations(ctx sdk.Context, r *rand.Rand,
```

x/participation/module_simulation.go
```diff
@@ -18,6 +18,10 @@ const (
 	// TODO: Determine the simulation weight value
 	defaultWeightMsgParticipate int = 100
 
+	opWeightMsgCreateAuction = "op_weight_create_auction"
+	// TODO: Determine the simulation weight value
```
giunatale, line 22: Shouldn't we determine this in this PR?

x/participation/module_simulation.go
```diff
@@ -18,6 +18,10 @@ const (
 	// TODO: Determine the simulation weight value
```
giunatale, line 18: Shouldn't we determine this in this PR?

giunatale (changes requested):

x/participation/simulation/simulation.go
```diff
@@ -0,0 +1,248 @@
+package simulation
+
+import (
```
lubtd, line 3: goimport

x/participation/simulation/simulation.go
```diff
@@ -0,0 +1,248 @@
...
+		balances := bk.GetAllBalances(ctx, acc.Address)
+		if balances.IsAllGT(desired) {
+			return acc, balances, true
+		}
+	}
+
+	return simtypes.Account{}, sdk.NewCoins(), false
+}
+
+func RandomAuction(ctx sdk.Context, r *rand.Rand, fk fundraisingkeeper.Keeper) (auction fundraisingtypes.AuctionI, found bool) {
```
lubtd, line 45: I think we should put methods for store retrieval in a separate file like `store.go` in `launch`

lubtd, line 45: We will need to test these methods. If not implemented in this PR we must create an issue

x/participation/simulation/simulation.go
```diff
@@ -0,0 +1,248 @@
...
+				nil
+		}
+
+		startTime := ctx.BlockTime().Add(time.Hour * 24)
+		endTime := startTime.Add(time.Hour * 24 * 7)
+		msg := sample.MsgCreateFixedAuction(simAccount.Address.String(), sellCoin, startTime, endTime)
+
+		mintAmt := sdk.NewCoins(msg.SellingCoin)
+		// must mint and send new coins to auctioneer
+		err = bk.MintCoins(ctx, minttypes.ModuleName, mintAmt)
```
lubtd, line 194: We should not directly access keepers in simulation testing. We should use balance set initially for selling coins

lubtd (changes requested):

x/participation/simulation/simulation.go
aljo242: What do you mean by this?

How will the seller have an initial state with this coin if it is some new denom?

x/participation/simulation/simulation.go
aljo242: We can test functions like `FindAccount`, but I'm not sure how we can appropriately test the proper functionality of these `Random*` functions since the outcome should be random.

x/participation/simulation/simulation.go
lubtd: > What do you mean by this?

We should only interact with State Transition to actually simulate operations like in a real environment

Accounts have a balance with one denom in the initial state. Isn't it possible to add other denom in the initial state? What I mean is that the balances should be defined before any operation instead of minting them on the fly during the operations

x/participation/simulation/simulation.go
lubtd: We can have test that FindRandom<Something> returns a value if one something has been created like we have for `x/launch/simulation/store_test.go`

x/participation/simulation/simulation.go
aljo242: I wasn't sure if that was something we could do.  It might be configurable before the sim is run.  I'll look into it

aljo242: Ok, the solution will be to come up  with a custom version of the `AppStateFn` that gives some accounts balances of the staking token as well as some of these custom tokens. 

Then the simulation logic will need to be modified to look for accounts with existing amounts of custom tokens sufficient to start auctions.

Thoughts?

Here is the [function](https://github.com/cosmos/cosmos-sdk/blob/c91f3b26512cc6723ebbbbe986d2842e9c74e698/simapp/state.go#L28).

lubtd: Seems like the right way to go.
We could use a specific denom used as the selling coin for all auctions created

x/participation/module_simulation.go
```diff
@@ -14,9 +14,10 @@ import (
 )
 
 const (
-	opWeightMsgParticipate = "op_weight_msg_participate"
-	// TODO: Determine the simulation weight value
-	defaultWeightMsgParticipate int = 100
+	opWeightMsgParticipate            = "op_weight_msg_participate"
+	defaultWeightMsgParticipate   int = 100
```
giunatale, line 18:
```suggestion
	defaultWeightMsgParticipate   int = 50
```

Just a small suggestion coming from observations made on other module's weight (e.g. campaign, launch).
Also FYI I've set `defaultWeightMsgWithdrawAllocations` to 50

giunatale (changes requested):
From what I've seen in other modules, I am guessing we will need tests for the functions in `store.go` (see `simulation_test.go` in other modules)

Right?

app/sim_util/sim_util.go
```diff
@@ -0,0 +1,109 @@
+package sim_util
```
giunatale, line 1: var-naming warning, is it strictly necessary to have it named with the '_'?

giunatale (changes requested):
LGTM 👍 
just some small nitpicks

lubtd (approved):

Issue #648, opened 2022-03-24
simulate `msgWithdrawAllocations`


giunatale:
Add simulation for this message

waiting on #638

Issue #496, opened 2022-02-08
feat: implement and update invariants


Pantani:
Missing invariants implementation:

- [ ] Reward
- [ ] Monitoring Consumer
- [ ] Monitoring Provider

Outdate implementation:
- [ ] Profile
- [ ] Launch 
- [ ] Campaign

lubtd:
Invariant for `reward`: module account balance must be equal or greater than the sum of coins in reward pools

lubtd:
No change I see for `profile` invariants

giunatale:
@lubtd I've got almost ready `reward` and `monitoring-consumer`. Sorry I've dragged this for too long, and thanks for taking care of the other.
I think `monitoring-producer` does not need any invariant, can you double-check?

With your contributions, we should be fully covered. My PR should be ready today

PR #649, opened 2022-03-24
feat(`participation`): implement full `msgWithdrawAllocations`

### LINKED ISSUE #636: Implement full `MsgWithdrawAllocations` message

giunatale:
Closes #636 

It currently does not include simulation. I will deal with it in a separate PR (tracked in #648)


x/participation/keeper/msg_withdraw_allocations.go
```diff
@@ -12,18 +12,21 @@ import (
...
 		return nil, sdkerrors.Wrapf(types.ErrAuctionNotFound, "auction %d not found", msg.AuctionID)
 	}
 
-	blockTime := ctx.BlockTime()
 	if !auction.IsAuctionStarted(blockTime) {
 		return nil, sdkerrors.Wrapf(types.ErrAllocationsLocked, "auction %d not yet started", msg.AuctionID)
 	}
 
-	// TODO check delay is reached
+	withdrawalDelay := k.WithdrawalDelay(ctx)
```
lubtd, line 26: I'm thinking would `!auction.IsAuctionStarted(blockTime)` actually be necessary?
`!auction.IsAuctionStarted(blockTime)` implies `!blockTime.After(auction.GetStartTime().Add(withdrawalDelay))`

x/participation/keeper/msg_withdraw_allocations_test.go
```diff
@@ -76,14 +78,23 @@ func Test_msgServer_WithdrawAllocations(t *testing.T) {
...
 			err:       types.ErrAllocationsLocked,
 		},
+		{
+			name: "should prevent withdrawal before withdrawal delay has passed",
+			msg: &types.MsgWithdrawAllocations{
+				Participant: validParticipant,
+				AuctionID:   auctionID,
+			},
+			blockTime: auctionStartTime,
+			err:       types.ErrAllocationsLocked,
```
lubtd, line 96: I would suggest `ErrAllocationWithdrawalTimeNotReached` for the error

lubtd (commented):
Looks good, I will take the time to do some further experimentation later in the day👍

Is this PR blocking for anything?

x/participation/keeper/msg_withdraw_allocations.go
aljo242: Yeah I agree.  As long as we enforce that the param is positive, then the first part is always implied

aljo242 (changes requested):

x/participation/keeper/msg_withdraw_allocations.go
giunatale: > Looks good, I will take the time to do some further experimentation later in the day👍
> 
> Is this PR blocking for anything?

Would be nice if merged before I'm done with #648, but I need to work more on that before it might become blocking, so I would say no for now.

giunatale:
> Looks good, I will take the time to do some further experimentation later in the day👍
> 
> Is this PR blocking for anything?

Might be better if merged before #648 is fully addressed, but I'm still working on that. If it becomes blocking I could start working on #496

x/participation/keeper/msg_withdraw_allocations.go
```diff
@@ -12,18 +12,21 @@ import (
 
 func (k msgServer) WithdrawAllocations(goCtx context.Context, msg *types.MsgWithdrawAllocations) (*types.MsgWithdrawAllocationsResponse, error) {
```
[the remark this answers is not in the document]
giunatale: The response is kind of irrelevant for txs since they makes sense only in blocking mode. I think it's quite usual for responses to be empty, rather we should use events. But we haven't fully addressed that part yet I believe.

aljo242 (approved):

x/participation/keeper/msg_withdraw_allocations.go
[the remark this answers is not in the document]
lubtd: We can see response format later based on the usage by the front-end. Sometime adding values is practical, the response can be used to describe the state transition that occurs

For some values it necessary: the campaign ID of a new campaign for example since it's generated on-chain. Here the value that be determined before sending the message so it's less important I would say

lubtd (approved):

lubtd:
@giunatale any idea why Checks are not running for this PR?

giunatale:
Github being weird yesterday. I was wondering the same

giunatale:
I think we can trigger a new round of tests if we merge something else first, I'll make sure to update the branch here and merge after

lubtd:
> I think we can trigger a new round of tests if we merge something else first, I'll make sure to update the branch here and merge after

Ok, I look at this one https://github.com/tendermint/spn/pull/655

lubtd:
Yeah it did the trick. Weird.

PR #654, opened 2022-03-24
refactor(`campaign`): remove dynamic shares

### LINKED ISSUE #647: Removing `DynamicShare` option for campaign
This option allows increasing the number of shares that can be allocated for projects that want more flexibility but with the risk of having diluted shares.

This option is currently disabled and we don't have any plan to enable it back in the foreseeable future.

I propose to simply remove all the logic related to this feature in the codebase. This complicates the codebase for something we don't have any plan to support. Example: Storing max amount of share for a campaign different than the max share constant

aljo242:
Closes #647

### What does this PR does?

- Remove the `dynamicShares` field from `campaign`
- Remove the `UpdateShares` Tx which is a feature associated with DynamicShares
- Remove associated tests


lubtd:
wow thanks for having looked into this so quickly

x/launch/simulation/simulation.go
```diff
@@ -104,6 +105,8 @@ func SimulateMsgEditChain(ak types.AccountKeeper, bk types.BankKeeper, k keeper.
 			modify,
 		)
 
+		fmt.Println(msg)
```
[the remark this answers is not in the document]
giunatale: May be a leftover from your internal testings, I assume

giunatale (changes requested):

x/launch/simulation/simulation.go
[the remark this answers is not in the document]
aljo242: ah thank you !

giunatale (dismissed):
LGTM. Haven't checked though if there's any reference left to dynamic shares in the codebase

Pantani (dismissed):

proto/campaign/campaign.proto
```diff
@@ -22,11 +22,10 @@ message Campaign {
     (gogoproto.casttype) = "github.com/cosmos/cosmos-sdk/types.Coin",
     (gogoproto.castrepeated) = "Shares"
   ];
-  bool dynamicShares = 8;
-  repeated cosmos.base.v1beta1.Coin totalShares = 9 [
+  repeated cosmos.base.v1beta1.Coin totalShares = 8 [
```
lubtd, line 25: We remove `totalShares` and related logic as well

`totalShares` was used to specify the new total when `MsgUpdateTotalShares` was used

lubtd (changes requested):
We can refactor `IsTotalSharesReached` in `types`

The method takes a `totalShares` share argument that specify the share with a custom total but now this argument become irrelevant since all shares will be compared to the `MaxShare` universal value

proto/campaign/campaign.proto
aljo242: Ok - I can pick this work up once #656 gets merged so I can use some of those additions.

Pantani (dismissed):

x/campaign/types/campaign.go
```diff
@@ -37,7 +39,7 @@ func (m Campaign) Validate() error {
 		return errors.New("invalid allocated shares")
 	}
 
-	if IsTotalSharesReached(m.AllocatedShares) {
+	if IsTotalSharesReached(m.AllocatedShares, spntypes.TotalShareNumber) {
```
lubtd, line 42: it's the same problematic here as in the previous comment
We should rely on the value set in the store

Since `Validate` is used for genesis validation, I would suggest adding `TotalShares` as part of the genesis state (not param though), the value can then be accessed in genesis `Validate` and this campaign `Validate` can take `TotalShare` as a parameter

testutil/sample/campaign.go
```diff
@@ -1,6 +1,7 @@
 package sample
 
 import (
+	spntypes "github.com/tendermint/spn/pkg/types"
```
lubtd, line 4: goimport

x/campaign/types/genesis.go
```diff
@@ -2,6 +2,7 @@ package types
 
 import (
 	"fmt"
+	spntypes "github.com/tendermint/spn/pkg/types"
```
lubtd, line 5: goimport

x/campaign/types/genesis_test.go
```diff
@@ -1,6 +1,9 @@
 package types_test
 
 import (
+	"fmt"
+	spntypes "github.com/tendermint/spn/pkg/types"
```
lubtd, line 5: goimport

lubtd (commented):
Everything looks good! Just need to fix the goimport

giunatale (dismissed):

lubtd (dismissed):

giunatale (dismissed):

lubtd (approved):

PR #655, opened 2022-03-24
fix(`simulation`): add fee checks

### LINKED ISSUE #652: Simulated Tx with `creationFee` can cause failure due to insufficient funds
Running `make test-sim-profile` on the `develop` gives the following error:

```shell
Simulating... block 18/500, operation 0/405.--- FAIL: BenchmarkSimulation
    simulate.go:300: error on block  18/500, operation (60/296) from x/launch:
        failed to execute message; message index: 0: 3stake is smaller than 7stake: insufficient funds [cosmos/cosmos-sdk@v0.44.6/x/bank/keeper/send.go:187]
        Comment: unable to deliver tx
```

This error can come from the `launch` or `campaign` modules.  Random accounts with enough balance to cover chain or campaign creation fees are chosen, but this can lead to a situation where the account does not have enough funds to cover this fee AND the gas required for the Tx.  This can be fixed by adding a check that the account has the balance to cover the fee + the gas during account selection.

Version: `latest-25bfc55f`

aljo242:
Closes #652

Add a balance check to `SimulateCreateCampaign` and `SimulateCreateChain` that also includes checking the account balance is enough to cover the creation fee as well as a pre-calculated random gas fee.

`make test-sim-profile` does not run into the issue described in the attached issue, but there are still simulation issues to be dug into.

### DIFF
--- a/x/launch/simulation/simulation.go
+++ b/x/launch/simulation/simulation.go
@@ -34,11 +34,23 @@ func SimulateMsgCreateChain(ak types.AccountKeeper, bk types.BankKeeper, k keepe
 			// No message if no coordinator
 			return simtypes.NoOpMsg(types.ModuleName, types.TypeMsgCreateChain, "skip create a new chain"), nil, nil
 		}
+
 		// skip if account cannot cover creation fee
 		creationFee := k.ChainCreationFee(ctx)
-		if !creationFee.Empty() && !bk.SpendableCoins(ctx, simAccount.Address).IsAllGTE(creationFee) {
-			return simtypes.NoOpMsg(types.ModuleName, types.TypeMsgCreateChain, "skip campaign creation"), nil, nil
+		customFee, err := simtypes.RandomFees(r, ctx, creationFee)
+		if err != nil {
+			return simtypes.NoOpMsg(
+					types.ModuleName,
+					types.TypeMsgCreateChain,
+					"skip chain creation"),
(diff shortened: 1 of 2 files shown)


giunatale (approved):

x/launch/simulation/simulation.go
```diff
@@ -34,11 +34,23 @@ func SimulateMsgCreateChain(ak types.AccountKeeper, bk types.BankKeeper, k keepe
...
-		if !creationFee.Empty() && !bk.SpendableCoins(ctx, simAccount.Address).IsAllGTE(creationFee) {
-			return simtypes.NoOpMsg(types.ModuleName, types.TypeMsgCreateChain, "skip campaign creation"), nil, nil
+		customFee, err := simtypes.RandomFees(r, ctx, creationFee)
+		if err != nil {
+			return simtypes.NoOpMsg(
+					types.ModuleName,
+					types.TypeMsgCreateChain,
+					"skip chain creation"),
+				nil,
+				nil
```
lubtd, line 47: I'm wondering, shouldn't it be an actual error at this point? is RandomFees supposed to fail under some circumstances based on the state of the chain?

aljo242: that's a good point.  This is an error that has nothing to do with the simulation / state of the chain, so it should be passed out of the function.

PR #656, opened 2022-03-25
feat(`campaign`): max shares query

### LINKED ISSUE #653: Create a query to get the maximum number of shares
The maximum of share is set to `DefaultTotalShareNumber = 100000` in `campaign/types/shares.go`

Since it is hardcoded we should provide a query under `campaign` namespace to fetch these values

I think we can also rename `DefaultTotalShareNumber` into `TotalShareNumber` and possibly move it in `pkg/types/constant.go` since it's quite an important value in the chain as a whole

aljo242:
Closes #653

### What does this PR does?

- Adds the `MaximumShares` query to the `campaign` keeper.
- Adds appropriate tests
- Adds a keeper entry for the MaximumShares value to prevent state disagreement between nodes (as discussed in #653).  The value is set from the constant defined in `constant.go` in during the app's `InitChainer`


x/campaign/client/cli/query_maximum_shares_test.go
```diff
@@ -0,0 +1,48 @@
+package cli_test
```
lubtd, line 1: I suggest removing this test as it doesn't bring value above `grpc_maximum_shares_test.go` and increase the likelihood to have the `address already in use` error for action

lubtd (approved):

lubtd:
Thinking about it, it may be better to name it simply `TotalShares`

`Maximum` reflect a max value that can eventually be reached but not obviously. While `Total` is the total when 100% of share are allocated

giunatale (approved):

lubtd (approved):

PR #661, opened 2022-03-28
refactor(`sample`): take random generators as objects

### LINKED ISSUE #660: Inconsistent random generation in simulation tests
Simulation tests can only have consistent state between runs if the same random generator is used for all operations (with the same seeding).  It appears to me that we are not creating the genesis state for our custom modules in a consistent way. 

To illustrate this, let's look at the difference between how the Randomized Genesis of the `x/bank` module is generate versus our `x/campaign` module:
- [bank](https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/simulation/genesis.go)
- [campaign](https://github.com/tendermint/spn/blob/87ee3fc6afc63fb7c79db0e9322e27758cc45fad/x/campaign/module_simulation.go#L45)

We should not use the `sample` package to create these initial states and should rather module the initial generation after how the other cosmos-sdk modules are created.

aljo242:
Closes #660

### What does this PR does?

The usage of the `sample` package is extensive throughout our tests as well as the simulation code.  Because of this, refactoring this package to take a random number generator (`r *rand.Rand`) as an additional seems like a better approach to tackling consistent random generation for tests.  

This PR refactors all `sample` code that use the `math/rand` package as well as all associated uses of them in our tests.  Now, all simulation objects are generated using a single `rand.Rand` object.

Unfortunately, this does not look like it is addressing the nondeterminism issue (try running `make test-sim-nondeterminism`).  Still, I think this should be merged because it makes testing and simulation more regular and takes one cause of nondeterminism off the table.


lubtd:
> Unfortunately, this does not look like it is addressing the nondeterminism issue (try running make test-sim-nondeterminism). Still, I think this should be merged because it makes testing and simulation more regular and takes one cause of nondeterminism off the table.

I will take some time on this tomorrow and we can discuss it in the call

lubtd:
I think the overall change is good. Although `sample` purpose is to have a sample way to instantiate objects for tests purpose the additional argument doesn't really complicate the interface

For the `AppHash` issue, I saw that we are also using `time.Now()` in `sample` for `launch` which is used in Launch genesis state. We must remove them as well since source of non determinism

Have you tried setting all genesis state for simulation as default genesis just to verify the issue would eventually be solved?

lubtd (approved):

giunatale (approved):

PR #664, opened 2022-03-29
fix(`simulation`): remove non-deterministic sim behavior

### LINKED ISSUE #659: Nondeterminism detected in simulation tests
Running `make test-sim-nondeterminsim` on `develop` yields the following result:

```shell
 Error Trace:    simulation_test.go:197
                Error:          Not equal: 
                                expected: " \x87D\xa7ͯ(<\x1e@ݫ/+\xee6\x9a\xd2%\x8f\xcfX\xec8\x9fv\xf7ł\xf2\xd7\x1f"
                                actual  : "0\xf27\xaa\xd0\x10\xaf\xf8SFA\x82\x9d#x\xeeK\xe4\x1e\x00F\xb3\x1d\x91<\x92\x05\xbb!\xca(\xa5"
                                
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                - �D�ͯ(<@ݫ/+�6��%��X�8�v�ł��
                                +0�7����SFA��#x�K�F��<��!�(�
                Test:           TestAppStateDeterminism
                Messages:       non-determinism in seed 5577006791947779410: 1/3, attempt: 2/5
```

This non-determinism could be due to many issues, but the best case scenario for us is that the issue has to do with how our simulation tests are set up rather than actually having issues in the chain.  I will open issues for possible causes of this behavior and encourage us all to try to investigate.

**Please provide output of commands**
 - `spn version`: `latest-87ee3fc6`
 - `go version`:  `go1.18 linux/amd64`
 - `uname -a`:  `Linux cozart-dev 5.16.16-arch1-1 #1 SMP PREEMPT Mon, 21 Mar 2022 22:59:40 +0000 x86_64 GNU/Linux`

aljo242:
Closes #659

### What does this PR does?

Removes the usage of a `map` in the setup code for simulating `MsgRequestRemoveAccount`.  The traversal of maps in range-based for loops is non-deterministic, so this was causing the issue.

To verify, run `make test-sim-nondeterminism` on your local machine.

### DIFF
--- a/x/launch/simulation/simulation.go
+++ b/x/launch/simulation/simulation.go
@@ -212,14 +212,26 @@ func SimulateMsgRequestRemoveAccount(ak types.AccountKeeper, bk types.BankKeeper
 	return func(
 		r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accs []simtypes.Account, chainID string,
 	) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
-		accsChain := make(map[string]uint64)
+		type accChain struct {
+			address  string
+			launchID uint64
+		}
+
+		accChainList := make([]accChain, 0)
+
 		genAccs := k.GetAllGenesisAccount(ctx)
 		for _, acc := range genAccs {
-			accsChain[acc.Address] = acc.LaunchID
+			accChainList = append(accChainList, accChain{
+				address:  acc.Address,
+				launchID: acc.LaunchID,
+			})
 		}
 		vestAccs := k.GetAllVestingAccount(ctx)
 		for _, acc := range vestAccs {
-			accsChain[acc.Address] = acc.LaunchID
+			accChainList = append(accChainList, accChain{
+				address:  acc.Address,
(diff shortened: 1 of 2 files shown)


x/launch/simulation/simulation.go
```diff
@@ -228,36 +240,37 @@ func SimulateMsgRequestRemoveAccount(ak types.AccountKeeper, bk types.BankKeeper
 			accChainID uint64
 		)
 		found := false
-		for acc, chainID := range accsChain {
-			if IsLaunchTriggeredChain(ctx, k, chainID) {
+		for _, accChain := range accChainList {
```
lubtd, line 243: It will find the first account in the list that fulfills all conditions

I suggest we add a shuffling in the list `r.Suffle` to add more entropy otherwise there is a high change only genesis account and not vesting account will be fetched

lubtd (commented):
Great the simulation test runs well!

Just one comment

lubtd (approved):

PR #674, opened 2022-03-30
feat(`participation`): check if auction is cancelled in messages

### LINKED ISSUE #665: Allow direct withdrawal of allocations if the auction is canceled
An auction can be canceled by the auctioneer https://github.com/tendermint/fundraising/tree/main/docs/How-To/cli#cancelauction

We should allow the withdrawal without delay condition if the state of the auction is `CANCELED`

### LINKED ISSUE #673: Prevent participation if auction is cancelled
Similarly to #665, we should prevent participation if the auction is cancelled.
An auction can be cancelled *only if the auction has not yet started*

giunatale:
Closes #673 
Closes #665


aljo242 (approved):
lgtm!

PR #676, opened 2022-03-31
chore: update fundraising version and add script for fundraising testing

lubtd:
- Update `fundraising` to the current development commit
- Add script for fundraising experimentation
- Set `WithdrawalDelay` to 5 seconds in Starport config for test purpose


scripts/test_fundraising/sale.py
```diff
@@ -0,0 +1,62 @@
...
+    jf['selling_coin']['denom'] = sellingDenom
+    jf['selling_coin']['amount'] = sellingAmount
+    jf['paying_coin_denom'] = payingDenom
+    jf['start_price'] = price
+    jf['start_time'] = startTime
+    jf['end_time'] = endTime
+    with open(saleFile, 'w', encoding='utf-8') as newF:
+        json.dump(jf, newF, ensure_ascii=False, indent=4)
+
+# Initialization
```
giunatale, line 31: Although probably not important here, I'd always advise including running code directly in the script and instead rely on the [main boilerplate](https://www.python-boilerplate.com/py3+executable)

scripts/test_fundraising/sale.py
```diff
@@ -0,0 +1,62 @@
+import json
+import subprocess
+import datetime
+import time
+
+valAddress = 'spnvaloper1ezptsm3npn54qx9vvpah4nymre59ykr9mx22g4'
```
giunatale, line 6: Also FYI, variable names  and functions in python should not use camelcase, but include '_' to separate words ([PEP8: function and variable names](https://peps.python.org/pep-0008/#function-and-variable-names)). We are not using python as our main language but we may want to follow standards as much as possible since we are writing OSS.

scripts/test_fundraising/sale.py
lubtd: Yes, it's just because it's simpler to modify without another indentation

lubtd:
I will wait for new release of `fundraising` before finalizing the task

lubtd:
Closing in favor of https://github.com/tendermint/spn/pull/816

PR #677, opened 2022-03-31
fix(`localnet`): remove additional comma in `genesis_template.json`

lubtd:
Everything in the title


aljo242 (approved):

lubtd:
EDIT: add `totalShares` value as well in the genesis. Otherwise the zero value `0` was used and no voucher was mintable

giunatale (approved):
thanks for the fix

PR #679, opened 2022-03-31
chore: upgrade ibc-go to v3.0.0

### LINKED ISSUE #670: Upgrading IBC from v2.0 to v3.0
https://github.com/cosmos/ibc-go/releases/tag/v3.0.0

aljo242:
Closes #670

### What does this PR does?

Follows the following [migration guide](https://github.com/cosmos/ibc-go/blob/main/docs/migrations/v2-to-v3.md).  Please test on your local setups or with any working IBC connections.


lubtd:
I have some issues with incentivized testnets, I may think I need to upgrade https://github.com/lubtd/orbit to v3.0 as well

If IBC version must be the same between two chains we will need the same version in provider testnet, which means different from the version we use with Starport scaffolding currently.

We can discuss it in the next Blockchain sync

pkg/types/consensus_state.go
```diff
@@ -4,12 +4,12 @@ package types
 import (
 	"encoding/base64"
 	"encoding/hex"
+	"gopkg.in/yaml.v2"
```
lubtd, line 7: I got check failure from `golangci-lint`

Are you sure the previous import was not right?

lubtd (commented):
We can remove `NegotiateAppVersion` from both monitoring modules since no longer part of the interface

lubtd:
The error happens on `ChanOpenInit` while the guide mention no modification on this part.
It could also be an issue with the relayer, I'm using Hermes `0.13`
```
ERROR ThreadId(14) send_tx{id=spn-1}:estimate_gas: failed to simulate tx. propagating error to caller: gRPC call failed with status: status: Unknown, message: "failed to execute message; message index: 0: channel open init callback failed: channel not found for channel ID: channel-0: channel not found: invalid handshake", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "53"} }
```

lubtd:
Let's keep this on hold for now and focus on finalizing Milestone A

lubtd:
Bug `channel-0: channel not found` still occuring. Even with `hermes` `0.15`
We may need to try CLI `reward release` with https://github.com/ignite-hq/cli/pull/2516

lubtd:
Found the issue, creating a PR for it

lubtd:
https://github.com/tendermint/spn/pull/828

lubtd (approved):

giunatale (approved):

PR #684, opened 2022-04-04
fix(`campaign`): fix sample genesis with inconsistent state

lubtd:
`CampaignGenesisState` from `sample` contains campaigns initialized with empty allocated shares and mainnet accounts containing shares.

This genesis state was inconsistent because the sum of shares allocated for accounts must be tracked by the `AllocatedShares` value of the campaign

This PR removes the accounts for a quick fix. A further change can include adding accounts with the allocated correctly tracked for better entropy


aljo242 (approved):
thanks !

giunatale (approved):

PR #685, opened 2022-04-04
feat(`launch`): add request status field

### LINKED ISSUE #658: Add `status` field in request
- Add `status` field that can be equal to `PENDING|APPROVED|REJECTED`
- `MsgSettleRequest` fails if `request.Status != PENDING`
- No longer remove requests from the store on settled

aljo242:
Closes #658

### What does this PR does?

- Adds a `Status` field to the `Request` proto. 
- requests are no longer removed from the store - only their status is changed
- auto-approved requests are also set in the store with the "APPROVED" status
- refactored tests to accommodate for these changes
- refactor simulation `store` functions to only use requests with a "PENDING" status


lubtd (approved):
LGTM

giunatale (approved):

PR #689, opened 2022-04-05
refactor(`launch`): improve invariant

lubtd:
Replace the launch timestamp invariant with an invariant that checks all chains are valid

### DIFF
--- a/x/launch/keeper/invariants.go
+++ b/x/launch/keeper/invariants.go
@@ -9,15 +9,15 @@ import (
 )
 
 const (
-	zeroLaunchTimestampRoute = "zero-launch-timestamp"
-	duplicatedAccountRoute   = "duplicated-account"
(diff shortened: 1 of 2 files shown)


giunatale (approved):
LGTM. 

We should also have an invariant that checks if the associated campaign really exists in the campaign keeper (if set), but I can handle it in a separate PR

giunatale:
> LGTM.
> 
> We should also have an invariant that checks if the associated campaign really exists in the campaign keeper (if set), but I can handle it in a separate PR

I'm actually thinking this check would be better suited in this invariant. Do you want to do it @lubtd? Otherwise I can either push onto this branch or address it after this PR is merged.

giunatale:
@aljo242 @Pantani if any of you can review this so we can merge it, I can then unblock and finalize also #694 🙏

aljo242 (approved):

PR #694, opened 2022-04-05
feat: implement remaining invariants

### LINKED ISSUE #496: feat: implement and update invariants

giunatale:
Closes #496 

Implements invariants for `reward`, `monitoringc`, and `launch`

[Waiting on #689 so I can add the campaign check in `launch`]


aljo242 (dismissed):
lgtm!

x/monitoringc/keeper/invariants.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+		}
+		for _, launchIDFromVerifiedClientID := range allLaunchIDFromVerifiedClientlID {
+			if _, ok := clientIDMap[clientIDKey(launchIDFromVerifiedClientID.LaunchID, launchIDFromVerifiedClientID.ClientID)]; !ok {
+				return sdk.FormatInvariant(
+					types.ModuleName, missingVerifiedClientIDRoute,
+					"client id from launchIDFromVerifiedClientID list not found",
+				), true
+			}
+		}
```
lubtd, line 56: The `launchIDFromVerifiedClientID` entry is created  when the associated when the client ID is added to `VerifiedClientID`. So that when the IBC connection is performed the corresponding launch ID can be retrieved.

Under some condition (`MsgRevertLaunch`) the `VerifiedClientID` is erased to prevent make an IBC connection with the chain but the `launchIDFromVerifiedClientID` is not removed, which would break this invariant.

I think the condition should be the opposite: Every verified client ID should have an associated `launchIDFromVerifiedClientID` to allow the correct configuration once the connection is performed.

A `launchIDFromVerifiedClientID` without an actual verified client ID would be just dead data in the store

lubtd (changes requested):

x/monitoringc/keeper/invariants.go
giunatale: Sounds good. I've removed completely the check for `ProviderClientID` and implemented the invariant how you suggested

testutil/keeper/bank.go
```diff
@@ -13,3 +13,9 @@ func (tk TestKeepers) Mint(ctx sdk.Context, address string, coins sdk.Coins) {
 	require.NoError(tk.T, tk.BankKeeper.MintCoins(ctx, minttypes.ModuleName, coins))
 	require.NoError(tk.T, tk.BankKeeper.SendCoinsFromModuleToAccount(ctx, minttypes.ModuleName, sdkAddr, coins))
 }
+
+// MintModule mints the specified coins into the module account balance
+func (tk TestKeepers) MintModule(ctx sdk.Context, moduleAcc string, coins sdk.Coins) {
+	require.NoError(tk.T, tk.BankKeeper.MintCoins(ctx, minttypes.ModuleName, coins))
+	require.NoError(tk.T, tk.BankKeeper.SendCoinsFromModuleToModule(ctx, minttypes.ModuleName, moduleAcc, coins))
```
lubtd, line 20:
```suggestion
	require.NoError(tk.T, tk.BankKeeper.MintCoins(ctx, moduleAcc, coins))
```

x/launch/keeper/invariants_test.go
```diff
@@ -33,18 +33,29 @@ func TestInvalidChainInvariant(t *testing.T) {
...
 		chain.LaunchID = tk.LaunchKeeper.AppendChain(ctx, chain)
 		msg, broken := keeper.InvalidChainInvariant(*tk.LaunchKeeper)(ctx)
 		require.True(t, broken, msg)
 	})
+
+	t.Run("invalid case - chain does not have a valid associated campaign", func(t *testing.T) {
+		chain := sample.Chain(r, 0, 0)
+		chain.HasCampaign = true
+		chain.CampaignID = 1000
+		msg, broken := keeper.InvalidChainInvariant(*tk.LaunchKeeper)(ctx)
```
lubtd, line 56: The chain is not set therefore the condition is not checked

The invariant is still broken because of the state above is reused

We should instatiate a new `tk` or defined another test function

x/reward/keeper/invariants_test.go
```diff
@@ -0,0 +1,62 @@
...
+			pool := sample.RewardPool(r, i)
+			tk.MintModule(ctx, types.ModuleName, pool.RemainingCoins)
+			tk.RewardKeeper.SetRewardPool(ctx, pool)
+		}
+		// add some invalid pools - do not mint coins
+		for i := uint64(10); i < uint64(20); i++ {
+			pool := sample.RewardPool(r, i)
+			tk.RewardKeeper.SetRewardPool(ctx, pool)
+		}
+		msg, broken := keeper.InsufficientRewardsBalanceInvariant(*tk.RewardKeeper)(ctx)
```
lubtd, line 59: Same comment as above. The store with broken invariant is reused, we should instantiate another tk

x/reward/keeper/invariants_test.go
```diff
@@ -0,0 +1,62 @@
...
+	t.Run("invalid case 1", func(t *testing.T) {
+		// add some valid pools
+		for i := uint64(0); i < uint64(10); i++ {
+			pool := sample.RewardPool(r, i)
+			tk.MintModule(ctx, types.ModuleName, pool.RemainingCoins)
+			tk.RewardKeeper.SetRewardPool(ctx, pool)
+		}
+		// add some invalid pools - mint a bit less for some coins
+		for i := uint64(10); i < uint64(20); i++ {
+			pool := sample.RewardPool(r, i)
```
lubtd, line 36: With `sample.RewardPool` we have random denoms for every reward pools.

To improve the test of the invariant I think we should reuse the same denom between rewards pool so that we check the balance must be greater than sum of coins in reward pools

testutil/keeper/bank.go
giunatale: What if the module does not have mint permissions?

giunatale: I think it is cleaner (and fail-safe) this way, tbh

lubtd: Yes you're right

x/launch/keeper/invariants_test.go
giunatale: Yeah, fair! my bad

giunatale: Note though: we reuse same `tk` basically in every other invariant test. I think for some it makes sense, but it could not for some others.

lubtd: I think most of the cases we should have valid and invalid cases where it doesn't pose a problem but we can have a check

giunatale: I checked, looks like we are good. Perhaps if you could give a look yourself at `campaign`, it's the only one I haven't looked at.

lubtd (approved):

aljo242 (approved):

PR #695, opened 2022-04-06
test: add invariant testing to simulation tests

### LINKED ISSUE #666: Add invariant checks to simulations
Our longer running simulations (especially non-determinism) tests exist to fully stress the blockchain, but they currently do not have Invariant checking enabled.

This is an easy addition, as we just have to set a boolean flag in the setup code.  I propose we make this addition to make these tests a bit stronger.

aljo242:
Closes #666

### What does this PR does?

- update the `fundraising` module to the latest version and make required changes
- Add the `fundraising` module to the simulation manager
- Move the order of the `fundraising` EndBlocker to before the `crisis` EndBlocker (where invariants are checked)
- update the Makefile targets and simulation_test code to improve sim testing
- Setup simulation accounts with more balance so sims run longer

### How to test?

- Run the simulation test targets (some may still fail due to  other errors)


app/simutil/app_state.go
```diff
@@ -2,6 +2,8 @@ package simutil
 
 import (
 	"encoding/json"
+	"fmt"
+	simappparams "github.com/cosmos/cosmos-sdk/simapp/params"
```
giunatale, line 6: goimport

x/participation/keeper/msg_participate.go
```diff
@@ -32,7 +32,7 @@ func (k msgServer) Participate(goCtx context.Context, msg *types.MsgParticipate)
 	}
 
 	// check if auction is already started
-	if auction.IsAuctionStarted(blockTime) {
+	if auction.ShouldAuctionStarted(blockTime) {
```
giunatale, line 35: I believe from the API name change that we are required now to check also the status, am I right?

giunatale: This could require a bit more work for the two messages (and tests) in participation.

giunatale (changes requested):

x/participation/keeper/msg_participate.go
lubtd: I think the logic remains the same

lubtd (commented):
No more non-determinism issue👍

At least until block 78, we're getting now with `make test-sim-nondeterminism` this issue which may be related to https://github.com/tendermint/spn/issues/369
```
    simulate.go:300: error on block  78/500, operation (81/87) from x/staking:
        out of gas in location: WritePerByte; gasWanted: 1000000, gasUsed: 1001847: out of gas [cosmos/cosmos-sdk@v0.45.1/baseapp/recovery.go:55]
        Comment: unable to deliver tx
```

x/participation/keeper/msg_participate.go
aljo242: The logic is the same, the function name was just changed

giunatale: I disagree. Although the logic remains the same, the API name-change is clear. Having `true` returned from this function indicates only that the auction *should* be started, not that it **has** started. Anywhere where we need to be really sure that it is the case, we should check the status (and not just this API).

I personally would consider this relevant, but I guess if both of you think differently, I'm fine with that.

aljo242: The logic inside the function is the same though.  They just changed the name.  This is because to actually check the status of the auction, you need to get the `Status` field.  

So perhaps we should open an issue to check if the `Status` field is correct, instead of this one.  These functions (`ShouldAuctionStarted`, etc) are used in the `fundraising` EndBlocker to modify the actual `Status` field of auctions.

giunatale: Yeah, I think we should

giunatale (dismissed):

lubtd (approved):

giunatale (approved):

lubtd (approved):

PR #699, opened 2022-04-07
fix(`participation`): proper auction status checks in `msgParticipate`

### LINKED ISSUE #696: Check `Status` field of auction in `MsgParticpate`
Currently we use the `ShouldAuctionStarted()` function to check if the auction is started in `MsgParticipate`.

We should instead use the `GetStatus()` method and evaluate based on the Auction Status field.

giunatale:
Closes #696

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/spf13/cast v1.4.1
 	github.com/spf13/cobra v1.3.0
 	github.com/stretchr/testify v1.7.0
-	github.com/tendermint/fundraising v0.1.2-0.20220406040803-d65fe31abb77
+	github.com/tendermint/fundraising v0.2.0
 	github.com/tendermint/starport v0.19.5
 	github.com/tendermint/tendermint v0.34.16
 	github.com/tendermint/tm-db v0.6.6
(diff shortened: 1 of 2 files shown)


x/participation/keeper/msg_participate_test.go
```diff
@@ -37,14 +37,22 @@ func Test_msgServer_Participate(t *testing.T) {
 	// initialize auction
 	tk.Mint(sdkCtx, auctioneer, sdk.NewCoins(sellingCoin1))
 	auctionID1 := tk.CreateFixedPriceAuction(sdkCtx, r, auctioneer, sellingCoin1, startTime, endTime)
-	// initialize auction with edge case start time
+	// initialize auction with edge case start time (forcefully set to status standby)
```
lubtd, line 40: Just a feedback that I should have already given before. I would suggest using more explicit names depending on the test case for the variables. Using numbers makes the tests hard to read
Example
`auctionID1` -> `auctionRegistrationPeriod`
`auctionID3` -> `auctionStarted`

aljo242 (approved):

lubtd (approved):

PR #700, opened 2022-04-07
test(`reward`): improve tests for `MsgSetRewards`

lubtd:
Part of https://github.com/tendermint/spn/issues/687

- Add tests for `grpc_params` for null queries
- Add test cases for `MsgSetRewards` when setting rewards for a new reward pool
- Add some helpers for test above

Issue #687, opened 2022-04-05
Code coverage - Check up


lubtd:
Report of the files where code coverage can potentially be improved
Checked from output in `cover.html` from `make test-cover` command

- Most of the files would only require minimal addition
- Missing tests I consider as high priority are pointed out
- `monitoring_packet` file tests from both monitoring modules would potentially be more complex, implying mocking from IBC

- [x] `campaign`
```
	genesis
	keeper/
		invariant
		total_shares
		grpc_params
	types/
		shares
```
- [x] `launch`
```
	keeper/
		grpc_params
		keeper
		msg_settle_request
		request
		validator_set
	types/
		msg_trigger_launch
		params
		request_content
```
- [x] `monitoringc`
```
	keeper/
		grpc_params
		grpc_verified_client_ids
		handshake
		monitoring_packet - not tested, would eventually be a bigger task for IBC mocking
		msg_create_client
	types/
		genesis - high priority
		params
```
- [x] `monitoringp`
```
	keeper/
		begin_block
		grpc_connection_channel_id
		grpc_consumer_client_id
		grpc_monitoring_info
		grpc_params
		handshake
		monitoring_packet - not tested, would eventually be a bigger task for IBC mocking
```
- [x] `participation`
```
	genesis
	keeper/
		grpc_params
		msg_participate
	types/
		genesis - high priority
		params
```
- [x] `profile`
```
	genesis
	types/
		genesis
```
- [x] `reward`
```
	keeper/
		grpc_params
		msg_set_reward - high priority, code verifying sufficient balance for reward pool is not covered
		reward_distribution
	types/
		genesis - high priority
		reward_pool
```

lubtd:
I will work on `msg_set_reward` and the different `grpc_params` (where we just need a test to handle nil request) @giunatale @aljo242

giunatale:
I can take the remainder of `participation` and `profile`.

giunatale:
I can also take on `campaign`

aljo242:
I will take on `launch` and `reward` to start

giunatale:
@aljo242 I'm taking `monitoringc`, can you take on `reward` and `monitoringp` so we can finish this task?

giunatale:
Eventually we'll do another round of check up, but I suggest we get this done as soon as possible.

PR #706, opened 2022-04-08
test: increase code coverage for `campaign`, `participation` and `profile`

giunatale:
Part of #687

Increase code coverage of the modules above with the excetpion of `keeper/grpc_params` (for `campaign` and `participation`) that are handled by @ludtb in #700

### DIFF
--- a/x/campaign/keeper/invariants.go
+++ b/x/campaign/keeper/invariants.go
@@ -113,14 +113,6 @@ func CampaignSharesInvariant(k Keeper) sdk.Invariant {
 			campaignID := campaign.CampaignID
 			expectedAllocatedSharesShares := accountSharesByCampaign[campaignID]
 
-			campaign, found := k.GetCampaign(ctx, campaignID)
-			if !found {
-				return sdk.FormatInvariant(
-					types.ModuleName, campaignSharesRoute,
-					fmt.Sprintf("%s: %d", types.ErrCampaignNotFound, campaignID),
-				), true
-			}
-
 			// read existing denoms from allocated shares of the campaign to check possible minted vouchers
 			allocated, err := types.SharesToVouchers(campaign.GetAllocatedShares(), campaignID)
 			if err != nil {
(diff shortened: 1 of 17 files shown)


x/campaign/keeper/invariants.go
```diff
@@ -113,14 +113,6 @@ func CampaignSharesInvariant(k Keeper) sdk.Invariant {
 			campaignID := campaign.CampaignID
 			expectedAllocatedSharesShares := accountSharesByCampaign[campaignID]
 
-			campaign, found := k.GetCampaign(ctx, campaignID)
```
giunatale, line 116: *note*: this was redundant, we were already getting all campaigns from the store and iterating over all campaigns in the loop. No need to access the KVStore again (and check existence)

x/campaign/keeper/invariants.go
```diff
@@ -140,14 +132,9 @@ func CampaignSharesInvariant(k Keeper) sdk.Invariant {
...
-			vShares, err := types.VouchersToShares(vouchers, campaignID)
-			if err != nil {
-				return sdk.FormatInvariant(
-					types.ModuleName, campaignSharesRoute,
-					"fail to convert vouchers to shares",
-				), true
-			}
+			// convert to shares and add to the campaign shares - since we are converting shares to vouchers earlier,
+			// this conversion back to shares will never fail by design, thus we can ignore the error
+			vShares, _ := types.VouchersToShares(vouchers, campaignID)
```
giunatale, line 137: As commented, we can just avoid checking the error. This conversion cannot fail (see L117 and the use of `types.SharesToVouchers`). I can re-add it but it would be dead code

aljo242 (approved):

x/campaign/keeper/invariants.go
aljo242: You can add a `//nolint` message here, perhaps

lubtd (changes requested):
LGTM

Just one comment

aljo242 (dismissed):

x/campaign/keeper/invariants.go
```diff
@@ -138,7 +138,7 @@ func CampaignSharesInvariant(k Keeper) sdk.Invariant {
 
 			// convert to shares and add to the campaign shares - since we are converting shares to vouchers earlier,
 			// this conversion back to shares will never fail by design, thus we can ignore the error
-			vShares, _ := types.VouchersToShares(vouchers, campaignID)
+			vShares, _ := types.VouchersToShares(vouchers, campaignID) //nolint
```
lubtd, line 141: Github Actions says `directive `//nolint` is unused (nolintlint)` do we actually need a `nolint` here?

I wouldn't think not handling an error is a listing issue

lubtd (dismissed):
One comment but not blocking

x/campaign/keeper/invariants.go
giunatale: right, I should remove it

lubtd (approved):

aljo242 (approved):

PR #709, opened 2022-04-11
feat(`reward`): simulate MsgSetRewards

### LINKED ISSUE #703: Implement simulation for `MsgSetRewards`
We should implement a method to return a chain with the associated coordinator and an eventual existing reward pool

Ideally we have simulation where
- new reward pools are created
- some are updated
- some are deleted (setting rewards or height to 0)

aljo242:
Closes #703

### What does this PR does?

- Simulate `MsgSetRewards` with 3 scenarios: create pool, edit pool, close pool
- Updates simulation files that used a `customFee` param.  Instead, we now populate the `CoinsSpentInMsg` param for the `txCtx`.  This essentially performs the same function we were doing with custom logic.
- Add tests for the new `store` method added in the `simulation` reward package

### How to test?

- run the `make` targets


x/reward/simulation/simulation.go
```diff
@@ -0,0 +1,95 @@
...
+		} else {
+			if closeRewardPool {
+				msg.LastRewardHeight = 0
+				msg.Coins = sdk.NewCoins()
+			} else {
+				pool, _ := k.GetRewardPool(ctx, chain.LaunchID)
+				msg.LastRewardHeight = pool.LastRewardHeight + r.Int63n(1000)
+				msg.Coins = wantCoin
+			}
+		}
```
lubtd, line 76: To make it more visual, I would suggest using an enum
`type Action = Create|Update|Delete`

`func getAction(r *rand.Rand) Action` returns one action depending on the probabilities

Then we can use `switch` here

aljo242: The change I made did not introduce a `getAction()` but I think this sufficiently simplifies the code to make it more readable.

giunatale (approved):
lgtm

lubtd (approved):

PR #711, opened 2022-04-12
feat(`campaign`): prevent adding shares to account after mainnet launch is triggered

### LINKED ISSUE #707: Prevent adding shares to accounts once the launch is triggered for campaign mainnet
To make genesis generation deterministic after the launch is triggered for the mainnet. Mainnet genesis accounts should not be modified. Therefore, all messages interacting with `MainnetGenesisAccount` and `MainnetVestingAccount` should fail if the mainnet launch is triggered.

Messages where the condition must be added:
```
MsgAddShares
MsgAddVestingOptions
MsgRedeemVouchers
MsgUnredeemVouchers
```

```
if campaign.MainnetInitialized
  if launchTriggered(campaign.MainnetChainID)
    throw
```

giunatale:
Closes #707 

I have added a function `IsCampaignMainnetLaunchTriggered` to the `campaign` keeper since I could not use
the `launch` keeper in the AppModule for simulation due to a cyclic import (see the expected keepers of `launch`).

I think it should work fine this way, but let me know your thoughts.


x/campaign/keeper/msg_add_shares.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 
+	spnerrors "github.com/tendermint/spn/pkg/errors"
```
aljo242, line 7: goimports

x/campaign/keeper/campaign_mainnet_test.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	testkeeper "github.com/tendermint/spn/testutil/keeper"
+	"github.com/tendermint/spn/testutil/sample"
+)
+
+func TestIsCampaignMainnetLaunchTriggered(t *testing.T) {
```
lubtd, line 12: I believe we are missing a test case where `MainnetInitialized == false` where the method must return `false`

x/campaign/keeper/msg_add_shares.go
```diff
@@ -19,6 +21,17 @@ func (k msgServer) AddShares(goCtx context.Context, msg *types.MsgAddShares) (*t
 		return nil, sdkerrors.Wrapf(types.ErrCampaignNotFound, "%d", msg.CampaignID)
 	}
 
+	mainnetLaunched, err := k.IsCampaignMainnetLaunchTriggered(ctx, campaign.CampaignID)
+	if err != nil {
+		return nil, spnerrors.Critical(err.Error())
+	}
+	if mainnetLaunched {
+		return nil, sdkerrors.Wrap(types.ErrMainnetLaunchTriggered, fmt.Sprintf(
+			"mainnet %d is already launched, action prohibited",
```
lubtd, line 30:
```suggestion
			"mainnet %d launch is already triggered",
```

If think `action prohibited` is not necessary, the action is prohibited since the message fails

x/campaign/keeper/msg_add_vesting_options.go
```diff
@@ -20,6 +20,17 @@ func (k msgServer) AddVestingOptions(goCtx context.Context, msg *types.MsgAddVes
 		return nil, sdkerrors.Wrapf(types.ErrCampaignNotFound, "%d", msg.CampaignID)
 	}
 
+	mainnetLaunched, err := k.IsCampaignMainnetLaunchTriggered(ctx, campaign.CampaignID)
+	if err != nil {
+		return nil, spnerrors.Critical(err.Error())
+	}
+	if mainnetLaunched {
+		return nil, sdkerrors.Wrap(types.ErrMainnetLaunchTriggered, fmt.Sprintf(
+			"mainnet %d is already launched, action prohibited",
```
lubtd, line 29: Same feedback as above for all errors

aljo242 (changes requested):

aljo242 (approved):

lubtd:
The CLI works well👍

lubtd (approved):

PR #727, opened 2022-04-14
fix(`profile`): ensure addresses are not equal for MsgAddValidatorOperatorAddress sim

### LINKED ISSUE #725: Simulate `MsgAddValidatorOperatorAddress` sometimes fail
https://github.com/tendermint/spn/runs/6023479811?check_suite_focus=true

Fails with
```
validator profile address and operator address must be different: address is duplicated 
```

We should ensure the `OperatorAddress` and `ValidatorAddress` used in the message are not the same

aljo242:
Closes #725

### What does this PR does?

- Add check to ensure that sim addresses are not the same

### DIFF
--- a/x/profile/simulation/simulation.go
+++ b/x/profile/simulation/simulation.go
@@ -108,12 +108,16 @@ func SimulateMsgAddValidatorOperatorAddress(ak types.AccountKeeper, bk types.Ban
 	return func(r *rand.Rand, app *baseapp.BaseApp, ctx sdk.Context, accs []simtypes.Account, chainID string,
 	) (simtypes.OperationMsg, []simtypes.FutureOperation, error) {
 
+		// choose two addresses that are not equal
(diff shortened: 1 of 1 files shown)


giunatale (approved):

giunatale:
question: should the `ValidatorOperatorAddress` not be a user (e.g. `spn1...`) address but rather a `valoper` (e.g. `spnvaloper1...`) address?

lubtd (approved):

lubtd:
> question: should the `ValidatorOperatorAddress` not be a user (e.g. `spn1...`) address but rather a `valoper` (e.g. `spnvaloper1...`) address?

We use `spn` prefix for simplicity since the message to add this address must contact a `AccAddress` to check signature. The `xxvaloper..` format is what you will have on the testnet, the address is converted when the monitoring IBC packet is received

PR #730, opened 2022-04-19
test(`monitoringc`): increase code coverage

giunatale:
Part of #687

### What does this PR does?

Increase code coverage for `x/monitoringc`

note: code coverage is not increased for `types/params.go` since the module does not have params and the file only contains boilerplate code


aljo242 (approved):

proto/types/ibc.proto
```diff
@@ -5,8 +5,8 @@ import "gogoproto/gogo.proto";
 
 option go_package = "github.com/tendermint/spn/pkg/types";
 
-// MerkelRool represents a Merkel Root in ConsensusState
-message MerkelRool {
+// MerkleRoot represents a Merkel Root in ConsensusState
```
lubtd, line 8:
```suggestion
// MerkleRoot represents a Merkle Root in ConsensusState
```

x/monitoringc/keeper/monitoring_packet_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := tk.MonitoringConsumerKeeper.OnRecvMonitoringPacket(ctx, tt.packet, tt.data)
+			if !tt.valid {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+		})
+	}
+}
```
lubtd, line 139: We can also add test cases to ensure `OnAcknowledgementMonitoringPacket` and `OnTimeoutMonitoringPacket` return errors

lubtd (commented):
👍

For the onrecv test with rewards distributions, I think we should rely on mocks at some point but it would concern a bigger refactoring for the `testkeeper` package that include mocked keeper dependencies

giunatale:
> 👍
> 
> For the onrecv test with rewards distributions, I think we should rely on mocks at some point but it would concern a bigger refactoring for the `testkeeper` package that include mocked keeper dependencies

I agree 100%.

lubtd (approved):

aljo242 (approved):

PR #736, opened 2022-04-20
test(`monitoringp`): increase code coverage

aljo242:
Part of https://github.com/tendermint/spn/issues/687

### What does this PR does?

Increase code coverage for `x/monitoringp`


x/monitoringp/keeper/begin_block_test.go
```diff
@@ -235,3 +235,111 @@ func TestKeeper_ReportBlockSignatures(t *testing.T) {
...
+		lastBlockHeight             int64
+		currentBlockHeight          int64
+		channelIDExist              bool
+		channelID                   types.ConnectionChannelID
+		expectedMonitoringInfoFound bool
+		expectedMonitoringInfo      types.MonitoringInfo
+		wantErr                     bool
+	}{
+		{
+			name:                        "currentBlockHeight < lastBlockHeight returns nil",
```
lubtd, line 277:
```suggestion
			name:                        "should perform no action if last block height is not reached",
```

lubtd (approved):
👍
I think we can create another issue to address `monitoring_packet.go` and `TransmitSignatures` where the packet is actually transmitted and that will require mocking `TransmitMonitoringPacket ` method so we can close https://github.com/tendermint/spn/issues/687 after this current issue and coverage for `monitoringc`

giunatale (approved):

PR #740, opened 2022-04-21
fix: usage of `rand` in simulation test

lubtd:
Change `rand` usage with `r` variable to prevent non-determinism issue in simulation tests

### DIFF
--- a/x/campaign/simulation/store.go
+++ b/x/campaign/simulation/store.go
@@ -137,6 +137,7 @@ func GetSharesFromCampaign(r *rand.Rand, ctx sdk.Context, k keeper.Keeper, campI
 
 // GetAccountWithVouchers returns an account that has vouchers for a campaign
 func GetAccountWithVouchers(
+	r *rand.Rand,
 	ctx sdk.Context,
 	bk types.BankKeeper,
 	k keeper.Keeper,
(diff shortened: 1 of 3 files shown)


x/campaign/simulation/store.go
```diff
@@ -179,7 +180,7 @@ func GetAccountWithVouchers(
 		coinCampID, err := types.VoucherCampaign(coin.Denom)
 		if err == nil && coinCampID == campID {
 			// retain a random portion of the balance in the range [0, coin.Amount)
-			retainAmt := sdk.NewInt(rand.Int63n(coin.Amount.Int64()))
+			retainAmt := sdk.NewInt(r.Int63n(coin.Amount.Int64()))
```
giunatale, line 183: I would consider using https://github.com/cosmos/cosmos-sdk/blob/master/types/simulation/rand_util.go#L43 instead

giunatale: or better probably `RandomAmount()`

giunatale (changes requested):

x/campaign/simulation/store.go
lubtd: Question: is there a specific reason we calculate `retainAmt` and subtract coin's amount from it, instead of directly set a random amount in `coins` structure?

aljo242: I think its because this is just a bit simpler than building a new `coin` with the same `denom` as the `coin` we are operating on

lubtd: I changed the logic a bit to avoid 0 token that trigger errors since `RandPositiveInt` include the max amount in the range

aljo242 (approved):

x/campaign/simulation/store.go
giunatale: nope, what you did seems perfect

giunatale (approved):

PR #742, opened 2022-04-21
Feat/profile events

### LINKED ISSUE #731: Implement events for `profile`
Implement typed events for the `profile` module based on the different messages of the module

aljo242:
Closes #731

### What does this PR does?

Emit typed events for the `profile` module.

The events currents are just Create* and Update*.   This differs from the Tx that we expose, but the state is still fully covered by these events


lubtd (commented):
I'm honestly not sure about the create/updated events
This would require context to understand the state change.
For example, if you have a Coordinator Updated event, you can't determine directly from the event data if the address of the coordinator or the coordinator description have been changed, you need to get the previous state of the coordinator

What is the downside of having per-message events?

aljo242:
> I'm honestly not sure about the create/updated events This would require context to understand the state change. For example, if you have a Coordinator Updated event, you can't determine directly from the event data if the address of the coordinator or the coordinator description have been changed, you need to get the previous state of the coordinator
> 
> What is the downside of having per-message events?

When any Msg is run in the Cosmos SDK, a basic "action" event is emitted, stating the type of the message.  The code can be found [here](https://github.com/cosmos/cosmos-sdk/blob/17279fdf300a125ce62e58ca8594795e13143f29/x/auth/middleware/run_msgs.go#L91).

So, in `spn` when you run any Tx, we already emit this basic event.  An  example can he seen here:

```shell

spnd tx profile create-coordinator --from steve 
{"body":{"messages":[{"@type":"/tendermint.spn.profile.MsgCreateCoordinator","address":"spn1mhyps2hlkm0nz6k2puumn69928cnvgg43zlrg7","description":{"identity":"","website":"","details":""}}],"memo":"","timeout_height":"0","extension_options":[],"non_critical_extension_options":[]},"auth_info":{"signer_infos":[],"fee":{"amount":[],"gas_limit":"200000","payer":"","granter":""}},"signatures":[]}

confirm transaction before signing and broadcasting [y/N]: y
code: 0
codespace: ""
data: 0A320A2C2F74656E6465726D696E742E73706E2E70726F66696C652E4D7367437265617465436F6F7264696E61746F7212020801
events:
- attributes:
  - index: true
    key: ZmVl
    value: ""
  type: tx
- attributes:
  - index: true
    key: YWNjX3NlcQ==
    value: c3BuMW1oeXBzMmhsa20wbno2azJwdXVtbjY5OTI4Y252Z2c0M3pscmc3LzA=
  type: tx
- attributes:
  - index: true
    key: c2lnbmF0dXJl
    value: RmNpZVRZNEdnbi9UUTdkNjlBR1hqdU9yZnB4TktNSHZTQzlILzlGamtWZC9VODRMVjlMdExVMHRUWFc5UERzTXNZUnNTakN0bXBkMjVNVXhYRisyNkE9PQ==
  type: tx
- attributes:
  - index: true
    key: YWN0aW9u
    value: Y3JlYXRlX2Nvb3JkaW5hdG9y
  type: message
gas_used: "61709"
gas_wanted: "200000"
height: "14"
info: ""
logs:
- events:
  - attributes:
    - key: action
      value: create_coordinator
    type: message
  log: ""
  msg_index: 0
raw_log: '[{"events":[{"type":"message","attributes":[{"key":"action","value":"create_coordinator"}]}]}]'
timestamp: ""
tx: null
txhash: 88ED7C296C1359217C1BF2E4909BFB6BD14F4281D28EC540D6DCEB4C5ED61200

```

My thinking is that we are providing additional events that detail how the state on-chain is being modified.  The "context" providing events are already there.

lubtd:
When fetching or listening for events, can you get the hash of the tx where the event is emitted? If yes, we can actually associate the message name event and custom event to get the full state transition but that still seems more work from the client from my point of view.
If a coordinator's address is updated, for example, you would just need to have the previous address and new address to interpret the state transition, the description doesn't matter

proto/profile/events.proto
```diff
@@ -0,0 +1,25 @@
...
+package tendermint.spn.profile;
+
+import "gogoproto/gogo.proto";
+import "profile/coordinator.proto";
+import "profile/validator.proto";
+
+
+option go_package = "github.com/tendermint/spn/x/profile/types";
+
+message EventCoordinatorCreated {
```
giunatale, line 11: I think it makes sense for these "custom" events to be more descriptive about the state transition (e.g. a `beforeValidator` and `afterValidator` in `EventValidatorUpdated`) instead of a single entry.

giunatale (changes requested):
Considering we already [emit events for messages](https://docs.cosmos.network/master/core/events.html#eventmanager), I agree with @aljo242's proposal of state-transition events. I would probably make them more informative of what changed though

proto/profile/events.proto
giunatale: I may have overengineerd this in my head. If you are tracing on-chain events you are pretty much guaranteed to have the previous data, so you only need to know the new one in the event.

giunatale (approved):
I've changed my mind. I like the way things are already.

@lubtd as a clarification (hopefully) events are always emitted as part of "raw_data" of transactions, which are comprised of a bunch of messages. Events are already bundled together per-message by the event manager, so they should be representing state-transitions that happens as the messages executes rather than be a recap of the message itself (since the data is included in the transaction anyway).

lubtd:
> so they should be representing state-transitions that happens as the messages executes rather than be a recap of the message itself (since the data is included in the transaction anyway).

Yes but in the case of `CoordinatorUpdated` we don't emit the state transition but the output state from the state transition. The state of the coordinator after the address or the description has been changed

giunatale:
The state transition happened on the coordinator object so you show how the object has changed (IMHO)

> Yes but in the case of `CoordinatorUpdated` we don't emit the state transition but the output state from the state transition. The state of the coordinator after the address or the description has been changed

Not sure I follow. could you clarify? what would practically be what you would emit in the event?

aljo242:
> > so they should be representing state-transitions that happens as the messages executes rather than be a recap of the message itself (since the data is included in the transaction anyway).
> 
> Yes but in the case of `CoordinatorUpdated` we don't emit the state transition but the output state from the state transition. The state of the coordinator after the address or the description has been changed

The state transmission is automatically already emitted with the `Action` event [emitted](https://docs.cosmos.network/master/core/events.html#eventmanager) for all transactions.  So, if a client is subscribing to an event stream, they will see:

- The type of message (already emitted per-tx)
- The associated state change (my addition)

Events are emitted and associated with transactions already, so associating them with a specific Tx is natural.

giunatale:
> The state transmission is automatically already emitted with the Action event [emitted](https://docs.cosmos.network/master/core/events.html#eventmanager) for all transactions. So, if a client is subscribing to an event stream, they will see:

```json
{"key":"action","value":"create_coordinator"}
```

in the example above

lubtd:
> Events are emitted and associated with transactions already, so associating them with a specific Tx is natural.

Ok, this is what I wanted to ensure.
I may be a bit biased with Solidity where you subscribe for a specific event type.
In any case, we can have feedback from Front-End in the future

x/profile/keeper/msg_update_validator_description.go
```diff
@@ -40,6 +40,9 @@ func (k msgServer) UpdateValidatorDescription(
 	}
 
 	k.SetValidator(ctx, validator)
+	err := ctx.EventManager().EmitTypedEvent(
```
lubtd, line 43: The message follows the same logic as `MsgAddValidatorOperatorAddress`.If the validator didn't exist, it is created and therefore should be a `ValidatorCreated` event

lubtd (changes requested):

lubtd (dismissed):

x/reward/keeper/reward_distribution_test.go
```diff
@@ -2,9 +2,10 @@ package keeper_test
 
 import (
 	"fmt"
-	spnerrors "github.com/tendermint/spn/pkg/errors"
 	"testing"
 
+	spnerrors "github.com/tendermint/spn/pkg/errors"
```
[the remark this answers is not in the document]
lubtd: Can we change this @aljo242 ?

lubtd:
Seems like we have some test errors since last `develop` merge

aljo242:
> Seems like we have some test errors since last `develop` merge

Tests pass now - just a CodeCov upload error

PR #743, opened 2022-04-22
test: ensure only `sdk.DefaultBondDenom` is used for fees in our modules simulation

### LINKED ISSUE #369: Simulation test: `out of gas` error for some simulation

giunatale:
Closes #369 

This PR does not really fix #369, as explained in the issue itself.
However, we at least switch over using only the `sdk.DefaultBondDenom` coin for fees in our modules' simulation.
Problem will stil exist however for modules we pull from the cosmos-sdk, so `out of gas` will still occasionally occur (and it's highly likely to occur for large simulations), but I propose we close #369 as this can't really be avoided.

I've also made sure we initialize the `rand` seed in `TestAppStateDeterminism`, and fixed an error in how we were selecting auctions for `msgParticipate` simulation. 

Apologies if I may have bundled too many things together in this PR.

### DIFF
--- a/app/simulation_test.go
+++ b/app/simulation_test.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/codec"
@@ -50,7 +51,6 @@ func interBlockCacheOpt() func(*baseapp.BaseApp) {
 	return baseapp.SetInterBlockCache(store.NewCommitKVStoreCacheManager())
 }
 
-// go test ./app -v -benchmem -run=^$ -bench ^BenchmarkSimulation -Commit=true -cpuprofile cpu.out
 func BenchmarkSimulation(b *testing.B) {
 	simapp.FlagVerboseValue = true
 	simapp.FlagOnOperationValue = true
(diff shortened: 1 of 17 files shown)


aljo242 (approved):
Do we need `RandomAuctionStandby` at this point?

giunatale:
> Do we need `RandomAuctionStandby` at this point?

Yes, for `msgCancelAuction`

app/simulation_test.go
```diff
@@ -118,6 +118,7 @@ func TestAppStateDeterminism(t *testing.T) {
 	config.OnOperation = true
 	config.AllInvariants = true
 
+	rand.Seed(time.Now().Unix())
```
lubtd, line 121: Is it for the simulation seed? Can't we have a flag for it?

giunatale: we use `rand.Int63()` below

We could use the seed flag, but why? We are not seeking repeatability, but rather find non-determinism. @aljo242 opinions?

I could use the `rand.Seed(simapp.FlagSeedValue)` in any case, if wanted

lubtd (approved):

PR #752, opened 2022-04-27
feat(`reward`): emit events

### LINKED ISSUE #732: Implement events for `reward`
Implement typed events for the `reward` module based on the different messages of the module, `SetRewards` in this case

Eventually, it could make sense to have an event for rewards distribution as well

aljo242:
Closes #732

Emits events for `Create`, `Update`, `Remove` actions on the `RewardPool` type.  Additionally emits `EventRewardsDistributed` during `k.DistributeRewards()`.


x/reward/keeper/msg_set_reward.go
```diff
@@ -57,25 +57,31 @@ func (k msgServer) SetRewards(goCtx context.Context, msg *types.MsgSetRewards) (
 		}
 	}
 
-	if msg.Coins.Empty() || msg.LastRewardHeight == 0 {
+	if (msg.Coins.Empty() || msg.LastRewardHeight == 0) && found {
```
lubtd, line 60:
```suggestion
	if (msg.Coins.Empty() || msg.LastRewardHeight == 0) {
```

Previous behavior is right
If the pool doesn't exist and the user wants to remove it, nothing will happen because `RemoveRewardPool` is idempotent
Maybe an optimization is to make the message fails if `msg.Coins.Empty() || msg.LastRewardHeight == 0` and the pool doesn't exist

However adding `&& found` in this condition change the behavior because the reward pool would be then created with empty coins or last reward height set to 0

lubtd: > Maybe an optimization is to make the message fails if msg.Coins.Empty() || msg.LastRewardHeight == 0 and the pool doesn't exist

Actually that could make sense to do this if we now emit an event

x/reward/keeper/reward_distribution.go
```diff
@@ -138,7 +145,9 @@ func (k Keeper) DistributeRewards(
 		rewardPool.Closed = true
 		rewardPool.RemainingCoins = rewardPool.RemainingCoins.Sub(rewardPool.RemainingCoins) // sub coins transferred
 		k.SetRewardPool(ctx, rewardPool)
-		return nil
+
+		err = ctx.EventManager().EmitTypedEvent(&types.EventRewardPoolUpdated{RewardPool: rewardPool})
+		return err
```
lubtd, line 150:
```suggestion
		return ctx.EventManager().EmitTypedEvent(&types.EventRewardPoolUpdated{RewardPool: rewardPool})
```

x/reward/keeper/reward_distribution.go
```diff
@@ -173,7 +182,8 @@ func (k Keeper) DistributeRewards(
 	// update the current reward height for next reward
 	rewardPool.CurrentRewardHeight = lastBlockHeight
 	k.SetRewardPool(ctx, rewardPool)
-	return nil
+	err = ctx.EventManager().EmitTypedEvent(&types.EventRewardPoolUpdated{RewardPool: rewardPool})
+	return err
```
lubtd, line 186:
```suggestion
	return ctx.EventManager().EmitTypedEvent(&types.EventRewardPoolUpdated{RewardPool: rewardPool})
```

x/reward/keeper/msg_set_reward.go
aljo242: I think its okay just remove this extra check I added.  I removed the `EventRewardPoolUpdated` as it is unecessary

x/reward/keeper/msg_set_reward.go
```diff
@@ -57,25 +57,32 @@ func (k msgServer) SetRewards(goCtx context.Context, msg *types.MsgSetRewards) (
 		}
 	}
 
-	if msg.Coins.Empty() || msg.LastRewardHeight == 0 {
+	if (msg.Coins.Empty() || msg.LastRewardHeight == 0) {
```
lubtd, line 60:
```suggestion
	if msg.Coins.Empty() || msg.LastRewardHeight == 0 {
```

x/reward/keeper/msg_set_reward.go
```diff
@@ -57,25 +57,32 @@ func (k msgServer) SetRewards(goCtx context.Context, msg *types.MsgSetRewards) (
...
 		rewardPool.LastRewardHeight = 0
 		k.RemoveRewardPool(ctx, msg.LaunchID)
+		err = ctx.EventManager().EmitTypedEvent(&types.EventRewardPoolRemoved{LaunchID: msg.LaunchID})
 	} else {
 		rewardPool.InitialCoins = msg.Coins
 		rewardPool.RemainingCoins = msg.Coins
 		rewardPool.Provider = msg.Provider
 		rewardPool.LastRewardHeight = msg.LastRewardHeight
 		k.SetRewardPool(ctx, rewardPool)
+		if !found {
```
lubtd, line 72:
```suggestion
		if !rewardPoolFound {
```
To keep context since `found` is initialize above

lubtd (approved):

giunatale (approved):

PR #760, opened 2022-05-04
chore: remove `SendVoucher` operation in simulation

lubtd:
Remove it since `bank` simulation perform sending coins operation


giunatale (approved):
nice 👍

PR #770, opened 2022-05-06
feat(`campaign`): add `AuctionsOfCampaign` query

### LINKED ISSUE #756: Create a query to show a list of fundraiser IDs associated with a particular campaign ID

giunatale:
Closes #756


x/campaign/keeper/grpc_auctions_by_campaign.go
```diff
@@ -0,0 +1,38 @@
...
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	// not performing any existence check on the campaign as it's not needed
+	reqCampaignID := req.CampaignID
+	resAuctionIDs := make([]uint64, 0)
+
+	k.fundraisingKeeper.IterateAuctions(ctx, func(auction fundraisingtypes.AuctionI) (stop bool) {
```
lubtd, line 24: We should also add that the auctioneer is the coordinator of the campaign asa condition

giunatale: It cannot work. If the auction has been created and the coordinator is updated, it would break. I thought about it and I think it could break very easily as a check

giunatale: Would be enforceable as a check only if we had the coordinator ID (not the coordinator address) stored in the auction data

lubtd: True

Maybe it's fine to not have this check if we can remove this query in the near future depending on advancement on the backend service

giunatale: We might also use the `created` [hooks](https://github.com/tendermint/fundraising/blob/main/x/fundraising/keeper/auction.go#L68) to:
1. get the campaign from the selling denom
2. check that at creation the coordinator of the campaign is the auction creator (so we can even prevent auctions creation if the coordinator is not the auctioneer)
3. store a direct mapping of `auctionIDs` to `campaignID` so fetching this info might be extremely faster

giunatale: but this would widen the scope of this PR, and change the logic a bit. So LMK your thoughts

x/campaign/keeper/grpc_auctions_by_campaign.go
```diff
@@ -0,0 +1,38 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/tendermint/spn/x/campaign/types"
+)
+
+func (k Keeper) AuctionsOfCampaign(goCtx context.Context, req *types.QueryAuctionsOfCampaignRequest) (*types.QueryAuctionsOfCampaignResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
```
lubtd, line 18: I think we should return an error not found if the campaign doesn't exist

lubtd (commented):
Just one comment. Otherwise the CLI works well👍

x/campaign/keeper/grpc_auctions_by_campaign.go
lubtd: As discussed in the call. Let's omit coordinator verification for now

x/campaign/keeper/grpc_auctions_by_campaign.go
giunatale: I did this on purpose. It should just return an empty list as of now, right?
Why the error should be better? It's a query, not a tx. You would know if the campaign exist already by querying for the campaign itself.

lubtd: That's a small optimization, you would not iterate auctions if the campaign doesn't exist and therefore no auction would be found

proto/campaign/query.proto
```diff
@@ -72,7 +72,12 @@ service Query {
     option (google.api.http).get = "/tendermint/spn/campaign/total_shares";
   }
 
-  // this line is used by starport scaffolding # 2
+  // Queries a list of auction IDs of a campaign
+	rpc AuctionsOfCampaign(QueryAuctionsOfCampaignRequest) returns (QueryAuctionsOfCampaignResponse) {
+		option (google.api.http).get = "/tendermint/spn/campaign/auctions_of_campaign/{campaignID}";
+	}
+
```
aljo242, line 79:
```suggestion
  // Queries a list of auction IDs of a campaign
  rpc AuctionsOfCampaign(QueryAuctionsOfCampaignRequest) returns (QueryAuctionsOfCampaignResponse) {
	option (google.api.http).get = "/tendermint/spn/campaign/auctions_of_campaign/{campaignID}";
  }

```

Format in line with all other `rpc`

x/campaign/client/cli/query_auctions_by_campaign.go
```diff
@@ -0,0 +1,49 @@
...
+	"strconv"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/spf13/cobra"
+
+	"github.com/tendermint/spn/x/campaign/types"
+)
+
+var _ = strconv.Itoa(0)
```
aljo242, line 13: ### 
```suggestion
```

ok to remove

aljo242 (changes requested):

lubtd (approved):

PR #771, opened 2022-05-08
refactor: upgrade `fundraising` to `0.3.0`

lubtd:
- Upgrade `fundraising` import version to `0.3.0`
- Remove create and cancel auction operations in `participation` as `fundraising` now implements its own simulations


giunatale (changes requested):
We have a verification error in `msg_participate_test.go` (github reporting this)
`auction.GetAllowedBidders undefined (type "github.com/tendermint/fundraising/x/fundraising/types".AuctionI has no field or method GetAllowedBidders)`

Also, did we check the impact of removing our simulations wrt the ones defined in `fundraising`? Especially for `create`, which was made ad-hoc for our simulation use-cases

lubtd:
>We have a verification error in msg_participate_test.go (github reporting this)
auction.GetAllowedBidders undefined (type "github.com/tendermint/fundraising/x/fundraising/types".AuctionI has no field or method GetAllowedBidders)

Yes, actually. I was assuming that errors were due to `listen tcp 0.0.0.0:45545: bind: address already in use` 🤦‍♂️ Which we are getting more and more

> Also, did we check the impact of removing our simulations wrt the ones defined in fundraising? Especially for create, which was made ad-hoc for our simulation use-cases

We get successful simulations of `MsgParticipate` and `MsgWithdrawAllocations` with simulations defined in `fundraising`

aljo242 (approved):

PR #778, opened 2022-05-09
feat(`participation`): emit events

### LINKED ISSUE #735: Implement events for the `participation` module
Implement typed events for the`participation` module based on the different messages of the module

aljo242:
Closes #735


proto/participation/events.proto
```diff
@@ -0,0 +1,17 @@
...
+
+option go_package = "github.com/tendermint/spn/x/participation/types";
+
+message EventAllocationsUsed {
+  string participant = 1;
+  uint64 auctionID = 2;
+  uint64 numAllocations = 3;
+}
+
+message EventAllocationsWithdrawn {
```
lubtd, line 14: We could maybe add `numAllocations` as well?

aljo242: The number of allocations withdrawn from the should always be the same `numAllocations` emitted in the corresponding `EventAllocationsUsed` message.  I think this would just be adding redundant data.

lubtd (approved):

giunatale (approved):

PR #786, opened 2022-05-10
feat(`campaign`): add special allocations

lubtd:
Implementing special allocations structure and add it to campaign

- Type `SpecialAllocations`
- Add tests for type
- Add empty by default special allocations to campaign
- Update `CampaignShares` invariant to include special allocations


x/campaign/types/campaign_test.go
```diff
@@ -45,6 +45,17 @@ func TestCampaign_Validate(t *testing.T) {
...
+		sample.Shares(r),
+		campaign.Shares(sdk.NewCoins(
+			sdk.NewCoin("foo", sdk.NewInt(100)),
+			sdk.NewCoin("s/bar", sdk.NewInt(200)),
+		)),
+	)
+	require.Error(t, invalidSpecialAllocations.Validate())
+	campaignInvalidSpecialAllocations := sample.Campaign(r, 0)
+	campaignInvalidSpecialAllocations.SpecialAllocations = invalidSpecialAllocations
+
```
aljo242, line 58: I think we should add a test where we are validating a "non-empty" special allocations field in the `campaign` object.

The best way to do this is to probably modify `sample.Campaign()` to automatically make that field non-null but valid

aljo242 (changes requested):

x/campaign/types/campaign.go
```diff
@@ -1,10 +1,10 @@
 package types
 
 import (
-	"errors"
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pkg/errors"
```
giunatale, line 7: was this dependency substitution intended?

giunatale (approved):
I am pre-approving, pending answer to my question above. Ping me if I need to re-review this

x/campaign/types/campaign.go
lubtd: `github.com/pkg/errors` provides the `errors.Wrap` utility

giunatale:
Unit tests are failing, it's not due to the usual port-binding issue.

lubtd:
Simulation tests failing now with something we never had🤔

aljo242 (approved):

PR #788, opened 2022-05-11
chore(`campaign`): disable `MsgAddVestingOptions`

lubtd:
Disable this message since from a client point of view it is more convenient to handle vesting options with special allocations


giunatale (changes requested):
minor linting nitpicks

giunatale (approved):

PR #793, opened 2022-05-12
feat(`campaign`): implement `MainnetAccountBalanceAll` and `MainnetAccountBalance` queries

lubtd:
Implement the queries `MainnetAccountBalanceAll` and `MainnetAccountBalance`
The following queries return accounts for a mainnet with final balance instead of shares
The queries are used by clients to generate the genesis for a mainnet

Implement `CoinsFromTotalSupply` to returns coins from shares depending on the total supply in campaign


x/campaign/types/shares.go
```diff
@@ -99,7 +99,44 @@ func (shares Shares) Empty() bool {
...
+			// coin balance = (supply * share) / total share
+			coinBalance := (supply.Amount.Mul(amount)).Quo(sdk.NewIntFromUint64(totalShareNumber))
+
+			if !coinBalance.IsZero() {
+				coins = append(coins, sdk.NewCoin(supply.Denom, coinBalance))
+			}
+		}
+	}
+
+	return
```
giunatale, line 141:
```suggestion
```

giunatale (changes requested):
minor linting error. rest looking good 👍 

Can you also do a `fmt` pass while addressing this? thanks

x/campaign/types/shares.go
lubtd: Weird, Goland doesn't tell me anything on this

giunatale (approved):

PR #794, opened 2022-05-13
feat(`mint`): add `mint` module

### LINKED ISSUE #780: Fork `mint` module
Fork `mint` module in our own codebase
Import this custom `mint` module into our app

### LINKED ISSUE #781: `mint` module: introduce a parameter for community pool block rewards
Depends on https://github.com/tendermint/spn/issues/780

Introduce parameter for the percentage of block rewards to be forwarded into the community pool

giunatale:
Closes #780 
Closes #781

### What does this PR does?

Adds a forked version of the `mint` module.
Unfortunately I could not fork the osmosis module since it depends on their epoch module, and so it was unfit.
I opted to start with the vanilla mint module as a baseline, with the dynamic inflation rate. In the future, we could explore some capping mechanism, for instance a reduction after a year worth of blocks as done in the juno [`mint`](https://github.com/CosmosContracts/juno/blob/main/x/mint/abci.go#L29) module.

The module distributes inflation to 3 parties:
- staking
- the future-to-be module account of the `incentive` modules (currently commented out in `keeper.go`)
   the idea is to send rewards directly to the `incentives` module account so that the module itself can handle the accumulated rewards independently.
- community pool

I have opted to not include at this time a "developers" rewards or "strategic reserve" rewards because this need to be thought out later. We could explore a similar design as per osmosis where a weighted list of addresses, that can be changed through governance, receive rewards from inflation. Not sure if doing this per-block is efficient though. Hence, leaving it aside.


lubtd (commented):
LGTM staking rewards can be distributed

I would suggest setting params for the module in `config.yml`

lubtd (approved):

x/mint/module.go
```diff
@@ -0,0 +1,186 @@
...
+	_ module.AppModuleBasic      = AppModuleBasic{}
+	_ module.AppModuleSimulation = AppModule{}
+)
+
+// AppModuleBasic defines the basic application module used by the mint module.
+type AppModuleBasic struct {
+	cdc codec.Codec
+}
+
+var _ module.AppModuleBasic = AppModuleBasic{}
```
aljo242, line 38:
```suggestion
```

This is duplicate no?

aljo242 (changes requested):
This and a minor lint issue need to be fixed imo.  I can just push these minor changes.

aljo242 (approved):

lubtd (approved):

PR #797, opened 2022-05-13
feat(`claim`): scaffold `claim` module

### LINKED ISSUE #782: Initialize the `claim` module
Initialize `claim` module with Ignite CLI without parameters or genesis state

The module has `bank` dependency

Eventually, some other dependency would be introduced in the future (staking, etc.. depending on actions) they can be dynamically added in the future

giunatale:
Closes #782


x/claim/client/cli/query.go
```diff
@@ -0,0 +1,31 @@
+package cli
+
+import (
+	"fmt"
+	// "strings"
+
+	"github.com/spf13/cobra"
+
```
lubtd, line 8:
```suggestion
```

aljo242 (approved):
I will format then re-approve

aljo242 (dismissed):

PR #802, opened 2022-05-17
feat(`campaign`): implement query `special-allocations-balance`

lubtd:
`special-allocations-balance` get the final token balance for special allocation.

This query is to be used by the client to verify the genesis is correctly formatted:

```
bank:
 supply == genesisDistribution
claim:
 airdropSupply == claimableAirdrop
```


aljo242 (changes requested):
We should have test cases for the two times that `CoinsFromTotalSupply()` can throw an error in this function.

x/campaign/keeper/grpc_special_allocations_balance.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/tendermint/spn/x/campaign/types"
+)
+
+func (k Keeper) SpecialAllocationsBalance(goCtx context.Context, req *types.QuerySpecialAllocationsBalanceRequest) (*types.QuerySpecialAllocationsBalanceResponse, error) {
```
Pantani, line 13:
```suggestion
func (k Keeper) SpecialAllocationsBalance(
	goCtx context.Context,
	req *types.QuerySpecialAllocationsBalanceRequest,
) (*types.QuerySpecialAllocationsBalanceResponse, error) {
```

PR #810, opened 2022-05-25
fix(monitoringc): fix not found query response for `VerifiedClientIds`

Pantani:


lubtd (approved):

lubtd (commented):
Looks good

Can we just change the expected error in `grpc_verified_client_ids_test.go`?

lubtd (approved):

PR #822, opened 2022-05-30
feat: Implement peer empty connection

ivanovpetr:


lubtd:
I'm trying to figure out if the NodeID, `id` in `Peer`, is still necessary for empty connection or if it could be moved into `tcpAddress` and `httpTunnel`

The node always has a node ID in the P2P network even if the public address is not exposed. But potentially this information is not necessary on-chain

lubtd (approved):

aljo242 (changes requested):

Pantani (approved):

ilgooz:
> I'm trying to figure out if the NodeID, `id` in `Peer`, is still necessary for empty connection or if it could be moved into `tcpAddress` and `httpTunnel`
> 
> The node always has a node ID in the P2P network even if the public address is not exposed. But potentially this information is not necessary on-chain

Good point, any updates on this?

In such case we don't need this PR, we can just NOT set the `peer` field in `GenesisValidator` which makes sense to me.

lubtd:
>Good point, any updates on this?
In such case we don't need this PR, we can just NOT set the peer field in GenesisValidator which makes sense to me.

We let it for now as the Node ID is a general information

lubtd (approved):

PR #824, opened 2022-05-30
feat(`events`): add coordinator address and ID for chain and campaign creation

lubtd:
Add for both `EventChainCreated` and `EventCampaignCreated` the address and ID of the coordinator for indexation purpose


jeronimoalbi:
> @jeronimoalbi @ilgooz tell me if you have any comment

looks good 👍 !

PR #828, opened 2022-06-01
fix(monitoring): `channel not found` error with IBC v3

lubtd:
IBC v3 brings changes that were breaking with our current logic for client verification during handshake: https://github.com/cosmos/ibc-go/blob/main/docs/migrations/v2-to-v3.md

The channel is not available in the store during `OnChanOpenInit` and `OnChanOpenTry` 

This PR modifies the logic to rely on the connection hops provided in the handler to retrieve the client ID to verify


aljo242:
Just run `make format` before merging !

PR #841, opened 2022-06-08
feat(`campaign`): add event for campaign auction creation

lubtd:
Add the event for campaign auction creation using `fundraising` hooks.

The event associates a campaign ID with an auction ID.


proto/campaign/events.proto
```diff
@@ -86,3 +86,8 @@ message EventMainnetVestingAccountUpdated {
   string address = 2;
   ShareVestingOptions vestingOptions = 3 [(gogoproto.nullable) = false];
 }
+
+message EventCampaignAuctionCreated {
+  uint64 campaignID = 1;
+  uint64 auctionID = 2;
+}
```
aljo242, line 93:
```suggestion
}

```

x/campaign/keeper/campaign.go
```diff
@@ -70,6 +70,12 @@ func (k Keeper) GetCampaign(ctx sdk.Context, id uint64) (val types.Campaign, fou
 	return val, true
 }
 
+// RemoveCampaign removes a campaign from the store
+func (k Keeper) RemoveCampaign(ctx sdk.Context, id uint64) {
```
aljo242, line 74: Is this function only being used internally for tests?

aljo242 (changes requested):
Check out the lint report as well

go.sum
```diff
@@ -1504,8 +1504,8 @@ github.com/tendermint/btcd v0.1.1 h1:0VcxPfflS2zZ3RiOAHkBiFUcPvbtRj5O7zHmcJWHV7s
 github.com/tendermint/btcd v0.1.1/go.mod h1:DC6/m53jtQzr/NFmMNEu0rxf18/ktVoVtMrnDD5pN+U=
 github.com/tendermint/crypto v0.0.0-20191022145703-50d29ede1e15 h1:hqAk8riJvK4RMWx1aInLzndwxKalgi5rTqgfXxOxbEI=
 github.com/tendermint/crypto v0.0.0-20191022145703-50d29ede1e15/go.mod h1:z4YtwM70uOnk8h0pjJYlj3zdYwi9l03By6iAIF5j/Pk=
-github.com/tendermint/fundraising v0.3.0 h1:VtHfmVlAS93MUDlt6Em21l3taw6s9kLY/w8Cd1FB9fM=
-github.com/tendermint/fundraising v0.3.0/go.mod h1:oJFZUZ/GsACtkYeWScKpHLdqMUThNWpMAi/G47LJUi4=
+github.com/tendermint/fundraising v0.3.1-0.20220613014523-03b4a2d4481a h1:DIxap6r3z89JLoaLp6TTtt8XS7Zgfy4XACfG6b+4plE=
```
Pantani, line 1507: can we remove the commit version and use only v0.3.1?

x/campaign/keeper/campaign_auction_event_hooks.go
```diff
@@ -0,0 +1,199 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	fundraisingtypes "github.com/tendermint/fundraising/x/fundraising/types"
+	"github.com/tendermint/spn/x/campaign/types"
+	profiletypes "github.com/tendermint/spn/x/profile/types"
+)
+
+// EmitCampaignAuctionCreated emits EventCampaignAuctionCreated event if an auction is created for a campaign from a coordinator
+func (k Keeper) EmitCampaignAuctionCreated(
+	ctx sdk.Context,
+	auctionId uint64,
```
Pantani, line 16:
```suggestion
	auctionID uint64,
```

x/campaign/keeper/campaign_auction_event_hooks.go
```diff
@@ -0,0 +1,199 @@
...
+	auctioneer string,
+	sellingCoin sdk.Coin,
+) (bool, error) {
+	campaignID, err := types.VoucherCampaign(sellingCoin.Denom)
+	if err != nil {
+		// not a campaign auction
+		return false, nil
+	}
```
Pantani, line 24:
```suggestion
	campaignID, err := types.VoucherCampaign(sellingCoin.Denom)
	if err == types.ErrNotAuctionCampaign {
		return false, nil	
	} else if err != nil {	
		return false, err
	}
```

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/spf13/cast v1.4.1
 	github.com/spf13/cobra v1.4.0
 	github.com/stretchr/testify v1.7.1
-	github.com/tendermint/fundraising v0.3.0
+	github.com/tendermint/fundraising v0.3.1-0.20220613014523-03b4a2d4481a
```
Pantani, line 17: can we avoid using the commit version and use only `v0.3.1`?

lubtd: We need a recent release of `fundraising` that includes after auction creation hooks

For final version, we will use `fundraising` post audit merge

x/campaign/keeper/campaign.go
lubtd: Yes, we could actually have it in the test package as a utility but I would think it's fine this way since `campaign.go` contains general low level methods to interact with campaigns in the store

x/campaign/keeper/campaign_auction_event_hooks.go
lubtd: I think this is fine `VoucherCampaign ` simply returns an error in general if the coin is not a voucher

Pantani (approved):

aljo242 (approved):

PR #894, opened 2022-07-17
refactor(`participation`): use `sdk.Int` type for allocations 

aljo242:
### What does this PR does?

Removes a potential panic when calling SDK `TruncateInt64` for the `TotalAllocations()` query.  An error is now returned before the panic can be reached.


tbruyelle:
Hey sorry to stick my nose here, but I really don't understand how a panic can happen when calling `TruncateInt64()` on a negative number. 

For instance `sdk.NewDec(-1).TruncateInt64()` doesn't trigger any panic and just return `-1`.

Do you have the trace of the panic @aljo242 please ?

giunatale:
> Hey sorry to stick my nose here, but I really don't understand how a panic can happen when calling `TruncateInt64()` on a negative number.
> 
> For instance `sdk.NewDec(-1).TruncateInt64()` doesn't trigger any panic and just return `-1`.
> 
> Do you have the trace of the panic @aljo242 please ?

`-1` is ok, but the panic risk do indeed exist. This is visible when looking at the source:
https://github.com/cosmos/cosmos-sdk/blob/main/math/dec.go#L676

An out of bound number would cause a panic

tbruyelle:
> An out of bound number would cause a panic. I would actually guess that the risk exist on both ends (so also a really big positive number)

Indeed, so maybe you can also cover the big positive number case with an additional condition, wdyt ?

aljo242:
Yes @tbruyelle , I hadn't fully understood the cause of the panic when I wrote this - apologies.  I ran into a panic with a large negative number during simulation testing.  Modifying this PR now !

lubtd:
> Hey sorry to stick my nose here

This is very appreciated!

My bad, I misread the method as well looking at the PR.

My suggestion would be to refactor and remove completely the usage of `uint64` and changing the logic of the module to use `sdk.Int`

tbruyelle:
Question, why do we deal with `uint64` in query response like 

```go
message QueryGetAvailableAllocationsResponse {
  uint64 availableAllocations = 1;
}
```

to represent something that can overlap `int64` (I assume this was the reason of the panic). Why not a simple string ?
I noticed that strings are often used in the cosmos-sdk to represent numbers, it is to mitigate overlapping, right?

giunatale:
> to represent something that can overlap `int64` (I assume this was the reason of the panic). Why not a simple string ? I noticed that strings are often used in the cosmos-sdk to represent numbers, it is to mitigate overlapping, right?
> 
> (both `sdk.Dec` and `sdk.Int` marshal into string)

It's because the protobuf encoding would be non-deterministic otherwise, that's the reason for using strings. Would be the same using bytes, I guess strings makes it easier also for reading the json equivalent.

> Question, why do we deal with `uint64` in query response like
Allocations cannot be negative, proper typing that scopes out the result prevents errors. The panic was due to an out-of-bound error, the fact that it was a negative number was just coincidental. A big positive number would have caused the same issue

aljo242:
> Thanks for the explanation @giunatale pray
> 
> Just a though about replacing `uint64` by `sdk.Int` in the return of `GetTotalAllocations`: it seems like the conversion from `sdk.Dec` to `sdk.Int` can still trigger a panic if the underlying `big.Int` is out of some specific bounds (see `sdk.Dec.TruncateInt()`). Not sure if this is the same bounds than in `sdk.Dec.TruncateInt64` that said.

It appears that [TruncateInt()](https://github.com/cosmos/cosmos-sdk/blob/6a3d10bc57169add2402ae7da54e6474f2533708/types/decimal.go#L685) does not actually have the panic, so this should be ok.

aljo242:
> Thanks for the explanation @giunatale pray
> 
> Just a though about replacing `uint64` by `sdk.Int` in the return of `GetTotalAllocations`: it seems like the conversion from `sdk.Dec` to `sdk.Int` can still trigger a panic if the underlying `big.Int` is out of some specific bounds (see `sdk.Dec.TruncateInt()`). Not sure if this is the same bounds than in `sdk.Dec.TruncateInt64` that said.

Yes, it appears that the max bitlength for `sdk.Int` is 256 as opposed to 64, so we at least have larger bounds to operate within.

lubtd:
Is this ready for review @aljo242 ?

aljo242:
Ready for review

x/participation/keeper/msg_withdraw_allocations_test.go
```diff
@@ -165,7 +167,12 @@ func Test_msgServer_WithdrawAllocations(t *testing.T) {
 			// check usedAllocationEntry is correctly decreased
 			postUsedAllocations, found := tk.ParticipationKeeper.GetUsedAllocations(tmpSdkCtx, tt.msg.Participant)
 			require.True(t, found)
-			require.Equal(t, preUsedAllocations.NumAllocations-preAuctionUsedAllocations.NumAllocations, postUsedAllocations.NumAllocations)
+			calculated := preUsedAllocations.NumAllocations.Sub(preAuctionUsedAllocations.NumAllocations)
+			if calculated.IsZero() {
+				require.True(t, postUsedAllocations.NumAllocations.IsZero())
+				return
+			}
+			require.Equal(t, calculated, postUsedAllocations.NumAllocations)
```
tbruyelle, line 175: You can do this with a single assertion :

```suggestion
			require.Equal(t, calculated.Int64(), postUsedAllocations.NumAllocations.Int64())
```

Or if you prefer to not call `Int64()` :

```suggestion
			require.True(t, postUsedAllocations.NumAllocations.Equal(calculated), "numAlloc %s not equal to calculated %s", postUsedAllocations.NumAllocations, calculated)
```

proto/participation/auction_used_allocations.proto
```diff
@@ -1,12 +1,17 @@
...
 message AuctionUsedAllocations {
   string address = 1;
   uint64 auctionID = 2;
-  uint64 numAllocations = 3;
-  bool withdrawn = 4;
+  bool withdrawn = 3;
+  string numAllocations = 4 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable) = false
+  ];
```
tbruyelle, line 16: I suppose there's no running testnet for spn, because this change can break it :sweat_smile:

lubtd: No testnet

We also used to have a `alpha` and a `nightly` deployed instances so an instance stays sable

testutil/networksuite/networksuite.go
```diff
@@ -234,7 +234,8 @@ func populateParticipation(r *rand.Rand, participationState participation.Genesi
 	// add used allocations
 	for i := 0; i < 5; i++ {
 		usedAllocations := participation.UsedAllocations{
-			Address: sample.Address(r),
+			Address:        sample.Address(r),
+			NumAllocations: sdk.NewInt(r.Int63()),
```
lubtd, line 238: What about implementing `sample.Int()`?

testutil/sample/participation.go
```diff
@@ -57,17 +57,17 @@ func ParticipationGenesisStateWithAllocations(r *rand.Rand) participation.Genesi
...
 			auctionUsedAllocs := participation.AuctionUsedAllocations{
 				Address:        addr,
 				AuctionID:      uint64(j),
-				NumAllocations: uint64(r.Int63n(5) + 1),
+				NumAllocations: sdk.NewInt(r.Int63n(5) + 1),
 				Withdrawn:      false,
 			}
 			genState.AuctionUsedAllocationsList = append(genState.AuctionUsedAllocationsList, auctionUsedAllocs)
-			usedAllocs.NumAllocations += auctionUsedAllocs.NumAllocations
+			usedAllocs.NumAllocations = usedAllocs.NumAllocations.Add(auctionUsedAllocs.NumAllocations)
```
lubtd, line 70:
```suggestion
			usedAllocs.NumAllocations = usedAllocs.NumAllocations.Mul(sdk.NewInt(2))
```

lubtd (commented):
LGTM

What about also applying my first suggestion?

>Would it be possible to replace in total_allocation_test.go, wantError bool with a more accurate err error and assert ErrorIs(...) with ErrInvalidAddress and ErrInvalidAllocationAmount?
So we ensure the correct check if performed for negative value

lubtd (changes requested):
We should fix genesis in `localnet`

The only changes should be to put `requiredAllocations` values in `spn/genesis_template.json` and `testnet/genesis_template.json` in quotes as those are now `sdk.Int` to be unmarshalled

localnet/spn/genesis_template.json
```diff
@@ -355,7 +355,7 @@
         "participationTierList": [
           {
             "tierID": 1,
-            "requiredAllocations": 1,
+            "requiredAllocations": "1",
```
lubtd, line 358: There are still values for other tier below

lubtd (changes requested):

tbruyelle (dismissed):

lubtd (approved):

PR #896, opened 2022-07-19
fix: app.go const lint

lubtd:
Fix lint error based on cosntant name format in `app.go`

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -127,11 +127,11 @@ import (
 )
 
 const (
-	// mission ID for staking mission to claim airdrop
-	MISSION_ID_STAKING = 1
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -127,11 +127,11 @@ import (
 )
 
 const (
-	// mission ID for staking mission to claim airdrop
-	MISSION_ID_STAKING = 1
+	// MissionIdStaking is the mission ID for staking mission to claim airdrop
```
aljo242, line 130: Since these constants are only used within the scope of this package, we should name them to `missionIDStaking` and `missionIDVoting`.  I think "Id" should be "ID" as well.

aljo242 (changes requested):

aljo242 (approved):

PR #901, opened 2022-07-22
chore: some localnet fixes

lubtd:
Include changes from `campaign` and `mint` modules in genesis templates


aljo242 (approved):
works locally for me !

Issue #909, opened 2022-07-26
Prepare for Cosmos SDK `v0.46.x`


aljo242:
**Is your feature request related to a problem? Please describe.**

Cosmos SDK `v0.46.x` is  a heavily API-breaking update.  In order to update `spn` the following must be done:
- [x] remove depreceated `rest` API imports https://github.com/tendermint/spn/pull/908
- [x] remove the usage of `sdk.Int.ToDec()`
- [x] decouple `spn` from `ignite` (this will avoid the current circular dependency)
- [X] upgrade `ibc-go` to a version that is compliant with Cosmos SDK `v0.46.x`
- [X] upgrade the imported `x/fundraising` module for compliance with Cosmos SDK `v0.46.x`

The `ignite` CLI depends on `spn` so this is a critical task that must be completed before `spn` can be upgraded in order to enable `ignite` to be upgraded

PR #911, opened 2022-07-26
refactor: remove `ignite` dep

aljo242:
### What does this PR does?

Removes the `ignite` dependency from the repository.  Importing `ignite` introduces a set of circular dependencies because:
- ignite` imports `spn`
- `spn` imports `ignite`
- `spn` imports `x/fundraising`
- `x/fundraising` imports `ignite`

This cycle has made upgrading the packages very challenging, so this PR helps alleviate that. The main import was the `pkg/cosmsocmd` module which has now been added to the local `cmd` package.  This functionality is commonly placed in this local package as can be seen with [gaia](https://github.com/cosmos/gaia/blob/198fc5bbcc288bbcef0845e78c3a0943e9389c41/cmd/gaiad/cmd/root.go).  The `openAPI` code has also been removed.  This can potentially be added back in later.  We should prioritize decoupling from `ignite-cli` though.

Part of #909

review them:

- [ ] Include tests
- [ ] Respect code style and lint
- [ ] Update documentation (*.md) (if needed)


app/app.go
```diff
@@ -876,7 +875,7 @@ func (app *App) RegisterAPIRoutes(apiSvr *api.Server, apiConfig config.APIConfig
 
 	// register app's OpenAPI routes.
 	apiSvr.Router.Handle("/static/openapi.yml", http.FileServer(http.FS(docs.Docs)))
-	apiSvr.Router.HandleFunc("/", openapiconsole.Handler(spntypes.Name, "/static/openapi.yml"))
+	//apiSvr.Router.HandleFunc("/", openapiconsole.Handler(spntypes.Name, "/static/openapi.yml"))
```
lubtd, line 878:
```suggestion
	// apiSvr.Router.HandleFunc("/", openapiconsole.Handler(spntypes.Name, "/static/openapi.yml"))
```

cmd/root.go
```diff
@@ -0,0 +1,470 @@
...
+	return exportableApp.ExportAppStateAndValidators(forZeroHeight, jailAllowedAddrs)
+}
+
+// initAppConfig helps to override default appConfig template and configs.
+// return "", nil if no custom configuration is required for the application.
+func initAppConfig() (string, interface{}) {
+	// The following code snippet is just for reference.
+
+	// WASMConfig defines configuration for the wasm module.
+	type WASMConfig struct {
```
lubtd, line 422: There may be unnecessary logic that we can remove, but we can also track this in further issue

lubtd (dismissed):
LGTM

lubtd (approved):
LGTM, can run commands, etc...

PR #914, opened 2022-07-27
feat: Cosmos SDK `v0.46.0`

aljo242:
Updates the Cosmos SDK to `v0.46.0`.  Please review this PR after the other onces I have submitted are completed.

This PR builds off of those, and will be much more manageable to review if the other PRs are already merged.  

- makes changes according to the [changelog](https://github.com/cosmos/cosmos-sdk/blob/release/v0.46.x/CHANGELOG.md)

This PR introduces temporary `replace` clauses in our `go.mod` for the following packages:
- `github.com/cosmos/ibc-go/v5 => github.com/aljo242/ibc-go/v5 v5.1.0`
- `github.com/tendermint/fundraising => github.com/aljo242/fundraising v0.1.2-0.20220726203752-c39abf16b02d`

Once the respective repos are updated, we can remove these lines.  I have an open [PR](https://github.com/tendermint/fundraising/pull/192) for `x/fundraising` and `ibc-go` is near completion.

Part of #909 

## Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Include tests
- [ ] Respect code style and lint
- [ ] Update documentation (*.md) (if needed)

PR #915, opened 2022-07-27
feat(`participation`): decay support

lubtd:
Add in the params of the module decay information.

If enabled, a decay factor is applied to received rewards depending on the current timestamp

Decay factor change is linear from start to end. At end airdrop is no longer claimable.


lubtd:
Let's wait for https://github.com/tendermint/spn/pull/914 before merging this PR

Otherwise, I still open it for review @aljo242

aljo242 (approved):
LGTM - just run `make format` before merging

PR #917, opened 2022-07-31
feat: buf proto generation

aljo242:
### What does this PR does?

Adds [buf](https://docs.buf.build/installation) tooling and proto targets.  This brings our codebase in line with the Cosmos-SDK and their proto compilation model.  

I know we previously wanted to have no scripts in the repo, but this is the only way to get proto compilation working until `ignite` is updated.  The work done in this PR can also serve as a guide for how to update `ignite`'s proto compilation steps. 

Either way, I think we should keep the `buf` tools as they can be pretty helpful.

Pulling this into our `v0.46.0` branch because that makes the most sense.

Most of the modified files are just formatted `.proto` file and regenerated `.pb.go` files.


lubtd:
Sounds good. I think we should support or adapt to buf with Ignite CLI

aljo242:
> Sounds good. I think we should support or adapt to buf with Ignite CLI

Yeah, I think this is possible and will be important.  Chains like: 
- [osmosis](https://github.com/osmosis-labs/osmosis/tree/main/proto)
- [sifchain](https://github.com/Sifchain/sifnode)
- [crypto.org](https://github.com/crypto-org-chain/chain-main)
use it.  And the [sdk](https://github.com/cosmos/cosmos-sdk/tree/main/proto) and [ibc-go](https://github.com/cosmos/ibc-go/tree/main/proto) use it as well.

lubtd (approved):
Works well for me

PR #936, opened 2022-08-26
feat(`launch`): default chain account balances

### LINKED ISSUE #927: `Chain` structure: add the ability to impose the account balance to be requested
**Is your feature request related to a problem? Please describe.**
When launching a testnet, the purpose of the account is to have some tokens to interact with the chain, performing a self-delegation fo a validator, etc... tokens has no monetary values and there is no interest for an account to have more tokens than other accounts

Letting accounts request more tokens can be dangerous if a coordinator approve by mistake an account with 10x more tokens for example and the validator end up with the ability to self-delegate more than 1/3 of the supply. This can be particularly dangerous for incentivized testnet.

**Describe the solution you'd like**
In `MsgCreateChain` and `Chain` structure, adding an option for account to impose a specific balance.

Example:
```
Chain:
  AccountBalance: sdk.Coins
```

Applying an `AddGenesisAccount` or `AddVestingAccount` automatically use this value for the account balance

aljo242:
Closes #927

### What does this PR do?

Adds the `DefaultAccountBalance` field to `Chain` and `MsgCreateChain`.  If this field is empty (`sdk.NewCoins()`).  the behavior is unchanged.  However, if it is populated, the balances of accounts that are created during `keeper.ApplyRequest()` are set to `chain.DefaultAccountBalance`.

### How to test?

```shell
ignite chain serve --skip-proto -r
spnd tx profile create-coordinator --from alice   
spnd tx launch create-chain chain-1 "" "" --from alice --default-account-balance 1000test
spnd tx launch request-add-account 1 20000othercoins --from alice     
spnd q  launch list-genesis-account 1                             
```


aljo242:
I regenerated the `proto` files using `make proto-gen` and `make proto-format` so many of them were modified.

lubtd:
What's your reasoning to consider it should be named `DefaultAccountBalance`?

To me, it's a balance imposed by the coordinator. Default would mean this is the value set if the account doesn't provide a specific amount.

But it can be argued as well we can name it to "default" because this the balance that will be defined by default or accounts at first before the chain is launched and coins are moving

proto/launch/tx.proto
```diff
@@ -29,27 +29,37 @@ service Msg {
...
-  string coordinator   = 1;
-  uint64 launchID      = 2;
-  bool   setCampaignID = 3;
-  uint64 campaignID    = 4;
-  bytes  metadata      = 5;
+  string   coordinator                                    = 1;
+  uint64   launchID                                       = 2;
+  bool     setCampaignID                                  = 3;
+  uint64   campaignID                                     = 4;
+  repeated cosmos.base.v1beta1.Coin defaultAccountBalance = 5 [
```
lubtd, line 57: This invalidate all previous approved requests (because the balance is persisted in the store in `GenesisAccounts`)

I think we should remove it from `MsgEditChain` and consider the coordinator should publish a new chain if they would to set this value

lubtd (commented):
Looks great

aljo242:
> What's your reasoning to consider it should be named `DefaultAccountBalance`?
> 
> To me, it's a balance imposed by the coordinator. Default would mean this is the value set if the account doesn't provide a specific amount.
> 
> But it can be argued as well we can name it to "default" because this the balance that will be defined by default or accounts at first before the chain is launched and coins are moving

Yeah, I agree default doesn't make as much sense as I originally thought.  I'm going to change it back to `AccountBalance`

lubtd:
Hey @aljo242 is it ready for reviews?

I also wanted to add that we should add a test case in `TestApplyRequest` for the particular case the balance is defined

But I realize the current structure of `TestApplyRequest` is hard to maintain and should be improved first https://github.com/tendermint/spn/issues/944

lubtd (approved):

PR #948, opened 2022-09-08
refactor: replace unix time stamps

### LINKED ISSUE #913: Replace Unix timestamps with `Duration` or `Time`
**Is your feature request related to a problem? Please describe.**
We used Unix timestamps initially to represent timestamps, then we switched to `Duration` as in the `participation` module.

Unix timestamps are still used in many places like launchTimestamp for the chain.

We should use `Duration` and `Time` everywhere as it is more powerful and allow better code readability

aljo242:
Closes #913 

The only remaining time that is not `time.Time` or `time.Duration` is in `proto/types/ibc.proto`, where a timestamp is [saved as a string](https://github.com/tendermint/spn/blob/aab4d3df1f2bce0e8b3163858740015a08d9f1b6/proto/types/ibc.proto#L17).  I wasn't sure what we wanted to do with this one.


lubtd:
> The only remaining time that is not time.Time or time.Duration is in proto/types/ibc.proto, where a timestamp is [saved as a string](https://github.com/tendermint/spn/blob/aab4d3df1f2bce0e8b3163858740015a08d9f1b6/proto/types/ibc.proto#L17). I wasn't sure what we wanted to do with this one.

We can actually let the string

The reason is we want the `ConsensusState` type to be directly the unmarshaled version of the consensus state file dumped by the `ibc` CLI query

lubtd (approved):
LGTM

PR #957, opened 2022-09-19
feat(`launch`): change param request

### LINKED ISSUE #930: Implementing `ChangeParam` request
Need https://github.com/tendermint/spn/issues/928 to be implemented first

Add in the available requests the `ChangeParam` request that allows requesting the modification of a module param

Proposed structure:
```
type ChangeParam stuct {
  Module string
  Param string
  Value []byte // json structure
}
```

Example:
```
{
  Module: "staking"
  Param: "max_validators"
  Value: "42"
}
```

Request to change the max number of validator param of staking module to 42.

This request can only modify parameters value. It can be argued that we eventually want a request to modify any genesis state value. `ChangeParam` is a simpler implementation as the path to the value is predictable `[module]/params/[name]`, and those represent the value aimed to be changed from community decision

aljo242:
Closes #930

@lubtd let me know what you think !


proto/launch/request.proto
```diff
@@ -31,6 +31,7 @@ message RequestContent {
     GenesisValidator genesisValidator = 3;
     AccountRemoval   accountRemoval   = 4;
     ValidatorRemoval validatorRemoval = 5;
+    ChangeParam      changeParam      = 6;
```
lubtd, line 34:
```suggestion
   ParamChange      changeParam      = 6;
```
To stay consistent with request name being noun instead of verb

x/launch/keeper/change_param.go
```diff
@@ -0,0 +1,51 @@
...
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.ChangeParamKeyPrefix))
+	b := k.cdc.MustMarshal(&changeParam)
+	store.Set(types.ChangeParamPath(
+		changeParam.Module,
+		changeParam.Param,
+	), b)
+}
+
+// GetChangeParam returns a changeParam from its index
+func (k Keeper) GetChangeParam(
```
lubtd, line 21: `ChangeParam` should also be indexed by the launch ID to determine which chain get param changed

x/launch/types/request_content.go
```diff
@@ -195,3 +197,25 @@ func (m ValidatorRemoval) Validate() error {
...
+				Param:  param,
+				Value:  value,
+			},
+		},
+	}
+}
+
+// Validate implements ChangeParam validation
+func (m ChangeParam) Validate() error {
+	if m.Module == "" || m.Param == "" {
```
lubtd, line 216: I think we could consider it only contains alpha chars

lubtd (changes requested):
We need to add logic in `genesis.go` for genesis export and import with ParamChange objects in the genesis state.

The easiest way to represent the genesis state would be to add `LaunchID` in `ParamChange` so information of type index is self-contained
```
message ParamChange {
  string module = 1;
  string param  = 2;
  bytes  value  = 3;
}
```
and adding in `(m RequestContent) Validate(launchID uint64)` verification of the launchID same as for `GenesisAccount` request

x/participation/simulation/simulation.go
```diff
@@ -74,69 +69,6 @@ func SimulateMsgParticipate(
 	}
 }
 
-func SimulateCreateAuction(
```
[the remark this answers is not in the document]
aljo242: This was just leftover from when `fundraising` didn't support simulation.  I can remove this change from the PR since this removal isn't necessarily relevant for it.  I must have removed when i was making other sim fixes.

lubtd (changes requested):
We should add a query to list all param changes from a `LaunchID`

I think only the list query is sufficient, show query would not be necessary

x/launch/client/cli/query.go
```diff
@@ -33,11 +33,14 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 		CmdListVestingAccount(),
 		CmdShowGenesisValidator(),
 		CmdListGenesisValidator(),
+		CmdListParamChange(),
 		CmdShowRequest(),
 		CmdListRequest(),
 		CmdQueryParams(),
 	)
 
+	cmd.AddCommand(CmdListParamChange())
```
lubtd, line 42:
```suggestion
```

lubtd (approved):

PR #958, opened 2022-09-20
feat(`launch`): config file for genesis

### LINKED ISSUE #395: Support Config file for initial genesis
We currently support two different initial genesis types for a chain launch

- `DefaultInitialGenesis`: the default genesis of the chain (from all default genesis state of modules)
- `GenesisURL`: a URL containing a genesis file

`GenesisURL` is practical to use when dumping the state of a previous chain but if the user simply want to change some params of some modules, it requires the user from uploading the genesis

We can simplify simple genesis modification by supporting initial genesis through Starport Config

`StarportConfigGenesis` or simply `ConfigGenesis` would use a specified Starport Config file to construct the initial genesis.
This would allow simple params modification:
```
genesis:
  app_state:
    staking:
      params:
        bond_denom: "spn"
```

We should store the path from the source directory to the Starport config and the hash of the genesis file. Source information is already stored in `Chain`
```
message ConfigGenesis {
  string path = 1;
  string hash = 2;
}
```

aljo242:
Closes #395

### What does this PR do?

Includes a refactor of `MsgCreateChain` and other associated functions to take `InitialGenesis` as argument instead of `genesisURL` and `genesisHash`.


proto/launch/chain.proto
```diff
@@ -56,4 +57,10 @@ message DefaultInitialGenesis {}
 message GenesisURL {
   string url  = 1;
   string hash = 2;
+}
+
+// ConfigGenesis specifies using a custom genesis from a starport config file
```
lubtd, line 62:
```suggestion
// ConfigGenesis specifies using a custom genesis from a Ignite CLI config file
```

x/launch/client/cli/tx_create_chain.go
```diff
@@ -20,9 +20,10 @@ import (
 )
 
 const (
-	flagGenesisURL     = "genesis-url"
-	flagCampaignID     = "campaign-id"
-	flagAccountBalance = "account-balance"
+	flagGenesisURL        = "genesis-url"
+	flagGenesisConfigFile = "genesis-config file"
```
lubtd, line 24:
```suggestion
	flagGenesisConfigFile = "genesis-config"
```

x/launch/client/cli/tx_create_chain.go
```diff
@@ -58,6 +62,23 @@ func CmdCreateChain() *cobra.Command {
...
+				initialGenesis = types.NewGenesisURL(genesisURL, genesisHash)
+			}
+
+			// parse genesis config file for initialGenesis
+			genesisConfigFile, err := cmd.Flags().GetString(flagGenesisConfigFile)
+			if err != nil {
+				return err
+			}
+			if genesisConfigFile != "" {
+				if args[1] == "" {
```
lubtd, line 74: Source URL is always necessary not just when providing config for initial genesis

x/launch/client/cli/tx_create_chain.go
```diff
@@ -58,6 +62,23 @@ func CmdCreateChain() *cobra.Command {
...
+					return errors.New("genesis config file requires a source url to be set")
+				}
+				initialGenesis = types.NewConfigGenesis(genesisConfigFile)
+			}
+
+			if genesisURL != "" && genesisConfigFile != "" {
```
lubtd, line 80: I don't think we need to return an error. If no option is specified then `DefaultInitialGenesis` is used

aljo242: This is catching the case where a user is setting BOTH genesisURL and Config file.  This is just enforcing that they can only set one or the other - not both, which is invalid.  The default case is still `DefaultInitialGenesis`.

aljo242: added a comment to clarify

lubtd (approved):

PR #959, opened 2022-09-21
chore: update proto files from ignite cli generation

lubtd:
Update proto files from generation from command
```
ignite c serve
```


aljo242:
no files changed now - i think we can close

PR #996, opened 2022-12-02
chore(ci/coverage): test all go packages and remove the non-testable files from coverage

Pantani:
## Description

If we have a package without test files (`_test.go`), the `go test` will exclude this package from the coverage. We should add the `-coverpkg=./...` to cover all packages. This PR also removes Go files autogenerated by the protobuf and, `app` and `cmd` packages from the coverage.


aljo242:
@Pantani this is sweet!! Do you think we can set something like this up for the CLI?

Looks like its just failing because there is no actual `./tests` directory

aljo242 (approved):

lubtd (approved):
LGTM
thanks

PR #997, opened 2022-12-06
feat: add `MaxMetadataLength` param to campaign keeper

### LINKED ISSUE #541: Create MaxMetadataLength param
Currently, the maximum length of a metadata field is hard coded [here](https://github.com/tendermint/spn/blob/develop/pkg/types/metadata.go).  It makes more sense to have this as an on chain param.

An issue here is that both the `launch` and `campaign` modules use the metadata field.  Params are stored per-module, but I think the ideal behavior here is that we have a singular MaxMetadataLength param shared by all modules.  

Currently, this can be done by storing the param in one of the modules (say, `campaign`) and then accessing the param from the other module (`launch`) using its keeper.

Pantani:
Closes #541

### What does this PR do?

Remove the hardcoded `MaxMetadataLength` variable storing it to campaign module params, and then access the param from the other `launch` using its keeper.

### How to test?

Run unit tests for the `campaign` and `launch` modules.

## Checklist

review them:

- [x] Include tests
- [x] Update documentation (*.md) (if needed)


aljo242:
Can you run `make format`?

Pantani:
> Can you run `make format`?

I already ran it, but sometimes the imports are not grouped correctly, so I manually fixed it and reran it

x/campaign/keeper/msg_create_campaign.go
```diff
@@ -14,6 +14,17 @@ import (
 func (k msgServer) CreateCampaign(goCtx context.Context, msg *types.MsgCreateCampaign) (*types.MsgCreateCampaignResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	// check if the metadata length is valid
+	p := k.GetParams(ctx)
+	maxMetadataLength := p.MaxMetadataLength
+	if len(msg.Metadata) > int(maxMetadataLength) {
```
lubtd, line 20:
```suggestion
	if uint64(len(msg.Metadata)) > maxMetadataLength {
```
we should not use this type of conversation

x/campaign/keeper/msg_create_campaign.go
```diff
@@ -14,6 +14,17 @@ import (
...
+		return nil, sdkerrors.Wrapf(types.ErrInvalidMetadataLength,
+			"data length %d is greater than maximum %d",
```
lubtd, line 22:
```suggestion
			"metadata length %d is greater than maximum %d",
```

x/campaign/keeper/msg_create_campaign_test.go
```diff
@@ -126,6 +127,16 @@ func TestMsgCreateCampaign(t *testing.T) {
 			},
 			err: types.ErrFundCommunityPool,
 		},
+		{
+			name: "should prevent validation of msg with invalid metadata length",
```
lubtd, line 131:
```suggestion
			name: "should fail when the change had too long metadata",
```
The message is already validated at this stage in the literal sense

x/campaign/keeper/msg_edit_campaign.go
```diff
@@ -14,6 +14,16 @@ import (
 func (k msgServer) EditCampaign(goCtx context.Context, msg *types.MsgEditCampaign) (*types.MsgEditCampaignResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	// check if the metadata length is valid
+	maxMetadataLength := k.GetParams(ctx).MaxMetadataLength
+	if len(msg.Metadata) > int(maxMetadataLength) {
```
lubtd, line 19: Same comment as above

x/campaign/keeper/msg_edit_campaign_test.go
```diff
@@ -101,6 +101,16 @@ func TestMsgUpdateCampaignName(t *testing.T) {
 			},
 			err: profiletypes.ErrCoordInvalid,
 		},
+		{
+			name: "should prevent validation of msg with invalid metadata length",
```
lubtd, line 105: Same comment as above for test name

x/campaign/types/params.go
```diff
@@ -12,12 +12,14 @@ import (
 )
 
 var (
-	DefaultMinTotalSupply      = sdkmath.NewInt(100)                   // One hundred
-	DefaultMaxTotalSupply      = sdkmath.NewInt(1_000_000_000_000_000) // One Quadrillion
-	DefaultCampaignCreationFee = sdk.Coins(nil)                        // EmptyCoins
+	DefaultMinTotalSupply             = sdkmath.NewInt(100)                   // One hundred
+	DefaultMaxTotalSupply             = sdkmath.NewInt(1_000_000_000_000_000) // One Quadrillion
+	DefaultCampaignCreationFee        = sdk.Coins(nil)                        // EmptyCoins
+	DefaultMaxMetadataLength   uint64 = 1000
```
lubtd, line 18: Can we actually increase it to `2000`?
Since we have more metadata we're considering for the launchpad

x/campaign/types/params.go
```diff
@@ -86,3 +100,10 @@ func validateCampaignCreationFee(i interface{}) error {
 	}
 	return v.Validate()
 }
+
+func validateMaxMetadataLength(i interface{}) error {
```
lubtd, line 104: We should add this method to `ValidateBasic` above

lubtd, line 104: Can't see a test for this method, do we have one?

x/launch/keeper/msg_create_chain.go
```diff
@@ -13,6 +13,16 @@ import (
 func (k msgServer) CreateChain(goCtx context.Context, msg *types.MsgCreateChain) (*types.MsgCreateChainResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
+	// check if the metadata length is valid
+	maxMetadataLength := k.campaignKeeper.GetParams(ctx).MaxMetadataLength
```
lubtd, line 17: This is not the same type of metadata for chains

We need to implement another param in the `launch` module for the chain metadata that can be updated independently

x/launch/keeper/msg_edit_chain.go
```diff
@@ -13,9 +13,20 @@ import (
...
 func (k msgServer) EditChain(goCtx context.Context, msg *types.MsgEditChain) (*types.MsgEditChainResponse, error) {
-	var err error
+	var (
+		err error
+		ctx = sdk.UnwrapSDKContext(goCtx)
+	)
 
-	ctx := sdk.UnwrapSDKContext(goCtx)
+	// check if the metadata length is valid
+	maxMetadataLength := k.campaignKeeper.GetParams(ctx).MaxMetadataLength
```
lubtd, line 22: Same, `launch` metadata param to be used

x/launch/keeper/msg_create_chain.go
```diff
@@ -13,6 +13,16 @@ import (
...
+	if len(msg.Metadata) > int(maxMetadataLength) {
```
lubtd, line 18: Change conversion as above

x/launch/keeper/msg_create_chain_test.go
```diff
@@ -132,6 +143,11 @@ func TestMsgCreateChain(t *testing.T) {
 			msg:  sample.MsgCreateChain(r, coordAddrs[4], "", false, campMap[coordAddrs[4]]),
 			err:  types.ErrFundCommunityPool,
 		},
+		{
+			name: "should create a chain with invalid metadata length",
```
lubtd, line 147:
```suggestion
			name: "should prevent create a chain with invalid metadata length",
```

x/launch/keeper/msg_edit_chain.go
```diff
@@ -13,9 +13,20 @@ import (
...
+	if len(msg.Metadata) > int(maxMetadataLength) {
```
lubtd, line 23: Change conversion as above

lubtd (changes requested):

x/campaign/types/params.go
Pantani: added

x/campaign/types/params_test.go
```diff
@@ -18,23 +18,43 @@ func TestParamsValidateBasic(t *testing.T) {
...
-			params: NewParams(DefaultMinTotalSupply, DefaultMaxTotalSupply, sdk.Coins{sdk.Coin{Denom: "foo", Amount: sdkmath.NewInt(-1)}}),
-			err:    errors.New("coin -1foo amount is not positive"),
+			name: "should prevent validate invalid coins for campaign creation fee",
+			params: NewParams(
+				DefaultMinTotalSupply,
+				DefaultMaxTotalSupply,
+				sdk.Coins{sdk.Coin{Denom: "foo", Amount: sdkmath.NewInt(-1)}},
+				DefaultMaxMetadataLength,
+			),
+			err: errors.New("coin -1foo amount is not positive"),
```
lubtd, line 57: We should add a test case in `ValidateBasic` for metadata length too big

x/launch/types/params_test.go
```diff
@@ -21,13 +21,27 @@ func TestParamsValidate(t *testing.T) {
...
-			params: NewParams(DefaultMinLaunchTime, DefaultMaxLaunchTime, DefaultRevertDelay, DefaultFee, DefaultFee),
+			name: "should validate valid params",
+			params: NewParams(
+				DefaultMinLaunchTime,
+				DefaultMaxLaunchTime,
+				DefaultRevertDelay,
+				DefaultFee,
+				DefaultFee,
+				DefaultMaxMetadataLength,
+			),
```
lubtd, line 44: Same, an additional test for invalid metadata length

x/campaign/types/params_test.go
Pantani: The `validateMaxMetadataLength` only checks if the type is right, not if the length is too big. Should we define a maximum length for the user can set thought params?

lubtd: Yes right, this is the validity of the new value to be provided that is checked.

What I meant is `ValidateBasic` check `validateMaxMetadataLength`, we should have a test case that verify `ValidateBasic` upon failure on `validateMaxMetadataLength`

Pantani: But if we create a validation regarding the `MaxMetadataLength` should have a max to be set, it makes sense to test the failure cases for `validateMaxMetadataLength`. IMHO don't make sense put the `MaxMetadataLength` as a param if we still have a hardcoded value limiting the maximum

lubtd: Yes, it's true we can't define invalid test case as the type is already cast in the call

lubtd (approved):
Working well👍

PR #998, opened 2022-12-06
refactor: avoid use paramset for `getParam` keeper method

Pantani:
### What does this PR do?

Fetch params from the launch without use the `paramset`


x/launch/keeper/params.go
```diff
@@ -33,14 +33,9 @@ func (k Keeper) RequestFee(ctx sdk.Context) (requestFee sdk.Coins) {
...
-func (k Keeper) GetParams(ctx sdk.Context) types.Params {
-	return types.NewParams(
-		k.LaunchTimeRange(ctx).MinLaunchTime,
-		k.LaunchTimeRange(ctx).MaxLaunchTime,
-		k.RevertDelay(ctx),
-		k.ChainCreationFee(ctx),
-		k.RequestFee(ctx),
-	)
+func (k Keeper) GetParams(ctx sdk.Context) (params types.Params) {
+	k.paramstore.GetParamSet(ctx, &params)
```
lubtd, line 37: AFAIK `params` module is to be removed from the SDK (instead having params directly stored in the module keeper space following a naming convention)
And therefore this new version would be obsolete as `paramstore` would be removed

cc @aljo242

Pantani: hmm, I didn't know it. So we should change the `GetParamSet` from the other modules. I will open an issue for it

aljo242: Yes,

The new convention will be:

- Define `MsgUpdateParams` tx in the given module.  [example](https://github.com/cosmos/cosmos-sdk/blob/ecb4ca2a82c45296ae9409ec7c7294d902067b61/proto/cosmos/mint/v1beta1/tx.proto#L23).
- Define some additional logic for it in `types` [example](https://github.com/cosmos/cosmos-sdk/blob/main/x/mint/types/msgs.go).
- Add logic in `keeper` [keeper.go](https://github.com/cosmos/cosmos-sdk/blob/ecb4ca2a82c45296ae9409ec7c7294d902067b61/x/mint/keeper/keeper.go#L83), [msg_server.go](https://github.com/cosmos/cosmos-sdk/blob/main/x/mint/keeper/msg_server.go).

lubtd (approved):

lubtd (changes requested):
Looks like there is an issue in the build process

lubtd (approved):

PR #1004, opened 2022-12-13
feat: add request creation airdrop mission

lumtis:
Add an airdrop mission triggering when a new request is created/sent

This is an actual mission as sending request is part of the wide usage of the protocol and this serve as an example on how to create a custom airdrop mission outside of regular modules

Add a mock for launch hooks for hook call assertion


app/airdrop_missions.go
```diff
@@ -0,0 +1,33 @@
+package app
```
Pantani, line 1: Can we move this file to the `pkg/airdrop` folder? For we can avoid increasing the `app` package

app/airdrop_missions.go
```diff
@@ -0,0 +1,33 @@
...
+
+// RequestCreated completes mission when a request is created
+func (h MissionSendRequestHooks) RequestCreated(
+	ctx sdk.Context,
+	creator string,
+	_,
+	_ uint64,
+	_ launchtypes.RequestContent,
+) {
+	h.k.CompleteMission(ctx, h.missionID, creator)
```
Pantani, line 32: maybe we should add some logs for success and also the `CompleteMission` error

Pantani (approved):

PR #1005, opened 2022-12-15
chore: upgrade IBC to v6

aljo242:
Updates:
- `ibc-go` to `v6`
- `sdk` to `v0.46.7`
- `cosmos-proto` to `v1.0.0-beta.1`
- `tendermint` to `v0.34.24`

Modified `monitoringp` and `monitoringc` modules because the signature for `SendPacket` has been changed.

Tests working for me, but this could use some further investigation/testing


aljo242:
`ignite/modules` needs to be upgrade to `ibc-go/v6` first.  Then I can update the dependency.

lubtd (approved):
LGTM

lubtd:
> Modified monitoringp and monitoringc modules because the signature for SendPacket has been changed.
Tests working for me, but this could use some further investigation/testing

There is just the sequence added as far as I can see and it is retrieve by the channel keeper.
I don't see modification that influence verification logic

aljo242:
Ok - feel free to merge

PR #1017, opened 2023-06-26
chore: update buf version and files

Pantani:


jeronimoalbi:
I'm not sure about this but it might be good to update/add the proto generation Makefile target as part of this PR to generate the code from the proto files using "Buf". Also we might consider changing the proto file layout to match the project layout. For example, `proto/spn/launch/chain.proto` might need to be `x/launch/types/chain.proto`.

jeronimoalbi (approved):

